PR #15, opened 2020-01-09
build(deps): BREAKING CHANGE: update `derive_more` 0.15 -> 0.99 for rumq-{client,core}

ErichDonGubler:
This is a breaking change because `From` implementations (in particular, those implemented by `derive_more::From` on `rumq_client::Error`) are public API surface -- this could be kept as an implementation detail in the future by using an error newtype.

### DIFF
--- a/rumq-core/src/lib.rs
+++ b/rumq-core/src/lib.rs
@@ -110,6 +110,7 @@ pub fn connect_return(num: u8) -> Result<ConnectReturnCode, Error> {
 
 #[derive(Debug, From)]
 pub enum Error {
+    #[from(ignore)]
     InvalidConnectReturnCode(u8),
     InvalidProtocolName(String),
     InvalidProtocolLevel(String, u8),
@@ -118,6 +119,7 @@ pub enum Error {
     UnsupportedProtocolName,
     UnsupportedProtocolVersion,
     UnsupportedQoS,
+    #[from(ignore)]
     UnsupportedPacketType(u8),
     UnsupportedConnectReturnCode,
     PayloadSizeIncorrect,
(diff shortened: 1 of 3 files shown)


rumq-core/src/lib.rs
```diff
@@ -110,6 +110,7 @@ pub fn connect_return(num: u8) -> Result<ConnectReturnCode, Error> {
 
 #[derive(Debug, From)]
 pub enum Error {
+    #[from(ignore)]
```
tekjar: Why do we need this ignore?

tekjar: Oh this is to prevent `From u8` conflict..I wonder how the old version of `derive-more` worked

tekjar: Let's do this instead

```
#[derive(Debug, From)]
pub enum Error {
    InvalidConnectReturnCode(u8),
    InvalidProtocolName(String),
    InvalidProtocolLevel(String, u8),
    IncorrectPacketFormat,
    InvalidTopic,
    UnsupportedProtocolName,
    UnsupportedProtocolVersion,
    UnsupportedQoS,
    UnsupportedPacketType(u8),
    UnsupportedConnectReturnCode,
    PayloadSizeIncorrect,
    PayloadTooLong,
    PayloadRequired,
    #[from]
    TopicNameMustNotContainNonUtf8(FromUtf8Error),
    TopicNameMustNotContainWildcard,
    MalformedRemainingLength,
    UnexpectedEof,
    #[from]
    Io(io::Error),
}

```

tekjar:
Also can we only select necessary features to have better compile times?

```
derive_more = { version = "0.99", default-features = false, features = ["from"] }
```

PR #42, opened 2020-03-07
Reconnection fix

tekjar:
* Fixes reconnection issue during idle state after a reconnection due to halfopen connection
* Notify network closes and done request streams
* Expose parts of `MqttEventLoop` to control the state before next reconnection


tekjar:
Ping @qm3ster @mojzu

PR #55, opened 2020-03-19
Last will + some changes to names

TotalKrill:
In short: 
 - Added so that last_will works
 - Removed a lot of trailing spaces / tabs
 - Added `new()` methods to Publish and Subscribe
 - Added a more descriptive `create_eventloop` that is just using `eventloop()` internally
 - Added deprecated flag to eventloop so that it can be renamed in the future without much hassle


tekjar:
All good. I'll merge this to a different branch (with connection logic changes)

PR #60, opened 2020-04-12
minor refactorings

JOE1994:
@tekjar First of all, Thank you for your hard work on this repo! :smiley_cat: 

I'm in the process of learning **MQTT**,
and I hope I could meaningfully contribute to this repo soon :smiley: 

This PR contains minor refactorings which won't affect the program logic.


tekjar:
@JOE1994 Thanks :). If you already have some experience with rust, I can mentor on a task to add higher level bindings to this crate so that users doesn't have to understand async or depend on tokio directly

JOE1994:
> @JOE1994 Thanks :). If you already have some experience with rust, I can mentor on a task to add higher level bindings to this crate so that users doesn't have to understand async or depend on tokio directly

That sounds intriguing :+1: I also joined the Discord chat for the project with the nickname `JOE1994`.

PR #62, opened 2020-04-15
Impl std::error::Error for a bunch of error enums

### LINKED ISSUE #61: impl Error for EventLoopError
Just saw that `EventLoopError` doesn't implement `std::error::Error`. Is this a known limitation or something I can easily add?

killercup:
- Adds dependency on `thiserror`, but also removes `derive-more`
- Adds Error implementation on all enums found with `enum.*Error`
- This also required implementations for Display, which were added in a
  best-effort fashion.

Fixes #61

### DIFF
--- a/rumq-broker/Cargo.toml
+++ b/rumq-broker/Cargo.toml
@@ -15,9 +15,9 @@ tokio-util = { version = "0.3", features = ["codec"] }
 futures-util = { version = "0.3", features = ["sink"] }
 tokio-rustls = "0.13"
 rumq-core = { path = "../rumq-core", version = "0.1.0-alpha.8" }
-derive_more = { version = "0.99", default-features = false, features = ["from"] }
 log = "0.4"
 serde = { version = "1", features = ["derive"] }
 bytes = "0.5"
+thiserror = "1.0.15"
 
 [dev-dependencies]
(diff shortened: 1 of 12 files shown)


killercup:
`thiserror` also allows you to do `#[error(transparent)]` instead of giving it a format string. Might be useful for some of these, but I didn't want to go through everything and make choices, so this this a very mechanical change only.

tekjar:
Thanks :)

killercup:
My pleasure! Can you do another release soon? Then I can use this in my project without referring to my fork :)

tekjar:
Sure. I'll do one now. I'm also thinking if I should lockstep `core` and `client` to avoid version confusion. What do you think?

killercup:
That'd definitely make releases easier. I've done that for most of my projects that consist of multiple crates, too.

tekjar:
Published to crates.

Also I've finished all the unit tests that I could possibly think of. Benchmarks are also really good when compared to go paho client.  I'll probably release a 1.0 by the end of this month (probably with the name rumqtt). Do let me know if you have any issues, I'll fix them on priority

killercup:
That's awesome to hear! I'll make sure to test our internal stuff using this crate a bit more and report back to you! :)

PR #72, opened 2020-05-15
Update doc comment

qm3ster:
Not sure how much of `.await` and `?` should be included.
The phrasing "poll it with tokio" sounds very beginnerish, so that might be relevant to include in the same vein.
That has indeed been a change, where it used to neither return a Future not a Result, all errors going inside the Stream.


tekjar:
Error goes into the stream and stream ends in next iteration

PR #74, opened 2020-05-16
Ecc key

amrx101:
Changes summary:

- Added `key_type` attribute to MqttOptions of Type `Option<String>`
- Panics if `key_type` supplied by user is illegal.
- Call appropriate key reading api depending upon type of key.
- By default assume `RSA`.

### DIFF
--- a/rumq-client/src/lib.rs
+++ b/rumq-client/src/lib.rs
@@ -257,6 +257,8 @@ pub struct MqttOptions {
     inflight: usize,
     /// Last will that will be issued on unexpected disconnect
     last_will: Option<LastWill>,
+    /// Key type for TLS 
+    key_type: Option<String>,
 }
 
 impl MqttOptions {
@@ -283,6 +285,7 @@ impl MqttOptions {
             throttle: Duration::from_micros(0),
             inflight: 100,
             last_will: None,
+            key_type: None,
         }
     }
 
@@ -433,6 +436,18 @@ impl MqttOptions {
     pub fn inflight(&self) -> usize {
         self.inflight
     }
+
+    /// Set key type
+    pub fn set_key_type(&mut self, key_type: String) -> &mut Self {
+        self.key_type = Some(key_type);
+        self
+    }
+
+    /// get key type
+    pub fn get_key_type(&self) -> Option<String>{
+        self.key_type.clone()
+    }
+
 }
 
 #[cfg(test)]
--- a/rumq-client/src/lib.rs
+++ b/rumq-client/src/lib.rs
@@ -439,7 +439,11 @@ impl MqttOptions {
 
     /// Set key type
     pub fn set_key_type(&mut self, key_type: String) -> &mut Self {
-        self.key_type = Some(key_type);
+        match key_type.as_str() {
+            "RSA" => self.key_type = Some(key_type),
+            "ECC" => self.key_type = Some(key_type),
+            _ => panic!("Key type should be either ECC or RSA"),
+        };
         self
     }
 
--- a/rumq-client/src/lib.rs
+++ b/rumq-client/src/lib.rs
@@ -469,4 +469,10 @@ mod test {
     fn no_client_id() {
         let _mqtt_opts = MqttOptions::new("", "127.0.0.1", 1883).set_clean_session(true);
     }
+
+    #[test]
+    #[should_panic]
+    fn set_key_type_illegal_panics() {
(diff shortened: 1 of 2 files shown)


TotalKrill:
Hello!

Generally I do not like it when code panics, I have not tried using Keys with this library yet. But isnt there someway we could return an error and let the user of the code determine if that is an panic, or if it should report the failure some other way. 

Either way I prefer having the user handle errors and turning them into errors instead of the library doing such things. MQTT is just a subsystem in my program, so the software will still have to do other things even if MQTT is not configured correctly.

amrx101:
@TotalKrill , 

I decided to panic when an user sets wrong key type for specific reasons:

- Key types are specific. Any user who sets up a TLS/SSL auth needs to be aware about all configs pertaining to TLS/SSL.
- It aligns with existing way of things done. We do raise a panic when an invalid `keep_alive` and `inflight` is configured by user. For authenticated connections `Key type` is as significant as `keep_alive` or `inflight`.

Yes, there is a way to return an `Error` when an illegal argument is set which the user can handle. But I believe that can be explored for less critical options than something which has to do with security. But I believe we would some way of restrictions users from specifying illegal options. I would not want to run things with illegal configs. More-ever if TLS auth is not desired, the changes dont affect the user at-all. 

I will let @tekjar take a decision on whether we should panic or return an `Error` when an illegal option is supplied by user.

TotalKrill:
If one would for example create a gui apolocation. It would be painful if that application crashed  if a user would input something invalid.

I mean, for a userless application maybe that's what you would want, sometimes. Which is why unwrap and expect exists.

Just because there are existing panics in configurations options does not make it a good design decision.

amrx101:
@TotalKrill , removed the need of panic. Now we control as to what can be the supplied values for `key_type` by exposing a `public enum`.

TotalKrill:
It looks much better!

rumq-client/src/lib.rs
```diff
@@ -257,6 +263,8 @@ pub struct MqttOptions {
     inflight: usize,
     /// Last will that will be issued on unexpected disconnect
     last_will: Option<LastWill>,
+    /// Key type for TLS 
+    key_type: Option<String>,
```
tekjar, line 267: This should be `Key` enum

rumq-client/src/lib.rs
```diff
@@ -433,6 +442,21 @@ impl MqttOptions {
     pub fn inflight(&self) -> usize {
         self.inflight
     }
+
+    /// Set key type
+    pub fn set_key_type(&mut self, key_type: Key) -> &mut Self {
```
tekjar, line 447: Use this setter to modify `key_type` enum value

rumq-client/src/network.rs
```diff
@@ -25,6 +25,8 @@ pub enum Error {
     TLS(#[from] TLSError),
     #[error("No valid cert in chain")]
     NoValidCertInChain,
+    #[error("Invalid key type")]
```
tekjar, line 28: I guess this won't be necessary after the above review comments are handled

rumq-client/src/lib.rs
tekjar: `` key_type: Key``

rumq-client/src/network.rs
```diff
@@ -52,7 +54,22 @@ pub async fn tls_connect(options: &MqttOptions) -> Result<TlsStream<TcpStream>,
...
+                    "RSA" => rsa_private_keys(&mut BufReader::new(Cursor::new(client.1.clone()))),
+                    "ECC" => pkcs8_private_keys(&mut BufReader::new(Cursor::new(client.1.clone()))),
+                    // Invalid Key Type.
+                    _ => return Err(Error::InvalidKeyType),
+                }
+            },
+            // Assume RSA key type by default?
+            None => rsa_private_keys(&mut BufReader::new(Cursor::new(client.1.clone()))),
+        };
+        let mut keys = match read_keys {
```
tekjar, line 69: Finally this block can be simplified to 

```rust
  let read_keys = match options.get_key_type() {
      Key::RSA => rsa_private_keys(&mut BufReader::new(Cursor::new(client.1.clone()))),
      Key::ECC => pkcs8_private_keys(&mut BufReader::new(Cursor::new(client.1.clone()))),
  }
```

tekjar:
Thanks :)

AlexandreCassagne:
Hello! Can you confirm this only applies to the client?

PR #79, opened 2020-06-16
Allow handling failed connection

TotalKrill:
Removed unwrap from connection code, we might be simplifying things, but we are not python, we dont want libraries to crash our application in rust.


TotalKrill:
added so that one can start the connection in the background from the client if one would want to

PR #88, opened 2020-07-06
add resending of unacknowledged publish packets

andyfozzy:
Adds a timer to the event loop to resend publish packets with QoS 1 which have not been acknowledged.
A resend timeout is added to the MqttOptions struct.
The outgoing_pub VecDeque now holds the original send time of the publish packet alongside the packet.

### DIFF
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -60,6 +60,8 @@ pub struct EventLoop<R: Requests> {
     cancel_tx: Option<Sender<()>>,
     /// Delay between reconnection (after a failure)
     reconnection_delay: Duration,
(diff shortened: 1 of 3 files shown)


tekjar:
Inflight queue logic in master looks like this

- Pretinitialize Vec<Option<Publish>> based on inflight queue size. 
- Each index represents packet id
- When all the inflight messages are acked, packet id is reset to 1

As a result
- Out of order acks are handled with O(1) instead of O(n) to find correct publish to remove
- Any unacked packets are detected due to collision while reusing packet ids. User can use it for retransmission

tekjar:
I'm closing this PR. But thank you for your work. Feel free to open a new issue if this doesn't work for you

PR #92, opened 2020-07-21
Fix compilation on Windows

skreborn:
`rumqttlog` - and by extension its dependents - cannot currently be compiled on and for Windows because `std::os::unix` is used unconditionally in the `segment` module. This change adds a conditionally-compiled inline function that keeps the current code for Unix systems while implementing a slightly different version for Windows.


tekjar:
Thanks!! :)

PR #96, opened 2020-07-24
Added a check on topics containing newline characters

VictorKoenders:
Ran into an issue where my topic accidentally had a newline character. rumqttc handled this by disconnecting and not logging any errors.

### DIFF
--- a/mqtt4bytes/src/lib.rs
+++ b/mqtt4bytes/src/lib.rs
@@ -32,6 +32,7 @@ pub enum Error {
     BoundaryCrossed,
     MalformedRemainingLength,
     InsufficientBytes(usize),
+    InvalidTopic,
 }
 
 /// Encapsulates all MQTT packet types
--- a/mqtt4bytes/src/packets/publish.rs
+++ b/mqtt4bytes/src/packets/publish.rs
@@ -287,5 +287,21 @@ mod test {
             ]
         );
     }
-}
 
+    #[test]
+    fn topic_no_newline() {
+        let publish = Publish {
+            dup: false,
+            qos: QoS::AtMostOnce,
+            retain: false,
+            topic: "a\nb".to_owned(),
+            pkid: 0,
+            payload: Bytes::from(vec![0xE1, 0xE2, 0xE3, 0xE4]),
+        };
+        let mut buf = BytesMut::new();
+        let result = publish.write(&mut buf);
+
+        let err = result.expect_err("Cannot write a topic with a newline");
+        assert_eq!(err, crate::Error::InvalidTopic);
+    }
+}
--- a/mqtt4bytes/src/packets/publish.rs
+++ b/mqtt4bytes/src/packets/publish.rs
@@ -88,6 +88,9 @@ impl Publish {
     }
 
     pub fn write(&self, payload: &mut BytesMut) -> Result<usize, Error> {
+        if self.topic.contains(|c| c == '\n' || c == '\r') {
+            return Err(Error::InvalidTopic);
+        }
         let dup = self.dup as u8;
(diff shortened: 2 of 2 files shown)


skreborn:
The MQTT specification [does not specifically prohibit anything other than the null character](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718109), so if a change is required to abide by the specification, we should probably do that instead. I don't know it's being checked now, but it should also be a maximum of 65535 characters long.

Now, would anyone realistically do that on purpose? Probably not, but we should still use the specification as the single source of truth, especially since nothing stops a user from interchangeably using `rumqtt` with any other compliant solution.

Can you perhaps open an issue instead and attach a minimal reproduction?

VictorKoenders:
The spec does indeed support newline characters, however either this library, or mosquitto caused a disconnect when using newline characters. I'm guessing somewhere else the spec is not being followed.

It's probably something that needs to be looked into, this was a quick monkey patch to notify people of weird behavior.

For now my issue is resolved and I don't have time to look into specifics right now, so I'll close this PR. If someone runs into this issue they can research it further

tekjar:
Thanks for the report!! I confirm and close this PR

skreborn:
@tekjar I've tried this against a number of brokers, and so far it seems like only a select few are capable of handling special characters in compliance with the specifications.

Namely, both [EMQX](https://emqx.io) and [ejabberd](https://www.ejabberd.im/) swallow virtually everything you throw at it (I've tested ASCII 1-255 specifically, and the message arrived back safe and sound using every character except for the obvious `/#+` trio).

Unfortunately, other brokers I tested ([mosquitto](https://mosquitto.org/) and [HiveMQ](https://www.hivemq.com/)) refuse to accept anything with any non-printable characters. `0x20` to `0x7E` is fine, but step outside the safe range and all hell breaks loose.

In any case, it seems to me that `rumqtt` generates the correct packet regardless of the characters within the topic name, and instead this error occurs on the broker side of things.

tekjar:
Thanks for testing this @skreborn. Appreciate it :). Hopefully these brokers atleast throw some useful errors in future

skreborn:
@tekjar Unfortunately, they do not seem to be leaning that way. Mosquitto quite literally reports a socket error in verbose mode and nothing else in case this happens. If `rumqd` can handle these well, I think you've already done everything you can and should.

ralight:
I just happened upon this and wanted to point out that you're mistaken. Both the MQTT v3.1.1 and v5.0 specs say that control characters in a UTF-8 string may be treated as being a protocol error, and hence the connection should be closed. See sections 1.5.3 and 1.5.4 respectively.

skreborn:
> Both the MQTT v3.1.1 and v5.0 specs say that control characters in a UTF-8 string may be treated as being a protocol error

Apologies for the oversight, you are right. Thank you for looking into it! For reference, here is the relevant portion of the specification @ralight mentions:

> ### 1.5.3 UTF-8 encoded strings
> 
> [...]
>
> The data SHOULD NOT include encodings of the Unicode [Unicode] code points listed below. If a receiver (Server or Client) receives a Control Packet containing any of them it MAY close the Network Connection:
>
> U+0001..U+001F control characters
> U+007F..U+009F control characters 
> Code points defined in the Unicode specification [Unicode] to be non-characters (for example U+0FFFF) 

The operating word, of course, is _MAY_, so as far as I can tell, there's not much if anything we can do here. If a broker or client can decide on a whim whether to drop a connection upon receiving a control character, we can only really strive to teach the users of the library through documentation to keep that in mind.

Issue #105, opened 2020-08-05
rumqttc regression: Publish::set_retain(true) is ignored since #99


qwandor-google:
#99 seems to have broken Publish::set_retain(true). Before this change it works; after it messages which I try to send retained are not.

PR #106, opened 2020-08-05
Don't lose pkid, dup and retain when converting Publish to PublishRaw.

qwandor-google:
This fixes bug #105, which was introduced by pull #99.


tekjar:
Thank you :)

PR #111, opened 2020-08-12
Set mtu

amrx101:
Fixes#110

`Rustls` set the mtu to maximum by default which may cause `TLS error` on networks wherein the mtu is less than the payload size. 

A default `mtu` of 1500 is set.

1. [Ref 1](https://community.cisco.com/t5/other-network-architecture/why-the-mtu-size-is-1500/m-p/105422/highlight/true#M35328)
2. [Ref 2](https://www.igvita.com/2013/10/24/optimizing-tls-record-size-and-buffering-latency/)
3. [Ref 3](https://fasterdata.es.net/network-tuning/mtu-issues/debugging-mtu-problems/)

It is advisable to do a ping test to figure the most optimal mtu for custom set ups.


amrx101:
Closing as some unwanted changes have come in

PR #112, opened 2020-08-12
Set MTU

### LINKED ISSUE #110: Option to set `MTU`.
We had an issue wherein the network we were trying to talk to had a `mtu` set to a value less than the payload we were trying to send.

`Openssl`, `Golang` set the `MTU` to minimum threshold whereas `RUSTLS` sets `mtu` to the maximum possible value in the spec. `RUSTLS` however exposes a method to set `mtu`. We need to utilize this and have an option to set `mtu` that percolates down to `RUSTLS`.

References

- [1](https://fasterdata.es.net/network-tuning/mtu-issues/debugging-mtu-problems/)
- [2](https://www.networkworld.com/article/2224654/mtu-size-issues.html)
- [3](https://blog.cloudflare.com/ip-fragmentation-is-broken/)

amrx101:
`RUSTLS` sets `MTU` to maximum possible size. This will cause `TLS` error if we send a payload greater than `MTU` of network.

A default value of 1500 is set as recommended by `Cisco` although a ping test is recommended to determine the best `MTU` for a set up.

References

1.[ Ref 1](https://blog.cloudflare.com/path-mtu-discovery-in-practice/)
2. [Ref 2 ](https://community.cisco.com/t5/other-network-architecture/why-the-mtu-size-is-1500/m-p/105422/highlight/true#M35328)
3. [Ref 3](https://networkdirection.net/articles/network-theory/mtu-and-mss/use-ping-to-find-the-mtu/)
4. [Ref 4](https://www.igvita.com/2013/10/24/optimizing-tls-record-size-and-buffering-latency/)

fixes #110

### DIFF
--- a/rumqttc/src/lib.rs
+++ b/rumqttc/src/lib.rs
@@ -254,6 +254,8 @@ pub struct MqttOptions {
     tls_client_config: Option<Arc<ClientConfig>>,
 
     conn_timeout: u64,
+    /// mtu
+    mtu: Option<usize>,
 }
 
 impl MqttOptions {
@@ -283,6 +285,7 @@ impl MqttOptions {
             key_type: Key::RSA,
             tls_client_config: None,
             conn_timeout: 5,
+            mtu: Some(1500),
         }
     }
 
@@ -491,6 +494,17 @@ impl MqttOptions {
     /// get timeout in secs
     pub fn timeout(&self) -> u64 {
         self.conn_timeout
+    }
+
+    /// set MTU. Perform a Ping test to determine most optimal MTU
+    pub fn set_mtu(&mut self, mtu: usize) -> &mut Self {
+        self.mtu = Some(mtu);
+        self
+    }
+
+    /// get mtu
+    pub fn get_mtu(&self) -> Option<usize> {
+        self.mtu
     }
 }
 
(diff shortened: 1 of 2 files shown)


tekjar:
I think it's not realistic to expose all rustls options via MqttOptions. Probably just use [set_client_tls_config](https://github.com/bytebeamio/rumqtt/blob/master/rumqttc/src/lib.rs#L463)  method?

amrx101:
@tekjar , I have an idea. Why dont we eliminate the creation of `ClientConfig` from inside the client altogether.

Earlier we had JWT token generation of `GCloud` in the library itself. If we take the creation of `ClientConfig` out, we can rely on users to create and provide appropriate `TLS Config`.

We can provider a builder in examples to drive home the point.


```

fn tls_client() -> Result<Client> {
    let mut cc = rustls::ClientConfig::new();
    let cert_bytes = include_bytes!("certs/cacert.pem");
    let cert = rustls::internal::pemfile::certs(&mut Cursor::new(&cert_bytes[..]))
        .map_err(|_| Error::from("Error parsing cert file"))?[0].clone();
    cc.root_store.add(&cert)
        .map_err(|e| Error::from_std_err(e))?;
}

```

And then this `ClientConfig` can be passed to our `Options`?

tekjar:
We can if we only want to support only rustls. Otherwise we should find a way to abstract these well.

tekjar:
Closing this in favor of ClientConfig

PR #114, opened 2020-08-15
Use incoming max size from options

tekjar:


FranciscoKnebel:
Any planned releases of rumqttc including these changes? Need a larger packet size than 1KB.

tekjar:
@FranciscoKnebel I'm planning a release today

PR #121, opened 2020-08-25
No need for EventLoop::new to be async.

qwandor-google:
It never calls await, so the async declaration is unnecessary.


tekjar:
Interesting. Last  I checked, I couldn't construct [delay](https://github.com/bytebeamio/rumqtt/blob/ac484644ee92753beca076cdc25247f4037c9833/rumqttc/src/eventloop.rs#L87) outside async.

qwandor-google:
Constructing it should be fine, it's only awaiting it that needs to be async.

tekjar:
Intuitively yes. But timer construction is checking if it's inside runtime

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=93e4ca361177411dd8feb9d84cc19d58

qwandor-google:
Is there any downside to constructing the `Delay` inside `select` instead?

tekjar:
Hmm..I'll think more about this 

* During high load, this might be too many new delay objects. But this doesn't matter as long as there is no difference in perf
* Mqtt server will expect a PING after 'keep alive' seconds of no outgoing activity from the client. So, client should save this state across multiple [poll](https://github.com/bytebeamio/rumqtt/blob/master/rumqttc/src/eventloop.rs#L113)s

qwandor-google:
To the second point, isn't that handled by saving the Instant on the EventLoop? If it ends up being in the past, then the Delay should complete immediately.

I'm not sure how the performance of constructing a new Delay versus changing an existing one compares, though aren't the other cases in the select also constructing new futures? So maybe it's not any worse than that? This is all speculation though, I haven't looked into the implementation details or run an benchmarks to compare.

tekjar:
> To the second point, isn't that handled by saving the Instant on the EventLoop? If it ends up being in the past, then the Delay should complete immediately.

I wrongly assumed that you meant to create a `Delay` in every `poll`. I think you mean to make it an Option. That works!!

qwandor-google:
> > To the second point, isn't that handled by saving the Instant on the EventLoop? If it ends up being in the past, then the Delay should complete immediately.
> 
> I wrongly assumed that you meant to create a `Delay` in every `poll`. I think you mean to make it an Option in `EventLoop`. That works!!

I didn't mean to make it an Option, though I guess that is also a possibility. Have a look at the second commit I pushed to this branch; I create the Delay in the `select` method, but I create it from an `Instant` which is saved on the `EventLoop` struct. The question is whether it is expensive to create a `Delay`, and my guess is that it shouldn't be particularly, but I don't have data to back that up.

tekjar:
@qwandor-google Ahh. Sorry I didn't notice the changes. This is good.

tekjar:
There seems to be a conflict.Can you check?

tekjar:
nvm. That's just 1 file. I'll just fix it here in github.

tekjar:
Thanks @qwandor-google :)

PR #124, opened 2020-08-30
Handle disconnect requests.

### LINKED ISSUE #123: Sending a Request::Disconnect to event loop handle causes panic
If I do something like this:

```
let event_loop = EventLoop::new(mqtt_options, REQUESTS_CAP).await;
let requests_tx = event_loop.handle();
requests_tx.send(Request::Disconnect).await;
```

It panics at state.rs:138 because it is unimplemented.

I'll send a patch which seems to fix this.

qwandor-google:
This should fix #123.


tekjar:
Thanks!! :)

PR #126, opened 2020-08-31
Use Bytes rather than String for LastWill.

### LINKED ISSUE #122: LastWill message is String instead of Vec<u8>
I am creating a Sparkplug™ B client as my first big Rust project. The Sparkplug™ B specification requires `LastWill` message be a binary Protocol Buffers type. But the type on rumqttc `LastWill` is `String`.

Is there any other way to create the connection using a `Vec<u8>` or `Bytes` or something of that nature as the data type for the `message` field of `LastWill` struct?

qwandor-google:
This also adds a constructor, consistent with Publish.

Fixes #122.

### DIFF
--- a/mqtt4bytes/src/packets/connect.rs
+++ b/mqtt4bytes/src/packets/connect.rs
@@ -1,4 +1,5 @@
 use alloc::string::String;
+use alloc::vec::Vec;
 use crate::*;
 use super::*;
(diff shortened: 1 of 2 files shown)


tekjar:
Thats fast. Thanks @qwandor-google . You are taking a lot of load off me :)

PR #129, opened 2020-09-05
Cleanup

### LINKED ISSUE #125: rustfmt everything
My editor is set to automatically run rustfmt on save. However, the rumqtt code hasn't been run through rustfmt before, so this always generates a bunch of spurious changes which I have to revert. How would people feel about running rustfmt over everything, to avoid this problem in future?

Alternatively, it might be possible to add a rustfmt.toml which disables it.

qwandor-google:
Run cargo fmt, and fix warnings about the return value of std::mem::replace being unused by just assigning instead.

Fixes #125.


tekjar:
Can you rebase?

qwandor-google:
Done.

PR #136, opened 2020-09-11
Change return type of rumqttc event loop

bastikr:
Hi!
First of all, thank you all for working on this very nice mqtt library. Looking through the code I was a little bit surprised by the choice that the rumqttc eventloop yielded
```rust
Result<(Option<Incoming>, Option<Outgoing>), ConnectionError>.
```
which made it non-obvious if it could be the case that both an incoming and an outgoing event would be available at the same time and/or if it might be possible that there might be no event at all. This pull request changes the return type to
```rust
Result<Event, ConnectionError>
```
where the event is just
```rust
pub enum Event {
    Incoming(Incoming),
    Outgoing(Outgoing),
}
```
However, since I'm not familiar with this code, there might be a good reason for the previous design choice, so please don't feel pressured to accept this pull request.

### DIFF
--- a/benchmarks/rumqttasync.rs
+++ b/benchmarks/rumqttasync.rs
@@ -1,4 +1,4 @@
-use rumqttc::{self, EventLoop, Incoming, MqttOptions, PublishRaw, QoS, Request};
+use rumqttc::{self, Event, EventLoop, Incoming, MqttOptions, PublishRaw, QoS, Request};
 use std::error::Error;
 use std::time::{Duration, Instant};
 
@@ -39,23 +39,11 @@ pub async fn start(id: &str, payload_size: usize, count: usize) -> Result<(), Bo
     let mut acks_count = 0;
     let start = Instant::now();
     loop {
-        let (notification, _) = eventloop.poll().await?;
-        let notification = match notification {
-            Some(n) => n,
-            None => continue,
-        };
-
-        match notification {
-            Incoming::PubAck(_puback) => {
-                acks_count += 1;
+        if let Event::Incoming(Incoming::PubAck(_)) = eventloop.poll().await? {
+            acks_count += 1;
+            if acks_count == count {
+                break;
             }
-            _notification => {
-                continue;
-            }
-        };
-
-        if acks_count == count {
-            break;
         }
     }
 
--- a/benchmarks/rumqttasyncqos0.rs
+++ b/benchmarks/rumqttasyncqos0.rs
@@ -1,4 +1,4 @@
-use rumqttc::{self, EventLoop, Incoming, MqttOptions, PublishRaw, QoS, Request};
+use rumqttc::{self, Event, EventLoop, Incoming, MqttOptions, PublishRaw, QoS, Request};
 use std::error::Error;
 use std::time::{Duration, Instant};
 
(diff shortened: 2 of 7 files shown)


tekjar:
Hi @bastikr . Thank you for the contribution. Return was `Option<Incoming>, Option<Outgoing>` due to historical reasons. Simultaneous Incoming and outgoing used to be possible before. But I removed it for simplicity.

As the client is externally polled, `Option<Incoming>, Option<Outgoing>` or `Option<Event>` will give a chance to do empty returns in future without breaking the API.

I don't mind either way but I'll need some consensus before merging. Is the current version not idiomatic?

bastikr:
Hi @tekjar!
I don't think it's a question of being idiomatic or not, it's more about ease of use and expectations. If I see that this function yields `(Option<Incoming>, Option<Outgoing>)`, I expect - or at least have to consider - that the returned value can be:
* `(i, o)`
* `(None, o)`
* `(i, None)`
* `(None, None)`

Now let's assume that we want to handle one specific outgoing and one specific incoming message. Matching is now really awkward:
```rust
match r {
  (Some(I(i)), None) => handle_i(i),
  (None, Some(O(o))) => handle_o(i),
  (Some(I(i)), Some(O(o))) => {handle_i(i); handle_o(o)}
  _ => ...
}
```
and we have to outsource the handling code into a separate function since it has to be used from different matching arms. Compare this to the proposed implementation:
```rust
match r {
  Event::Incoming(I(i)) => { /* handle case i here */ }
  Event::Outgoing(O(o)) => { /* handle case o here */ }
}
```
I prefer this variant by a lot. Do you think there will ever might be a reason to return an incoming and outgoing event together? I can't think of a reason at the moment but that does of course not mean there might not be one. What I don't understand is why you want to return an empty event. If there is nothing to do then there is no reason for a future to complete.

tekjar:
> Now let's assume that we want to handle one specific outgoing and one specific incoming message. Matching is now really awkward:
> 
> ```rust
> match r {
>   (Some(I(i)), None) => handle_i(i),
>   (None, Some(O(o))) => handle_o(i),
>   (Some(I(i)), Some(O(o))) => {handle_i(i); handle_o(o)}
>   _ => ...
> }
> ```

I usually do 

```rust
let (incoming, outgoing) = eventloop.poll.await?;

if let Some(incoming) = incoming {
    // handle incoming
}

if let Some(outgoing) = outgoing {
    // handle outgoing
}
```

What I like with your approach is that it explains  underlying behaviour of the eventloop
    - Either incoming or outgoing
    - Always returns either incoming or outgoing (can catch bugs with empty polls)

> Do you think there will ever might be a reason to return an incoming and outgoing event together

Previously `poll()` used to return everything that's happening with the eventloop. Incoming qos 1 publishes returns `(Publish, PubAck)`. 

For simplicity, I changed the return types to only return User requested network out packets like publish, subscribe etc. So pubacks are not reported in outgoing now. We might add it if someone requests it.

> What I don't understand is why you want to return an empty event. If there is nothing to do then there is no reason for a future to complete.

It's mostly about "something has happened" but "we don't want it report to the user". We can ignore this. I don't see going back to this architecture.

tekjar:
I'll think a bit about simultaneous incoming and outgoing and ping you in a day or two?

bastikr:
No worries, take your time. It's just a suggestion and absolutely not urgent.

tekjar:
Thank you @bastikr. I think this is the right direction to take. We can break in future if necessary.

PR #144, opened 2020-09-21
Add keywords and categories for crates.io.

qwandor-google:
This may make it easier for people to find these crates.


tekjar:
Thanks!!

PR #158, opened 2020-09-30
Fix subcrate versions

tekjar:


jplatte:
mqtt4bytes is part of the public API of rumqttc, so bumping it to a semver-incompatible version is itself a change that requires a semver-incompatible (breaking change) release. It won't affect many users I guess since there is no reason to directly depend on mqtt4bytes when using rumqttc, but it would be good to make sure this doesn't happen again.

PR #159, opened 2020-09-30
Expose constructor to create AsyncClient from senders.

qwandor-google:
This is useful for creating a fake `AsyncClient` for tests, and potentially also for applications which need to mix use of `AsyncClient` with the raw `Sender`s.

### DIFF
--- a/rumqttc/src/client.rs
+++ b/rumqttc/src/client.rs
@@ -29,7 +29,7 @@ pub struct AsyncClient {
 }
 
 impl AsyncClient {
-    /// Create a new `Client`
+    /// Create a new `AsyncClient`
     pub fn new(options: MqttOptions, cap: usize) -> (AsyncClient, EventLoop) {
         let mut eventloop = EventLoop::new(options, cap);
         let request_tx = eventloop.handle();
@@ -41,6 +41,15 @@ impl AsyncClient {
         };
 
         (client, eventloop)
+    }
+
+    /// Create a new `AsyncClient` from a pair of async channel `Sender`s. This is mostly useful for
+    /// creating a test instance.
+    pub fn from_senders(request_tx: Sender<Request>, cancel_tx: Sender<()>) -> AsyncClient {
+        AsyncClient {
+            request_tx,
+            cancel_tx,
+        }
     }
 
     /// Sends a MQTT Publish to the eventloop


tekjar:
Now that we have `AsyncClient`. Do you think we need a raw sender?

qwandor-google:
It might be useful for some applications with complicated requirements?

But my main motivation for doing this is to make it possible to create an `AsyncClient` that talks to a channel that I created rather than actually talking to the network, so I can write unit tests for my code which uses the `AsyncClient`. The idea is that I can make my own async channel pair, create an `AsyncClient` with the sender, pass that to my client, and then write tests where my client code does something and I check that what comes out the other end of the channel is what I expect.

tekjar:
If I understood you correctly
- Use user provided channels
- Use user provided network

That is the direction I started with, but I hard-coded channels and network for simplicity.

One motivation I had for that was to be able to 
- replace tcp with quic
- replace normal channel with disk aware channel to flush data to disk when the network is slow

This PR doesn't satisfy those needs beacuse the sender `from_senders` method is  returned by `Eventloop`. Instead `Eventloop` should have an option to take user provided `Sender` and `Network`

qwandor-google:
For now, I'd just like to be able to migrate to using `AsyncClient` in my library while still being able to keep my existing unit tests, which rely on being able to use a fake channel. Being able to inject a different channel into `EventLoop` could also be useful, but is not necessary for making a fake `AsyncClient` for tests.

Perhaps it helps to include an example of how I am using this. My tests currently look something like this, where `DevicePublisher` is my struct which implements the logic which I am testing:

```rust
let (requests_tx, requests_rx) = async_channel::unbounded();
let publisher = DevicePublisher::new(requests_tx, ...);

/// Test calling methods on publisher
/// Expect entries in requests_rx
```

And the code to use the library in a real application would look like:

```rust
let event_loop = EventLoop::new(mqtt_options, REQUESTS_CAP);
let publisher = DevicePublisher::new(event_loop.handle(), ...);
```

With `AsyncClient`, I would like to be able to write tests like this:

```rust
let (requests_tx, requests_rx) = async_channel::unbounded();
let (cancel_tx, cancel_rx) = async_channel::unbounded();
let client = AsyncClient::from_senders(requests_tx, cancel_tx);
let publisher = DevicePublisher::new(client, ...);

/// Test calling methods on publisher
/// Expect entries in requests_rx and/or cancel_tx
```

And the user of the library would do something like:

```rust
let (client, event_loop) = AsyncClient::new(mqtt_options, REQUESTS_CAP);
let publisher = DevicePublisher::new(client, ...);
```

tekjar:
Ok. Let's merge this. We can think about eventloop taking sender and network in the next release

PR #160, opened 2020-10-02
Move keepalive_timeout assignment (fix panic)

Dirk007:
I use rumqttc in a context, where the `event_loop.poll()` is made inside another `tokio::select!` in a `loop` and where it is therefore possible that another future finishes this select first before a mqtt connection could be completely established (wrong credentials in my case). 

The next call to the `event_loop.poll()` seems to have the `self.network` being `Some()` and thus we have the following:
- the first `connect_or_cancel()` seem not to have returned
- `self.keepalive_timeout` is therefore still `None`
- `self.select()` is being called from `poll()`, resulting in 
- - `_ = self.keepalive_timeout.as_mut().unwrap()` to panic as it is still `None`

I simply moved the part where the `self.keepalive_timeout` is enforced above the `connect_or_cancel()` and - at least for me - the panics are gone.

### DIFF
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -475,6 +475,26 @@ mod test {
     }
 
     #[tokio::test]
+    async fn unsuccessful_connection_should_not_panic_on_ping() {
+        task::spawn(async move {
+            let _broker = Broker::new(1880, false).await;
+            time::delay_for(Duration::from_secs(5)).await;
+        });
+
+        time::delay_for(Duration::from_secs(1)).await;
+        let options = MqttOptions::new("dummy", "127.0.0.1", 1880);
+        let mut eventloop = EventLoop::new(options, 5);
+
+        for _ in 0..2 {
+            let some_event = time::delay_for(Duration::from_secs(1));
+            select! {
+                _ = some_event => {}
+                _poll = eventloop.poll() => {}
+            }
+        }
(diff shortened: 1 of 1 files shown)


tekjar:
So `self.keepalive_timeout` should always be Some. But is with `connect_or_cancel`. Can you add `self.network = None` [here](https://github.com/bytebeamio/rumqtt/blob/master/rumqttc/src/eventloop.rs#L288). Possibly an unit test to check failed `connect_or_cancel` should set network to `None`

Dirk007:
Assigning `None` to `connect_or_cancel` does not help, because `self.network` is always assigned a value instantly again here https://github.com/Dirk007/rumqtt/blob/move_timeout_assignment/rumqttc/src/eventloop.rs#L335

Dirk007:
Ah, I think I got your idea. But I think my way to solve that is ugly to read right now - but have a look :D

tekjar:
Yeah. This the direction I mean. But the CI is failing :). I can take this up on Friday if you can't spend time.

Dirk007:
Maybe I find can some time to learn whats happening there so that the test fails. If not, your help is very appreciated :)

tekjar:
@Dirk007 Can you check if 'fixes' branch solved your problem?

Dirk007:
@tekjar Fixed it - many thanks!

PR #162, opened 2020-10-03
Handle PubRel in outgoing packets

rbino:
PubRel was missing in possible outgoing packets and would panic the program
when publishing with QoS2


tekjar:
Thanks!! I've never tested or got a need use qos 2. More testing before using is suggested :)

PR #165, opened 2020-10-07
rumqttc: Reduce memory usage of MqttState

amikhalev:
Previously MqttState would allocate a `Vec<Option<u16>>` of length 2^16
which would take up 4 * 2^16 bytes (256KiB) for incoming publishes.
Replcae this with a `HashSet` which should have comparable performance
(if not better due to memory locality) and uses much less memory.

### DIFF
--- a/rumqttc/src/state.rs
+++ b/rumqttc/src/state.rs
@@ -1,7 +1,7 @@
 use crate::{Incoming, Request};
 
 use mqtt4bytes::*;
-use std::{mem, time::Instant};
+use std::{collections::HashSet, mem, time::Instant};
 
 #[derive(Debug, thiserror::Error)]
 pub enum StateError {
@@ -58,7 +58,7 @@ pub struct MqttState {
     /// Packet ids of released QoS 2 publishes
     pub(crate) outgoing_rel: Vec<Option<u16>>,
     /// Packet ids on incoming QoS 2 publishes
-    pub incoming_pub: Vec<Option<u16>>,
+    pub incoming_pub: HashSet<u16>,
     /// Last collision due to broker not acking in order
     pub collision: Option<PublishRaw>,
 }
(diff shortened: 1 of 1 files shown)


tekjar:
Performance won't be comparable (atleast in micro benchmarks)  as this involves hashing and memory locality will actually be worse as keys are more dispersed. 

I think the way to go here is to set vec capacity to inflight limit of the broker.

amikhalev:
> I think the way to go here is to set vec capacity to inflight limit of the broker.

Isn't the issue here that the broker decides the pkid, so it is not necessarily going to be less than the inflight limit?

> Performance won't be comparable (atleast in micro benchmarks) as this involves hashing and memory locality will actually be worse as keys are more dispersed.

I would expect that keys are only as dispersed as the number of unreleased packets (well, divided by load factor), which I think stays small but I haven't looked at benchmarks. The default `std` hasher is more expensive for a single `u16` than I thought it would be, but I think in this case it would be appropriate to use a [no-op hash function](https://github.com/paritytech/nohash-hasher).

Another option is to change `incoming_pub` to a [`BitVec`](https://docs.rs/bitvec/0.19.3/bitvec/) which stores a single bit per pkid, cutting memory usage by a factor of 32. This could also be done for `outgoing_rel`.

tekjar:
> Isn't the issue here that the broker decides the pkid, so it is not necessarily going to be less than the inflight limit?

Oh yeah. you are right

> I would expect that keys are only as dispersed as the number of unreleased packets (well, divided by load factor), which I think stays small but I haven't looked at benchmarks.

I don't completely understand this. I mean the calculated hashes are always dispersed to some degree when compared to vec where they are always contiguous. So it won't have better cache locality than vec (as you suggested in the previous comment)

So 256K is not a big deal for a lot of use cases. I'll give some more thought later

TotalKrill:
The nohash-hasher has quite good performance, and i have used the IntMap and IntSet as a drop in replacement for HashMap and HashSet. Depending on the input data I could see 2-5x speedup compared to HashMap some tests i made in a throwaway repository.

mneumann:
My PR https://github.com/bytebeamio/rumqtt/pull/239 uses a bitset for the `incoming_pub`. That drops memory usage from 256 KiB to just 8 KiB.

amikhalev:
Nice!

PR #183, opened 2020-11-06
Update tokio to 0.3

jakubtrnka:
I upgraded tokio to version 0.3.

PR #184, opened 2020-11-10
Update tokio to 0.3

jakubtrnka:
the same as #183

### DIFF
--- a/mqtt4bytes/Cargo.toml
+++ b/mqtt4bytes/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "mqtt4bytes"
-version = "0.3.0"
+version = "0.4.0"
 readme = "README.md"
 authors = ["tekjar <raviteja@bytebeam.io>"]
 edition = "2018"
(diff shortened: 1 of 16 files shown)


tekjar:
I'm ok merging this but I'm not sure how it affects others. Tokio 1.0 is just a month away. Do you think it's wise to wait till then?

jakubtrnka:
That's true with the tokio 1.0.
However, tokio 0.3 is, according to their release notes, beta for tokio 1.0. I expect that upgrade from 0.3 -> 1.0 will be very smooth, perhaps even just formal.

So regardless of whether we wait or do it now, we have to review this and merge.

Concerning users of tokio 0.2 runtime: Like with any other upgrade, if major version number (leftmost nonzero) increases, breaking changes should be expected. Typically cargo dependency definitions use caret notation that doesn't allow breaking changes.

I didn't do rigorous testing, because I don't understand the internals of this library. I only patched changes to tokio and I'm happily using it in my project and wanted to share it with you.

PR #188, opened 2020-11-19
Remove unused enum SecurityOptions

mkfrey:
The code contains an enum `SecurityOptions` which seems to be a leftover of the old rumqtt library and which is not being used anywhere. This PR removes the enum.


tekjar:
Thankyou @mkfrey :)

PR #193, opened 2020-11-23
Add support for websocket transport with optional TLS

MathiasKoch:
This PR will add support for websocket transport through `tungstenite` with optional TLS through `rustls` 

I am not personally super happy about the `set_protocol` requirement in `MqttOptions`, and would have liked a way to automatically detect it, but I cannot quite see a way for that without changing most of the `host`, `port` inputs to be parsed. Eg a single input of the format `localhost:1883`, that could be `wss://localhost:8883` or `https://localhost:port` etc.

Feedback is encouraged!


MathiasKoch:
Hmm.. Not sure why the CI fails? Every test passes locally?

tekjar:
Thank you. CI is failing because of openssl. Do we need `tokio-tls`?

MathiasKoch:
Ahh, well spotted.

> Do we need tokio-tls?

No, i think it might have sneaked past me. Should be fixed now.

tekjar:
Hmm still openssl. Can you check your local cargo lock file to figure out which main dependency is including it?

jplatte:
Probably a missing `default-features = false` ;)

MathiasKoch:
Hmm..

I don't get it.. It comes from:
```
 rumqttc v0.3.0 (/home/mathias/git/blackbird/other/rumqtt/rumqttc)
│       ├── async-channel feature "default" (*)
│       ├── log feature "default" (*)
│       ├── thiserror feature "default" (*)
│       ├── async-tungstenite feature "tokio-rustls"
│       │   ├── async-tungstenite v0.10.0
│       │   │   ├── tungstenite v0.11.1
│       │   │   │   ├── bytes feature "default" (*)
│       │   │   │   ├── log feature "default" (*)
│       │   │   │   ├── rand feature "default" (*)
│       │   │   │   ├── base64 feature "default"
│       │   │   │   │   ├── base64 v0.12.3
│       │   │   │   │   └── base64 feature "std"
│       │   │   │   │       └── base64 v0.12.3
│       │   │   │   ├── byteorder feature "default"
│       │   │   │   │   ├── byteorder v1.3.4
│       │   │   │   │   └── byteorder feature "std"
│       │   │   │   │       └── byteorder v1.3.4
│       │   │   │   ├── http feature "default"
│       │   │   │   │   └── http v0.2.1
│       │   │   │   │       ├── bytes feature "default" (*)
│       │   │   │   │       ├── itoa feature "default" (*)
│       │   │   │   │       └── fnv feature "default"
│       │   │   │   │           ├── fnv v1.0.7
│       │   │   │   │           └── fnv feature "std"
│       │   │   │   │               └── fnv v1.0.7
│       │   │   │   ├── httparse feature "default"
│       │   │   │   │   ├── httparse v1.3.4
│       │   │   │   │   └── httparse feature "std"
│       │   │   │   │       └── httparse v1.3.4
│       │   │   │   ├── input_buffer feature "default"
│       │   │   │   │   └── input_buffer v0.3.1
│       │   │   │   │       └── bytes feature "default" (*)
│       │   │   │   ├── native-tls feature "default"
│       │   │   │   │   └── native-tls v0.2.6
│       │   │   │   │       ├── log feature "default" (*)
│       │   │   │   │       ├── openssl feature "default"
│       │   │   │   │       │   └── openssl v0.10.30
```

But when i look at `async-tungstenite v0.10.0`: https://github.com/sdroege/async-tungstenite/blob/master/Cargo.toml#L37 it clearly states `default-features = false` ? But it still looks like `native-tls` is pulled in by it's `default` feature set?

MathiasKoch:
Ahh, think i found it: https://github.com/najamelan/ws_stream_tungstenite/blob/dev/Cargo.toml#L18 I'll open a PR

MathiasKoch:
That did it!

Now i just have the one issue mentioned in OP left. I am not too fond of the way to enter (host, port, protocol), but i think any better way would come with a breaking change of storing a `Url` rather than the three?

tekjar:
I went through the code. I start a thorough review and test when you think this is ready to be merged. I didn't understand why we need `Protocol::Http/Https` though.

MathiasKoch:
> I didn't understand why we need Protocol::Http/Https though.

I don't think we do, and that's kinda my point on the one remark i have left, that i wanted your feedback on tackling..
But we do need some way of defining whether to connect using http(s) or ws(s) somehow?

tekjar:
Oh I see. I think what you mean to add in protocol enum is `Protocol::Tcp`. Not `Protocol::Http`. Since we are replacing transport, I would change `Protocol` to `Transport`.

(I will bring my plan to add QUIC as transport in future in the below example)

```
enum Transport {
    Tcp,
    Tls(ca, Option<(client_certificate, client_key)>,
    Ws,
    Wss(ca, Option<(client_certificate, client_key)>,
    Quic(ca, Option<(client_certificate, client_key)>,
}
```

```
MqttOptions.set_transport(&mut self, transport: Transport) {}
```

This is one way to make transport modular

MathiasKoch:
That would work great for me 👍 i will give it a shot tomorrow, and update my pr

MathiasKoch:
I think that turned out quite elegantly!
I also changed a bit of TLS config handling, and keys to prevent some possible user errors at compile-time.

tekjar:
Thankyou for your contribution :)

PR #208, opened 2020-12-19
use pretty_env_logger in broker.rs example

MariusVB:


tekjar:
Why is this `use` necessary?

PR #209, opened 2020-12-20
Async Broker init

qm3ster:
First 3 commits are fixes

Last one adds async version of locallink, with a runnable usage example


qm3ster:
closing in favor of #228

PR #210, opened 2020-12-22
Add broker prefix to config TOML

MariusVB:


tekjar:
Sorry for the delay. Existing broker config should not be changed because embedded broker binary expects the config that way. `broker.` should be prefixed only if broker config is a "part" of the app config

https://github.com/bytebeamio/rumqtt/commit/b638ffbaf26330e0f16ac8672ba493421e527e6f

This commit in should fix the example

PR #215, opened 2020-12-30
Make console optional.

qwandor:
If no `[console]` section is provided in the config then the console server will not be started. This allows a more minimal configuration when the console is not needed.


tekjar:
Thanks!!

PR #216, opened 2020-12-31
Log info rather than error when a client disconnects cleanly.

qwandor:


qwandor:
The test failure looks unrelated, can you re-run the tests in case it's a flake?

Issue #217, opened 2021-01-04
Async client example does not compile


detly:
If I create a project with the rumqttc example code and try to compile it (under tag R3 or head of master), rumqttc fails to compile with:

```plain
   Compiling rumqttc v0.3.0 (https://github.com/detly/rumqtt.git?tag=R3#b8688681)
error[E0308]: mismatched types
   --> /home/jason/.cargo/git/checkouts/rumqtt-38df926bebc19038/b868868/rumqttc/src/eventloop.rs:292:40
    |
292 |             Network::new(WsStream::new(socket), options.max_incoming_packet_size)
    |                                        ^^^^^^ expected struct `async_tungstenite::WebSocketStream`, found struct `WebSocketStream`
    |
    = note: expected struct `async_tungstenite::WebSocketStream<_>`
               found struct `WebSocketStream<async_tungstenite::stream::Stream<TokioAdapter<tokio::net::TcpStream>, TokioAdapter<tokio_rustls::client::TlsStream<tokio::net::TcpStream>>>>`
    = note: perhaps two different versions of crate `async_tungstenite` are being used?

error[E0277]: the trait bound `WsStream<_>: AsyncRead` is not satisfied
   --> /home/jason/.cargo/git/checkouts/rumqtt-38df926bebc19038/b868868/rumqttc/src/eventloop.rs:292:26
    |
292 |             Network::new(WsStream::new(socket), options.max_incoming_packet_size)
    |                          ^^^^^^^^^^^^^^^^^^^^^ the trait `AsyncRead` is not implemented for `WsStream<_>`
    | 
   ::: /home/jason/.cargo/git/checkouts/rumqtt-38df926bebc19038/b868868/rumqttc/src/framed.rs:23:29
    |
23  |     pub fn new(socket: impl N + 'static, max_incoming_size: usize) -> Network {
    |                             - required by this bound in `Network::new`
    |
    = note: required because of the requirements on the impl of `N` for `WsStream<_>`

error[E0277]: the trait bound `WsStream<_>: AsyncWrite` is not satisfied
   --> /home/jason/.cargo/git/checkouts/rumqtt-38df926bebc19038/b868868/rumqttc/src/eventloop.rs:292:26
    |
292 |             Network::new(WsStream::new(socket), options.max_incoming_packet_size)
    |                          ^^^^^^^^^^^^^^^^^^^^^ the trait `AsyncWrite` is not implemented for `WsStream<_>`
    | 
[...]

PR #218, opened 2021-01-04
Add example of how to use rumttc wih TLS.

### LINKED ISSUE #214: Add an example for TLS usage
It would be very nice if there was an example to show how to use rustls-native-certs or webpki-roots to configure rumqttc to use TLS with a "regular" (non-self-signed) certificate. Maybe this could even be built-in.

qwandor:
This example loads the system root certiifcates with rustls-native-certs.

Fixes #214.


qwandor:
The build failure appears to be the same error as #217.

tekjar:
Hmm..Did some dependent crate have semver issues?

PR #220, opened 2021-01-05
There shouldn't be any need to sleep between polling the EventLoop.

qwandor:
I can't see any reason why this example has a sleep in the polling loop, it will just slow everything down.


tekjar:
Test failure looks unrelated. I've raised a separate issue to fix tests which are infrequently failing

Issue #206, opened 2020-12-18
How to handle faliures in subscribing?


mgritzmacher:
Hi. Thanks for creating this library. I'm running into some problems trying to add error handling with my use of the rumqttc client.

When I tell the client to subscribe to something I know the topic, but I don't know what pkid the client will use
```rust
client.subscribe(topic, QoS::AtLeastOnce).await
```
when I get an Event for an outgoing packet, I only have the pkid
```
Ok(Outgoing(Subscribe(4)))
```
when I get an acknowledgement, I get the pkid and if each topic was successful or unsucessful, but I've lost the topic
```
Ok(Incoming(SubAck(SubAck { pkid: 4, return_codes: [Success(AtLeastOnce)] })))
```

I can see that the Subscribe type for Incoming packets https://docs.rs/rumqttc/0.3.0/rumqttc/struct.Subscribe.html has the topic information and pkid in presumably the same order as for Incoming SubAck, but there's no matching topic information on the outgoing Subscribe variant https://docs.rs/rumqttc/0.3.0/rumqttc/enum.Outgoing.html#variant.Subscribe

I want to connect the information together so that I can try to resubscribe to topics that failed, but I can't see how to do this with the public API of the library at the moment.

tekjar:
Yeah I think it makes sense to add subscription filters to `Outgoing::Subscribe`. But this can lead to the rabbit hole of adding topics for `Outgoing::Publish` and some other metadata for other packets.

I think the deeper problem is that the client doesn't have synchronous mode yet. Does your subscriptions happen during initialization or is it a complicated logic? If it's during initialization, you can simulate synchronous client like this

```
client.subscribe();
for event in connection.iter() {
    /// wait until suback packet
}
```

Few other folks asked for synchronous mode before. I'll probably implement it in January

mgritzmacher:
In my particular use case I'm using the client asyncronously and my subscriptions are quite complicated. I've got a many to one relationship with services that tell the client to do things, and a central service that polls the event loop, so trying to control the order of things (in order to know which ack corresponds to which subscribe) would be challenging.

tekjar:
Ok. I'll add filters to `Outgoing::Subscribe` for now. Let me know if you want to implement this. This should be straight forward

mgritzmacher:
I've made a pull request that works for my needs with error handling, let me know if it needs any adjustments.

PR #225, opened 2021-01-07
Expose outgoing topics in public API on publish, subscribe and unsubscribe

mgritzmacher:
This provides enough information for the topic and pkid to be linked together as they are outgoing, so code monitoring the events from the EventLoop can link together the pkid on incoming PubAck, SubAck and UnsubAck packets with the topic from the outgoing packet, so it can respond to errors for particular topics as needed. See https://github.com/bytebeamio/rumqtt/issues/206 for context.

### DIFF
--- a/rumqttc/src/lib.rs
+++ b/rumqttc/src/lib.rs
@@ -121,12 +121,12 @@ pub type Incoming = Packet;
 /// Current outgoing activity on the eventloop
 #[derive(Debug, Eq, PartialEq, Clone)]
 pub enum Outgoing {
-    /// Publish packet with packet identifier. 0 implies QoS 0
-    Publish(u16),
-    /// Subscribe packet with packet identifier
-    Subscribe(u16),
-    /// Unsubscribe packet with packet identifier
-    Unsubscribe(u16),
+    /// Publish packet with packet identifier and topics. 0 implies QoS 0
+    Publish(u16, Vec<String>),
+    /// Subscribe packet with packet identifier and topics
+    Subscribe(u16, Vec<String>),
+    /// Unsubscribe packet with packet identifier and topics
+    Unsubscribe(u16, Vec<String>),
     /// PubAck packet
     PubAck(u16),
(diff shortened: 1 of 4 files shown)


rumqttc/src/lib.rs
```diff
@@ -121,12 +121,12 @@ pub type Incoming = Packet;
...
 #[derive(Debug, Eq, PartialEq, Clone)]
 pub enum Outgoing {
-    /// Publish packet with packet identifier. 0 implies QoS 0
-    Publish(u16),
-    /// Subscribe packet with packet identifier
-    Subscribe(u16),
-    /// Unsubscribe packet with packet identifier
-    Unsubscribe(u16),
+    /// Publish packet with packet identifier and topics. 0 implies QoS 0
+    Publish(u16, Vec<String>),
```
qwandor, line 125: Maybe it would be more readable to give these fields names now that there is more than one? Like
```
Publish{
  packet_id: u16,
  topics: Vec<String>,
}
```

mgritzmacher: Yep that makes sense. I've gone with `pkid` for parity with the named fields on the incoming structs.

rumqttc/src/lib.rs
```diff
@@ -121,12 +121,21 @@ pub type Incoming = Packet;
...
-    /// Publish packet with packet identifier. 0 implies QoS 0
-    Publish(u16),
-    /// Subscribe packet with packet identifier
-    Subscribe(u16),
-    /// Unsubscribe packet with packet identifier
-    Unsubscribe(u16),
+    /// Publish packet with packet identifier and topics. 0 implies QoS 0
+    Publish {
+        pkid: u16,
+        topics: Vec<String>,
```
tekjar, line 127: Why does publish have a list of topics instead of 1 topic?

mgritzmacher: I didn't realise publish can only have one. I've made it a single topic now.

PR #227, opened 2021-01-13
client: Move websocket support behind a Cargo feature

### LINKED ISSUE #226: Put websocket transport behind a Cargo feature
Currently, depending on rumqttc means depending on async-tungstenite and ws_stream_tungstenite transitively and compiling them as part of ones own project. A lot of use cases don't need this, so it would be nice to be able to disable it.

This would be a breaking change since it changes the API under `default-feature = false` (even though that previously did nothing, one could still do it).

Would you welcome a PR that implements this?

jplatte:
I haven't added the new `websocket` feature to the default features, if you want I can update the PR to do that.

Resolves #226.

### DIFF
--- a/rumqttc/Cargo.toml
+++ b/rumqttc/Cargo.toml
@@ -9,15 +9,20 @@ edition = "2018"
 keywords = ["mqtt", "iot", "coap", "http"]
 categories = ["network-programming"]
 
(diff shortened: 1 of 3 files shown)


tekjar:
I believe not having websockets is the correct default. Thanks!! :)

PR #232, opened 2021-01-22
Fix compilation on Windows once more

skreborn:
**Woop-woop!** That's the sound of da *Windows* police!

The [`pprof` crate](https://lib.rs/crates/pprof) is unfortunately [not compatible with Windows](https://github.com/tikv/pprof-rs/issues/9) as of now, and thus its inclusion prevents the library from being built even if the relevant code is never accessed. It is only "used" in [`rumqttd/src/bin.rs`](https://github.com/bytebeamio/rumqtt/blob/master/rumqttd/src/bin.rs), and even then it really isn't, so it is not what you would necessarily call [essential](https://i.imgur.com/0WXFHtw.png).

This change simply gets rid of the offending piece of code and its dependency. Rejoice, for this lets the library be built on Windows once more!

### DIFF
--- a/rumqttd/Cargo.toml
+++ b/rumqttd/Cargo.toml
@@ -32,7 +32,6 @@ pretty_env_logger = "0.4"
 bytes = "1.0"
 prost = "0.6"
 warp = "0.2"
-pprof = { version = "0.3", features = ["flamegraph", "protobuf"] }
 ctrlc = "=3.1"
 futures-util = "0.3.8"
 
--- a/rumqttd/Cargo.toml
+++ b/rumqttd/Cargo.toml
@@ -37,3 +37,4 @@ futures-util = "0.3.8"
 
 [target.'cfg(not(target_env = "msvc"))'.dependencies]
 jemallocator = "0.3"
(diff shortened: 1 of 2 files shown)


tekjar:
Instead of removing, can we instead feature gate pprof with `!cfg mvsc`?

tekjar:
I thing a `profile` feature flag also makes sense

skreborn:
> Instead of removing, can we instead feature gate pprof with `!cfg mvsc`?

That would've been my first approach, but since it was commented out and not doing anything, I've instead opted to remove it completely. If you'd like, I can put it back (either commented or uncommented), but being unable to compile and run the binary with profiling on, I wasn't feeling very confident just uncommenting it, fearing that it might break on a Linux.

Is there a reason it was commented out?

tekjar:
I think it was commented out by mistake while I was checking throughput difference with and without profiling on

skreborn:
I've re-added the profiling code for non-Windows targets. I've tested it and it builds and runs successfully on both Windows and Linux - and generates the profiling data successfully on the latter.

tekjar:
Thanks!! :)

PR #234, opened 2021-01-22
Cleanup

tekjar:


rumqttd/src/lib.rs
```diff
@@ -181,13 +187,22 @@ impl Broker {
...
             let console = Arc::new(console);
-            consolelink::start(console);
+            runtime.spawn(async {
+                consolelink::start(console).await;
+            });
         }
 
+        runtime.block_on(async {
+            signal::ctrl_c().await.unwrap();
+        });
```
skreborn, line 204: Waiting for CTRL+C really shouldn't be part of the server - especially in a non-configurable way. At the very least, it should be toggle-able, but I would much rather the responsibility of waiting and blocking was on the embedder. As of now, it interferes with a SIGINT handler that the embedder sets up and silently swallows it.

tekjar: Yeah I agree that this is not great. But I need to find out if it's possible to output profiling information after ctrl + c

PR #239, opened 2021-01-26
Rumqttc/no panics and improved performance

mneumann:
* No longer panics when a PubAck, PubComp or PubRec paket is received with
  a `pkid` outside of the inflight range. 

* Avoid cloning `Publish` messages with QoS=1 or QoS=2. Also avoid cloning a `Publish` in case of a collision. Cloning is rather expensive as all the memory has to be duplicated (topic and payload).

* Use a bitset for `incoming_pub`. This drops memory from 256 KiB to just 8 KiB. Also use a bitset for `outgoing_rel`, but as this is limited to just `max_inflight` the saving are usually much less.

* Factor out code that maintains the list of currently inflight Publish messages into `ds.rs`. I have some code that makes the data-structures interchangeable, but as that required some generic traits I have not included it here.

### DIFF
--- a/rumqttc/Cargo.toml
+++ b/rumqttc/Cargo.toml
@@ -29,6 +29,7 @@ async-channel = "1.5"
 log = "0.4"
 thiserror = "1.0.21"
 http = "^0.2"
+fixedbitset = "0.3"
 
 [dev-dependencies]
 pretty_env_logger = "0.4"
--- a/rumqttc/src/ds.rs
+++ b/rumqttc/src/ds.rs
@@ -1,0 +1,155 @@
+//! Data structures to track outgoing `Publish` messages as well as a set of `pkid`s.
+
+use crate::Publish;
+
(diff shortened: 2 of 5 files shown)


tekjar:
Thanks!! These are a good list of changes. I'll run these through benchmark suite

tekjar:
Hi. Sorry for the delay on this. I was in a vacation. I'm on a time crunch and I might not get enough time to properly test these changes. Can you create smaller PRs with a subset of features?

Other option is to wait until I have some time in hand to test these properly

mneumann:
@tekjar Hi, yes I think that PR grew too big. I will have to split this into several independent PRs. Most importantly for me is to get away with `panics` in case of out-of-range packet ids. But also the unneccessary cloning of `Publish` is something that I want to go away. I think, the first 4 commits will fit into a separate PR. Then I will make a PR that will use the same data structure as `rumqttc` is using right now, but that will fix the panics.

Issue #236, opened 2021-01-23
Runtime error on wrong private key format (TLS)


jaredwolff:
If the keys are not in the correct format this line causes a runtime error:

```
2021-01-23T04:25:16Z DEBUG rustls::anchors] add_pem_file processed 1 valid and 0 invalid certs
thread 'main' panicked at 'removal index (is 0) should be < len (is 0)', library/alloc/src/vec.rs:1057:13
```

Looks like it's related to this line.

```
server_config.set_single_cert(certs, keys.remove(0))?;
```

My keys looks like it would work with the `pkcs8_private_keys` function as it begins and ends with:

```
-----BEGIN PRIVATE KEY-----
...
-----END PRIVATE KEY-----

```

jaredwolff:
Looks like my keys were actually the correct format for `rsa_private_keys`. Just the wrong headers. Funky.

```
[2021-01-23T04:58:35Z DEBUG rustls::anchors] add_pem_file processed 1 valid and 0 invalid certs
[2021-01-23T04:58:35Z INFO  librumqttd] Waiting for connections on 0.0.0.0:8883. Server = 1
```

tekjar:
Would you like to send a PR with a check before `remove`?

jaredwolff:
Yup a PR is coming your way at some point. Also have some improvements related to the router loop that doesn't yield. ;)

PR #241, opened 2021-01-28
Handling error cases if the key is parsed but is not valid.

jaredwolff:
The cause can be something as simple as not having RSA in -----BEGIN RSA PRIVATE KEY-----. It applies both to the client
and server side. Mentioned this in #236


tekjar:
Thanks!! :)

PR #242, opened 2021-01-28
Prevents starving threads of cpu time

jaredwolff:
Prevents starving threads/tasks of cpu time.

Since it's a for loop that doesn't yield, it prevents any computation from happening in any other task/thread. This is vitally important for multithread apps where there are multiple tasks running side-by-side.

### DIFF
--- a/rumqttlog/src/router/mod.rs
+++ b/rumqttlog/src/router/mod.rs
@@ -30,6 +30,8 @@ pub enum Event {
     Disconnect(Disconnection),
     /// Get metrics of a connection or all connections
     Metrics(MetricsRequest),
+    /// Gets fired off when the connect is ready for more processing
+    ConnectionReady,
 }
 
 /// Requests for pull operations
--- a/rumqttlog/src/router/router.rs
+++ b/rumqttlog/src/router/router.rs
@@ -52,6 +52,8 @@ pub struct Router {
     /// replicators. Each connection will have a tx handle which they use
     /// to send data and requests to router
     router_rx: Receiver<(ConnectionId, Event)>,
+    /// Channel sender for the above reciever
+    router_tx: Sender<(ConnectionId, Event)>,
     /// Aggregates for all connections
     metrics: RouterMetrics,
 }
@@ -91,10 +93,21 @@ impl Router {
             data_waiters,
             topics_waiters,
             router_rx,
+            router_tx: router_tx.clone(),
             metrics,
         };
 
         (router, router_tx)
+    }
+
+    /// Waits on incoming events when the readyqueue.
+    pub async fn async_run(&mut self) -> Result<(), RouterError> {
+        loop {
+            match self.router_rx.async_recv().await {
+                Ok((id, data)) => self.route(id, data),
+                Err(RecvError) => return Err(RouterError::Disconnected),
+            }
+        }
     }
 
     /// Waits on incoming events when ready queue is empty.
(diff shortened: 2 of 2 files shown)


tekjar:
Hi @jaredwolff. I'm trying to understand this PR. There is already a connection ready [event](https://github.com/bytebeamio/rumqtt/pull/242/files#diff-d145c414bfbde5f712bb2e7f2b2335f3e3dac3171e77a38d2e6da8431c6a0981R26) to process more data when the connection is ready. 


> Since it's a for loop that doesn't yield, it prevents any computation from happening in any other task/thread. This is vitally important for multithread apps where there are multiple tasks running side-by-side.

Can you explain this a bit? When the router doesn't have any more data to process or when the connections are not ready to process more data, it'll just be idle without consuming any cpu.

tekjar:
The way readiness works is

1. Router ties to put  data into connection channel using connection's [notify](https://github.com/bytebeamio/rumqtt/blob/master/rumqttlog/src/router/connection.rs#L89)
2. When there is only one empty slot in the channel, that empty slot is filled with `Notification::Pause`
3. When connection sees `Notification::Pause` after processing preceding data, is send `Event::Ready` to router to inform that it is ready to process more data 

Is there any bug that led you create this PR?

jaredwolff:
> Can you explain this a bit? When the router doesn't have any more data to process or when the connections are not ready to process more data, it'll just be idle without consuming any cpu.

I think that's your intention. Which of the calls in `start` (in router.rs) yields when not being used? `self.readyqueue.is_empty()` doesn't yield. `self.router_rx.try_recv()` doesn't yield. It simply tries to receive and continues on. `self.readyqueue.pop_front()` doesn't yield either.

Essentially, when there's nothing to do it becomes a very tight infinite loop which doesn't yield to any other working threads/tasks. I discovered the issue when trying to run it along side all the other threads in [pyrinas](https://github.com/pyrinas-iot/pyrinas-server-rs). Once it got to initializing the MQTT code, the MQTT stuff worked great but everything else came to a screeching halt. 😅

I run into this problem all the time in the embedded world. Threads have to coexist nicely with each other otherwise nothing happens like you thing it should.

jaredwolff:
Also I'll put a bug together for this but that's the gist of it.

tekjar:
When you say `yield`, I'm assuming `yield to os`. 

https://github.com/bytebeamio/rumqtt/blob/master/rumqttlog/src/router/router.rs#L108

That line should block. Are you seeing 100% cpu in the router when there is nothing to process?

jaredwolff:
Doesn't necessarily need to be 100% cpu. Correct me if I'm wrong but `self.router_rx.recv` on it's own is not async. In order to yield the thread you'd have to call the async version of that which would be self.router.rx_recv().await no?

Drilling down further, looks like the synchronous `recv()` isn't much different. The lower level implementation looks like it uses `thread::park` which is essentially a blocking wait. i.e. it doesn't yield to other tasks. 

Don't take all my word for this. Throw the start() function in a Tokio::spawn() along with some other concurrent tasks. You'll see exactly what I'm talking about.

tekjar:
Oh I get your concerns now. Yeah, router is not designed asynchronous. It's supposed to run in a system thread, not tokio task (which is cooperative, not preemtive like system thread). Alternate if you want to use tokio is to use `spawn_blocking`. But it'll just spawn a background thread implicitly.

tekjar:
Feel free to reopen if you have more concerns. Gist of the discussion is that, `router start` is a synchronous blocking function. You shouldn't use it in tokio tasks.

PR #243, opened 2021-01-28
Make profiler a feature, not target.cfg

qm3ster:


tekjar:
@skreborn This PR uses non-default  `prof` feature instead of ` #[cfg(not(target_env = "msvc"))]`. This should work for you right?

skreborn:
@tekjar Thanks for checking in! Should be alright, but it definitely needs to be documented that it's not available on Windows. Since it's non-default, however, at least it shouldn't interfere with the build process unless you specifically set a feature, and that should be easy to roll back for any developer if they happen to run into the issue.

qm3ster:
Since https://github.com/rust-lang/cargo/issues/1197#issuecomment-268203727 is not even available on nightly cargo, there isn't much better I could do. Do you feel like profiler support even needs to be a default feature on supported platforms? `cargo install` does support `--features` for instance, if this is an end user thing after all. And if you distribute binaries, your build can be however complex.

I do feel like `rumqttd` and `librumqttd` should be separate crates though, as all of the below is bin-specific:
```toml
[features]
prof = ["pprof"]

[dependencies]
argh = "0.1.3"
confy = "0.4.0"
pretty_env_logger = "0.4"
pprof = { version = "0.4", features = ["flamegraph", "protobuf"], optional = true }

[target.'cfg(not(target_env = "msvc"))'.dependencies]
jemallocator = "0.3"
```
Tell me if this is something you'd be ok with, so that I know to take a smacc at it whenever I can.



@tekjar also, are any of these necessary?
```toml
[lib]
name = "librumqttd"
path = "src/lib.rs"

[[bin]]
name = "rumqttd"
path = "src/bin.rs"
```
this should be enough to maintain the current behavior (by moving `src/bin.rs` to `src/bin/rumqttd.rs`):
```toml
[lib]
name = "librumqttd"
```

skreborn:
> I do feel like `rumqttd` and `librumqttd` should be separate crates though, as all of the below is bin-specific:

@qm3ster @tekjar I agree that they should be separate - mostly because of the dependencies they don't share, but also out of utility. There's not a lot of scenarios where one might need both. You either need - for any given project - and embeddable or standalone server.

PR #247, opened 2021-02-06
First attempt at getting to compile for both tokio as well as async-std

Frans-Willem:
Hi,

I'm more of a fan of async-std than tokio, and needed an MQTT client library. I've put most of the changes behind features such that you can choose to compile with support for either async-std or tokio.

Currently helpers.rs is a bit of a mess that I'd like to clean up a little, any guidance on this would be appreciated.

### DIFF
--- a/rumqttc/Cargo.toml
+++ b/rumqttc/Cargo.toml
@@ -14,21 +14,32 @@ all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
-websocket = ["async-tungstenite", "ws_stream_tungstenite"]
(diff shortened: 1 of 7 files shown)


tekjar:
Hi @Frans-Willem . Appreciate your efforts on this. Unfortunately I don't have the bandwidth to maintain multiple runtimes. I'll be happy to support async std as soon as I find a co-owner/maintainer

Frans-Willem:
No worries, I'll just make it a fork for now, that I'll try to keep up to date with your original, and if in the future there's reason to merge it we can still do so at that time :)

PR #248, opened 2021-02-09
Ensure that outgoing publish has a valid `pkid`

mneumann:
Ensures that an outgoing publish message has always a valid `pkid` set, according to the MQTT spec.

### DIFF
--- a/rumqttc/src/state.rs
+++ b/rumqttc/src/state.rs
@@ -290,7 +290,8 @@ impl MqttState {
 
     /// Adds next packet identifier to QoS 1 and 2 publish packets and returns
     /// it buy wrapping publish in packet
-    fn outgoing_publish(&mut self, publish: Publish) -> Result<(), StateError> {
+    fn outgoing_publish(&mut self, mut publish: Publish) -> Result<(), StateError> {
+        self.assign_pkid(&mut publish);
         let publish = match publish.qos {
             QoS::AtMostOnce => publish,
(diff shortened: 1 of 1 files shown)


tekjar:
qos 0 publish won't consider packet id during serialization. There is no need to set packet id to 0 explicity

mneumann:
@tekjar This is the 2nd and 3rd commit of PR #239. I'll break that PR into smaller pieces.

tekjar:
Sure. But are these changes necessary? As far as I see, we don't need to set pkid to 0.

 If this is so that you don't have to rework those commits in #239, I can merge and fix this later

mneumann:
@tekjar Hi tekjar, sorry for the delay. I thought #239 is too big to get accepted as one piece so I started to split them into separate parts. This is the first part. But if you merge #239, you can ignore this PR completely. #239 contains everything in this PR (and much more).

You are right, there is no "need" to set it to `pkid` as you have special code in another crate to deal with it, but it does not do any harm either. I think it's cleaner to explicitly set to 0 , so we do not accidentially access it somewhere. So, yes, you are right, it's not strictly neccessary to set it to 0, but then it depends on some assumptions from `mqttbytes` to be correct. Better safe than sorry.

tekjar:
Now serialization and deserialiaztion are tightly integrated into rumqttc instead of separate crate (for ease of maintenance). I feel it's ok to special case

Issue #257, opened 2021-02-27
`rustls` does not support less secure (yet widely used) cipher suites


jaredwolff:
Some IoT devices do not run the recommended cipher suites in industry. The rational was described in [this writeup here](https://github.com/ctz/rustls/blob/c9838230b0d20938cf4849f1cc6a8a771f6c8a95/rustls/src/manual/tlsvulns.rs). Despite not being in TLS1.3 anymore, many devices don't support these newer cipher suites. 

Would an acceptable option be using something like `tokio-native-tls` with a feature to switch between the two? I think in most cases, this will provide the coverage. 

For example here's the cipher suite support list  for the nRF9160 ([source](https://www.nordicsemi.com/Products/Low-power-cellular-IoT/nRF9160/Download#infotabs)):

```
TLS-ECDHE-ECDSA-WITH-AES-256-CBC-SHA384
TLS-ECDHE-ECDSA-WITH-AES-256-CBC-SHA   
TLS-ECDHE-ECDSA-WITH-AES-128-CBC-SHA256
TLS-ECDHE-ECDSA-WITH-AES-128-CBC-SHA  
TLS-ECDHE-RSA-WITH-AES-256-CBC-SHA         
TLS-ECDHE-RSA-WITH-AES-128-CBC-SHA256    
TLS-ECDHE-RSA-WITH-AES-128-CBC-SHA       
TLS-PSK-WITH-AES-256-CBC-SHA      
TLS-PSK-WITH-AES-128-CBC-SHA256   
TLS-PSK-WITH-AES-128-CBC-SHA
TLS-PSK-WITH-AES-128-CCM-8
TLS-EMPTY-RENEGOTIATIONINFO-SCSV
```

`rustls` doesn't support any of these. Only ran in to the issue when `rumqttd` barfed that there was no supported cipher suites.

jaredwolff:
This is addressed for now with native-tls functionality. In order for `rumqttd` to be secure, a MQTT password is required since there is no TLS client authentication

PR #258, opened 2021-02-28
rumqttd: adding optional native-tls support.

jaredwolff:
By using the `use-native-tls` feature, this crate can now use tokio-native-tls vs tokio-rustls. Discussed this in #257. Turns out it was trivial to add since `tokio` uses the same API for both packages. The only thing that seems like a bummer right now is that it requires use of a `pkcs12` cert rather than 3 separate ones.

### DIFF
--- a/.github/workflows/features.yml
+++ b/.github/workflows/features.yml
@@ -21,6 +21,18 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: test
+          args: --release --no-default-features
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --release --features use-rustls
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
+          args: --release --no-default-features --features use-native-tls
+      - uses: actions-rs/cargo@v1
+        with:
+          command: test
           args: --release --all-features
 
 
--- a/rumqttd/Cargo.toml
+++ b/rumqttd/Cargo.toml
@@ -18,13 +18,15 @@ name = "rumqttd"
 path = "src/bin.rs"
 
 [features]
+default = ["use-rustls"]
 prof = ["pprof"]
+use-rustls = ["tokio-rustls"]
+use-native-tls = ["tokio-native-tls"]
 
 [dependencies]
 rumqttlog = { path = "../rumqttlog", version = "0.6"}
 mqttbytes = { path = "../mqttbytes", version = "0.3" }
 tokio = { version = "1.0", features = ["full"] }
-tokio-rustls = "0.22"
 serde = { version = "1", features = ["derive"] }
 log = "0.4"
 thiserror = "1"
@@ -35,6 +37,10 @@ bytes = "1.0"
 warp = "0.3"
 futures-util = "0.3.8"
 pprof = { version = "0.4", features = ["flamegraph", "protobuf"], optional = true }
+
+# Optional
+tokio-rustls = { version = "0.22", optional = true }
+tokio-native-tls = { version = "0.3", optional = true }
 
 [target.'cfg(not(target_env = "msvc"))'.dependencies]
 jemallocator = "0.3"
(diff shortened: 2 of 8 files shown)


jaredwolff:
@tekjar looks like CI needs openssl or a way to build it.

tekjar:
Thanks @jaredwolff . I'll try to find some time. I'm a bit occupied with other work for the whole month

jaredwolff:
@tekjar no worries. I'm kicking the tires here to make sure it's working ok.

jaredwolff:
Looks like this is running `cargo test` with `--all-features` will bork the build. Created separate builds for (default) rustls and native-tls. Wasn't sure if there was a better way to do it.

tekjar:
This is a very useful addition. Thanks @jaredwolff 

Cargo features are only additive. In this case, I think we are trying to create mutually exclusive features. This is a [problem](https://github.com/rust-lang/cargo/issues/2980). If this library is included in the dependency tree multiple times, using `native-tls` feature is going to globally enable `native-tls` (overwriting other dependency's wish to use rustls)

What we need is a single feature flag `native-tls` and the ability to dynamically choose `nativetls` or `rustls` at runtime

jaredwolff:
> (overwriting other dependency's wish to use rustls)

Is this really the case? I thought if `rustls` is 'disabled' in this crate, for example, you can still use it outside. I could be wrong here.

`reqwest` comes to mind as a pertinent example. The package gives you the choice to either use the default (`native-tls`) or `rustls`. Sean did something similar in his [CI setup](https://github.com/seanmonstar/reqwest/blob/master/.github/workflows/ci.yml) by building the mutually exclusive features separately.

> What we need is a single feature flag native-tls and the ability to dynamically choose nativetls or rustls at runtime

It seems like a runtime option could increase the amount of unused cruft that you need to compile when building the `rumqttd` package. If you layer `rumqttd` on top of other projects it gets hefty fast. Right now one of my sub projects is 350+ dependencies. This is more of a developer quality of life thing more than anything but it does matter for folks!

tekjar:
I'll check reqwest in depth tomorrow but reqwest features are still additive

```
cargo build --all-features
cargo test --all-features
```

builds and works fine



> Is this really the case? I thought if `rustls` is 'disabled' in this crate, for example, you can still use it outside. I could be wrong here.

Yeah. This is a well know problem. [Here](https://github.com/rust-lang/cargo/issues/4328#issuecomment-628988567) is a good explanation. In short, multiple dependencies using rumqtt library and enabling different features shouldn't cause compilation errors or silently switch tls implementation underneath

jaredwolff:
> I'll check reqwest in depth tomorrow but reqwest features are still additive

Hmm ok! I'll look into more how he did it as well so I can modify the PR as necessary.

> Here is a good explanation. In short, multiple dependencies using rumqtt library and enabling different features shouldn't cause compilation errors or silently switch tls implementation underneath

Agh! Ok now I get it. It doesn't necessarily matter about other deps using a `rustls` or `native-tls` it's whether or not someone is using `rumqtt` with both features. I guess there hypothetically be a need to use `rustls` and `native-tls` at the same time. Looking at the [resulting documentation](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification) this is clicking now! Thanks @tekjar

jaredwolff:
@tekjar updated the code so it works in all 4 cases:

* No features
* Rustls
* Native TLS
* Both enabled

I'm going to kick the tires a bit and will push any other changes I find to make this work.

tekjar:
@jaredwolff  This is nice. Do you think we'll need both the feature flags? I think it's fine to leave rustls as is and have just one feature flag to enable native-tls. 

What you've done makes sense when you want disable rustls and enable native-tls. In our usecase, we usually want a rustls server on a port always but also run native-tls on a different port for microcontrollers that can't work with rustls

jaredwolff:
> Do you think we'll need both the feature flags?

In my use case it's handy. I don't use Rustls anywhere else since I use a separate ACME reverse proxy for web stuff.

tekjar:
Cool. We'll merge as is then. This work is very useful for us as well. Thanks for your contribution

jaredwolff:
Awesome @tekjar! I have to make some small tweaks but those will come in another PR. 😎

PR #262, opened 2021-03-11
Replace hard coded 0.0.0.0 bind of broker and console with configuration option `listen`

flxo:
The address that the broker and console binds is hardcoded to the v4 any address 0.0.0.0. In order to operate the broker on
a dedicated address or a v6 ip this needs to be added to the configuration. Replace the lonely `port` setting in the configuration with a `std::net::SocketAddr`.

There are host and port configs in `rumqttlog/timestmp.toml` and `rumqttlog/mesh.toml` which I couldn't identify. Looks like outdated stuff.

### DIFF
--- a/rumqttd/config/rumqttd.conf
+++ b/rumqttd/config/rumqttd.conf
@@ -13,7 +13,7 @@ max_connections = 10001
 
 # Configuration of server and connections that it accepts
 [servers.1]
-port = 1883
+listen = "0.0.0.0:1883"
 next_connection_delay_ms = 1
     [servers.1.connections]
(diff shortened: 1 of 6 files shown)


flxo:
The CI seems to be stuck somewhere? Is there something I can do?

tekjar:
Hi @flxo . Apologies for the delay. I had to stop our self hosted runner. The problem should be fixed now

flxo:
> Hi @flxo . Apologies for the delay. I had to stop our self hosted runner. The problem should be fixed now

No worries. I rebased. Let's see what happens ;-)

PR #271, opened 2021-04-02
Just some v5 stuff

qm3ster:


michalfita:
I like this change, the code looks leaner.

PR #272, opened 2021-04-20
V5 disconnect packet reason code and properties

rfael:
I added to `mqttbytes` V5 protocol Disconnect Reason Codes and Disconnect Properties as described in [documentation](https://docs.oasis-open.org/mqtt/mqtt/v5.0/csprd01/mqtt-v5.0-csprd01.html#_Toc489530226)

### DIFF
--- a/mqttbytes/src/v5/connack.rs
+++ b/mqttbytes/src/v5/connack.rs
@@ -49,7 +49,7 @@ impl ConnAck {
     }
 
     fn len(&self) -> usize {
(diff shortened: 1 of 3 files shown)


tekjar:
Thanks!! :)

qm3ster:
Very nice.
However, I think you should have introduced a `Error::InvalidDisconnectReturnCode` instead of using `Error::InvalidConnectReturnCode`.

PR #275, opened 2021-05-18
mqttbytes v5 disconnect make disconnect packet fields public

rfael:
I forgot about this in my last PR :p 

However  @qm3ster commented my last PR:
> Very nice.
> However, I think you should have introduced a `Error::InvalidDisconnectReturnCode` instead of using `Error::InvalidConnectReturnCode`.

It seems to me it would be a good idea to rename `Error::InvalidConnectReturnCode` to `Error::InvalidReturnCode` because this error is used in other packets, not only in Connect. What do you think?

### DIFF
--- a/mqttbytes/src/v5/disconnect.rs
+++ b/mqttbytes/src/v5/disconnect.rs
@@ -128,10 +128,10 @@ pub struct DisconnectProperties {
 #[derive(Debug, Clone, PartialEq)]
 pub struct Disconnect {
     /// Disconnect Reason Code
-    reason_code: DisconnectReasonCode,
+    pub reason_code: DisconnectReasonCode,
 
     /// Disconnect Properties
-    properties: Option<DisconnectProperties>,
+    pub properties: Option<DisconnectProperties>,
 }
 
 impl DisconnectProperties {


qm3ster:
If it's already is, then probably. But ideally, it would be a nested enum.
`Error::InvalidReturnCode(Operation::Disconnect)`, anyone?

rfael:
@qm3ster So considering this, every error should contain information about the packet/operation in which it occurs?
In my opinion it's not necessary.

qm3ster:
@rfael I understand what you mean, that at the point of `fn Disconnect::read` that is superfluous information. However, it is often used through `fn v5::read`:
https://github.com/bytebeamio/rumqtt/blob/044d8124551758e81c01a847ff14a82cdb2cfa3c/mqttbytes/src/v5/mod.rs#L118-L153
at which point it becomes completely opaque, we lose track of what packet had an unsupported/unknown/invalid reason.
Perhaps it doesn't matter, because the specification forbids it. And it's definitely not something that we can fix at runtime.
Perhaps `log:error!` should be used in all `impl TryFrom<u8> for XReasonCode`?

tekjar:
We've moved mqttbytes out of this repo. It exists [here](https://github.com/bytebeamio/mqttbytes) now. Please reopen a new PR if you still need this

Issue #254, opened 2021-02-24
[Broker] Auth connection


cyrixSpec:
Hi there,
is it possile to implement auth (username, password) on the broker-side for incoming connections?

Thanks in advance.

tekjar:
Hi @cyrixSpec . Apologies for the delay. Yeah that should be easy enough but unfortunately I don't have the bandwidth now. But I can help with implementation and PRs.

jaredwolff:
@cyrixSpec should be added now. Give it a whirl and let us know if you run into any snags.

PR #283, opened 2021-06-02
Adding server side password check.

jaredwolff:
Changed:
* Created InvalidServerPass error for pkcs12 password check failure.
* Checking credentials on connection if needed.
* Changed login in ConnectionSettings to be a list rather than a single set of username/password

Added:
* Additional error condition if no valid user/password was found
* Added validate function for Login credentials in codec.
* Checking authentication failure list on initial connection to boot it if it's too soon.
* Propagates device address and AuthenticationAttempts list to Connector & RemoteLink
* Adding or clearing connection from AuthenticationAttempts list depending on password check success
* Setting max lockout duration to 24h (prevent overflow)

### DIFF
--- a/mqttbytes/src/v4/connect.rs
+++ b/mqttbytes/src/v4/connect.rs
@@ -267,6 +267,10 @@ impl Login {
 
         connect_flags
     }
+
+    pub fn validate(&self, username: &String, password: &String ) -> bool {
+        (self.username == *username) && (self.password == *password)
+    }
 }
 
 
--- a/rumqttd/src/lib.rs
+++ b/rumqttd/src/lib.rs
@@ -83,6 +83,8 @@ pub enum Error {
     InvalidCACert(String),
     #[error("Invalid server cert file {0}")]
     InvalidServerCert(String),
+    #[error("Invalid server pass")]
+    InvalidServerPass(),
     #[error("Invalid server key file {0}")]
     InvalidServerKey(String),
     RustlsNotEnabled,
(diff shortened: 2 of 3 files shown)


jaredwolff:
Still working on this. Addresses #254 

One thing @tekjar, would it make sense to have a `validate` function within the codec like I have here? The password/username is set private so it's not accessible outside the Login struct. I'll change it however works for you.

Also, I was thinking of implemented a back-off mechanism for invalid authentication attempts. I was thinking of putting it in this PR or in a separate. 

All thoughts are welcome!

jaredwolff:
Added the backoff bits so as long as the server doesn't get reset, someone only has one time every 24 hours to attempt to brute force a password (from the same IP). Feedback on this is appreciated. I figured having some type of backoff algorithm is better than none!

tekjar:
Thanks! I'll review this in a few days

jaredwolff:
@tekjar do you see any areas for improvement? It'd be great to get this into `rumqttd`.

tekjar:
@jaredwolff If I understand this correctly we are using exponential backoff per connection when authentication has failed. While this is fine,  I think global connection throttling is good enough. I would like to reduce the maintenance complexity at this moment. Can we change this PR to just include authentication without exponential backoff?

tekjar:
Thanks @jaredwolff

jaredwolff:
@tekjar when do you think you'll roll a new release?

PR #287, opened 2021-06-10
Various Refactorings

jplatte:
Most of these are based on clippy suggestions.

### DIFF
--- a/benchmarks/clients/rumqttsync.rs
+++ b/benchmarks/clients/rumqttsync.rs
@@ -45,7 +45,7 @@ pub fn start(id: &str, payload_size: usize, count: usize) -> Result<(), Box<dyn 
     }
 
     let elapsed_micros = start.elapsed().as_micros();
-    let throughput = (acks_count * 1000_000) / elapsed_micros as usize;
+    let throughput = (acks_count * 1_000_000) / elapsed_micros as usize;
 
     // --------------------------- results ---------------------------------------
 
--- a/benchmarks/clients/rumqttsync.rs
+++ b/benchmarks/clients/rumqttsync.rs
@@ -35,6 +35,10 @@ pub fn start(id: &str, payload_size: usize, count: usize) -> Result<(), Box<dyn 
 
     let mut acks_count = 0;
     let start = Instant::now();
+
+    // flatten isn't shorter because the if let is needed anyways, and
+    // flattening a result is not an entirely obvious operation either.
+    #[allow(clippy::manual_flatten)]
     for event in connection.iter() {
         if let Ok(Event::Incoming(Incoming::PubAck(_))) = event {
             acks_count += 1;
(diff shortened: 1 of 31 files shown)


jplatte:
Ping @tekjar

tekjar:
@jplatte Thank you. This repo is going through design changes internally. I'll incorporate these changes in the next major release

jplatte:
Ping @tekjar, are things still moving?

de-sh:
Hey @jplatte, sorry for the delay from our side. We had been unable to focus on rumqtt for a while due to certain unforeseen circumstance. I have spoken with @tekjar and will be taking over the responsibility of reviewing and merging your PR from him.

jplatte:
The only changes that are unrelated to clippy are the first (manual refactoring, just for better readability) and third (clearly) commits. The only ones that could have an impact on runtime performance are using `&str` instead of `&String` (removal of an unnecessary pointer indirection) and removing unnecessary clones (removes unnecessary work, though likely just for the optimizer rather than the resulting program).

Compile time should not be affected in a meaningful way; readability / maintainability should be equivalent or improved for all of the changes.

rumqttlog/src/logs/acks.rs
```diff
@@ -38,7 +38,7 @@ impl Acks {
         match qos {
             1 => self.acks.push(Packet::PubAck(PubAck::new(pkid))),
             2 => self.acks.push(Packet::PubRec(PubRec::new(pkid))),
-            _ => return,
+            _ => {}
```
de-sh, line 41: We might have missed a oxford `,` here?

benchmarks/clients/rumqttsync.rs
```diff
@@ -35,8 +35,8 @@ pub fn start(id: &str, payload_size: usize, count: usize) -> Result<(), Box<dyn
 
     let mut acks_count = 0;
     let start = Instant::now();
-    for event in connection.iter() {
-        if let Ok(Event::Incoming(Incoming::PubAck(_))) = event {
+    for event in connection.iter().flatten() {
```
de-sh, line 38: I had not known about `flatten()`, nice usage. Should we add a comment atop or will this be self-explanatory?

de-sh (changes requested):
@jplatte I have added a few comments, but otherwise this looks ready to merge!

rumqttlog/src/logs/acks.rs
jplatte: Commas after `{}` in match arms qre removed by rustfmt by default.

benchmarks/clients/rumqttsync.rs
jplatte: I can add a comment.

jplatte: Actually I'll just revert this change, after thinking more about it I don't like it anymore.

jplatte:
Looks like something broke that's unrelated to this PR. I'll try rebasing to see whether that helps.

de-sh (approved):
LGTM

de-sh:
Thanks @jplatte 🍾

PR #293, opened 2021-06-29
let username and password fields in connect message in public

railwaycat:
I'm trying to use mqttbytes as a library in my C project though Rust FFI, but username and password in Connect::Login as private fields prevent they be visited through FFI. 

Set the username and password fields as pub is straight forward and we have all other packet struct fields as pub too.


railwaycat:
The Github Actions error looks like a common build error under Linux? I can't link it to this change.

tekjar:
Thanks! This is a simple PR. So I'll merge this with build error and then fix them today or tomorrow

PR #297, opened 2021-07-20
Fix benchmarks build issue: Changed prost::Message to pprof::protos::…

klispap:
This small change will fix the build issues of benchmark crates for R8 and master.
It will allow Issue #292 to pass the pipeline as well.


tekjar:
Thanks @klispap

PR #300, opened 2021-08-08
Reconnection policy algorithms

dghilardi:
Hi, I implemented a way to define custom reconnection policies, maybe could be useful to others too.

In order to give the ability to define a re-connection policy I've defined a new trait `ReconnectionStrategy` and nested in the event-loop.
If a reconnection strategy is defined for the eventloop, before connecting it will await the instant given by that reconnection strategy, it will also invoke callbacks for connection established / lost events.
I also implemented that trait for a truncated exponential backoff following google guidelines: https://cloud.google.com/storage/docs/retry-strategy

### DIFF
--- a/benchmarks/clients/rumqttasync.rs
+++ b/benchmarks/clients/rumqttasync.rs
@@ -24,7 +24,7 @@ pub async fn start(id: &str, payload_size: usize, count: usize) -> Result<(), Bo
     mqttoptions.set_keep_alive(20);
     mqttoptions.set_inflight(100);
 
-    let (client, mut eventloop) = AsyncClient::new(mqttoptions, 10);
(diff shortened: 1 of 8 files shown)


tekjar:
Thanks @dghilardi for your contribution. This repo is going through a massive revamp in private. I would like keep things simple at this point until I have more time in my hand

de-sh:
Hey @dghilardi, thanks for the PR. It seems like this is a lot of complexity that we don't feel will be very maintainable and for this reason alone we don't see any reason to merge this PR right now. But we also feel that your effort could be very helpful to folks who want to include the feature in their own implementations and hence would encourage you to close the PR and instead [start a discussion item](https://github.com/bytebeamio/rumqtt/discussions/new) to link to your code.

Thanks!

PR #308, opened 2021-09-22
Add unix domain socket support

### LINKED ISSUE #307: Support for unix socket transport
We use an mqtt broker as a convenient IPC mechanism where it is confined to the machine it is running on.  We'd really prefer to use a unix socket as the transport, rather than a tcp socket, given that we don't need any of the tcp reliability stuff.  I'd be happy to do the work if this would be a welcome addition.

xedrac:
Adds support for unix domain sockets.  Resolves #307.


tekjar (approved):

tekjar:
Thanks! :)

Issue #294, opened 2021-06-29
Missing tokio features: "rt", "macros", "io-util"


MightyPork:
Building `rumqttc` v0.8.0 without the tokio features "rt", "macros", and "io-util" results in compile errors. The crate does not ask for them but then uses them in code.

Workaround is to add this to the project's Cargo.toml:

```
tokio = { version = "1.0", features = ["rt", "macros", "io-util"] }
```

I imagine most people don't notice this because everyone uses the "full" tokio feature, but for non-tokio applications this is a problem.

Errors:

```
error[E0432]: unresolved import `tokio::runtime`
  --> /home/ondra/.cargo/registry/src/github.com-1ecc6299db9ec823/rumqttc-0.8.0/src/client.rs:10:5
   |
10 | use tokio::runtime;
   |     ^^^^^^^^^^^^^^ no `runtime` in the root

error[E0432]: unresolved import `tokio::runtime`
  --> /home/ondra/.cargo/registry/src/github.com-1ecc6299db9ec823/rumqttc-0.8.0/src/client.rs:11:12
   |
11 | use tokio::runtime::Runtime;
   |            ^^^^^^^ could not find `runtime` in `tokio`

error[E0432]: unresolved import `tokio::select`
  --> /home/ondra/.cargo/registry/src/github.com-1ecc6299db9ec823/rumqttc-0.8.0/src/eventloop.rs:10:5
   |
10 | use tokio::select;
   |     ^^^^^^^^^^^^^ no `select` in the root

error[E0432]: unresolved imports `tokio::io::AsyncReadExt`, `tokio::io::AsyncWriteExt`
 --> /home/ondra/.cargo/registry/src/github.com-1ecc6299db9ec823/rumqttc-0.8.0/src/framed.rs:4:28
  |
4 | use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
  |                            ^^^^^^^^^^^^              ^^^^^^^^^^^^^ no `AsyncWriteExt` in `io`
  |                            |
  |                            no `AsyncReadExt` in `io`
  |
help: a similar name exists in the module
  |
4 | use tokio::io::{AsyncRead, AsyncRead, AsyncWrite, AsyncWriteExt};
  |                            ^^^^^^^^^
help: a similar name exists in the module
  |
4 | use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWrite};
  |                                                      ^^^^^^^^^^

[...]

PR #311, opened 2021-09-25
Add missing tokio features: rt, macros, io-util

didier-wenzek:
This is required by any crate that depends on `rumqttc` but not `tokio`.

See #294.


tekjar:
Thanks! :)

PR #312, opened 2021-09-30
Change set_keep_alive to use Duration instead of u16

shikoku1:
Since most libraries use `Duration` to express timeouts, I propose to change the `set_keep_alive` function to consume `Duration`


tekjar:
Thanks!

Issue #306, opened 2021-09-20
Release a new version to crates.io with the latest changes


Thoralf-M:
@tekjar would it be possible to do a new release to crates.io with the latest changes? Would be great because otherwise we can't release a new version for our crate

tekjar:
Just released a new version. Can you please check and close this issue?

Thoralf-M:
Hey @tekjar, thanks for doing it, but unfortunately this commit https://github.com/bytebeamio/rumqtt/commit/b0b1d1228a01a6b28858c4039f4ca163c1e73060 breaks it since `UnixStream` only works with the `net` feature of tokio and on Unix, but we want to use it also on other platforms 😢

Thoralf-M:
Thanks!

PR #317, opened 2021-10-07
Add unix feature

Thoralf-M:
Without this feature it works only on unix platforms
Related to #306


tekjar:
Thanks! Will merge this on Monday

PR #319, opened 2021-10-13
Only include UnixSteam on unix-like platforms

Archipel:
Rumqttc was unable to build on Windows before this change


tekjar:
https://github.com/bytebeamio/rumqtt/pull/317/files

@Archipel Can you check if that PR solves your problem?

IniterWorker:
@tekjar you can close this one. https://github.com/bytebeamio/rumqtt/pull/317/files is the same thing much cleaner.

PR #320, opened 2021-10-13
TryFrom<Url> impl for MqttOptions

### LINKED ISSUE #255: Possibility start client by specifying mqtt broker address using URI
It would be really helpful if `MqttOptions.new()` would also accept URI instead of the current host/port pair.

This makes it a lot simpler to load the broker address via `std::env::var(...)`:
```rust
let mut mqttoptions = MqttOptions::new("rumqtt-async", std::env::var("MQTT_URI")?)?;
```

JeanMertz:
I noticed #255 and had a need for this feature as well, to integrate mqtt into [Vector](https://github.com/vectordotdev/vector).

The implementation is based on https://github.com/mqtt/mqtt.org/wiki/URI-Scheme, but extended with query support to enable full configuration of the client through the URI.

closes #255

### DIFF
--- a/rumqttc/Cargo.toml
+++ b/rumqttc/Cargo.toml
@@ -29,6 +29,7 @@ async-channel = "1.5"
 log = "0.4"
 thiserror = "1.0.21"
 http = "^0.2"
+url = { version = "2.2", default-features = false, optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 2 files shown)


tekjar:
Thanks @JeanMertz! Could you give more context how you are using rumqtt and why URIs are more convinient?

PR #321, opened 2021-10-18
Improve CI testing

EdJoPaTo:
this should prevent mistakes

windows fails should be fixed by #317.
Its not clear to my why macOS fails.

See running example: https://github.com/EdJoPaTo/rumqtt/actions/workflows/features.yml


de-sh:
Thanks for the contribution @EdJoPaTo 🍾

PR #325, opened 2021-11-21
Fix keep_alive_secs parameter when converting from URL in MqttOptions

belltoy:
As the code changes.

The `keep_alive_secs`  URL query parameter should be removed (taken) from the `queries` HashMap when trying to convert to `MqttOptions`.


belltoy:
@tekjar

Issue #323, opened 2021-11-05
It should be rendered possible to match against a CertificateExpired error


shikoku1:
Right now, if a connection is attempted with an expired certificate, the error returned from `poll()` is very deep and tls::Error is private

`Network(Io(Custom { kind: InvalidData, error: AlertReceived(CertificateExpired) }))`

It should be rendered possible to match against CertificateExpired

de-sh:
Hey @shikoku1, looks like this issue is resolved, hence closing, please reopen if not.

PR #326, opened 2021-12-08
Make tls::Error public in rumqttc

shikoku1:
Right now it's impossible to use / match against `rumqttc::tls::Error`, therefore I propose this change
Realted to #323


tekjar:
Thanks @shikoku1 :)

PR #336, opened 2022-01-05
rumqttc: update rustls to 0.20

RAvenGEr:
Update rumqttc to use version 0.20 of rustls and update all dependent crates.

### DIFF
--- a/rumqttc/examples/tls.rs
+++ b/rumqttc/examples/tls.rs
@@ -1,8 +1,8 @@
 //! Example of how to configure rumqttd to connect to a server using TLS and authentication.
 
 use rumqttc::{self, AsyncClient, Event, Incoming, MqttOptions, Transport};
-use std::error::Error;
 use rustls::ClientConfig;
+use std::error::Error;
 
 #[tokio::main]
 async fn main() -> Result<(), Box<dyn Error>> {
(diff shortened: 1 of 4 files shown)


rumqttc/examples/tls.rs
```diff
@@ -13,11 +13,19 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
-    // To customise TLS configuration we create a rustls ClientConfig and set it up how we want.
-    let mut client_config = ClientConfig::new();
     // Use rustls-native-certs to load root certificates from the operating system.
-    client_config.root_store =
-        rustls_native_certs::load_native_certs().expect("Failed to load platform certificates.");
+    let mut root_cert_store = rustls::RootCertStore::empty();
+    for cert in rustls_native_certs::load_native_certs().expect("could not load platform certs") {
+        root_cert_store
+            .add(&rustls::Certificate(cert.0))
+            .unwrap();
```
de-sh, line 21: Seems like this could just return on err with `?` instead of panicking.

de-sh (commented):
Thanks for the PR @RAvenGEr, I have added a comment, otherwise this looks good to go 😄

rumqttc/examples/tls.rs
RAvenGEr: Yes, the ? is suitable there. The end result doesn't really change, but the code looks a little cleaner.

de-sh:
LGTM 👍🏼

PR #341, opened 2022-02-04
Support for manual acks 

wiktorsikora:
This PR introduces support for manual acknowledgements.
With `MqttOptions.manual_acks = true` every message must be manually with `client.ack()` method.
By using manual acknowledgements users are able to guarantee that they will not loose any messages.


tekjar:
LGTM. @de-sh Please review as well

de-sh:
LGTM 👍🏼

PR #342, opened 2022-02-04
Remove `must_use` attribute over `poll()`

de-sh:
Currently the `must_use` attribute is being applied to the `Future` instead of `Eventloop::poll()` itself.


de-sh:
Closing this as the proposed solution does not properly solve the problem at hand.

PR #365, opened 2022-03-06
Make rustls optional

### LINKED ISSUE #305: Pure Rust support (make rustls optional)
`rumqttc` has the dependency chain `tokio-rustls` => `rustls` => `ring` => `cc`.

This makes builds of `rumqttc` fail in pure Rust environments (no C compiler).
Since `rumqttc` advertises itself in Cargo as a pure Rust client, it would be nice to make TLS support a feature that can be disabled to ensure pure Rust support.

de-sh:
Closes #305

Enables the compilation of `rumqttc` without rustls/TLS support by using `--no-default-features` flag during compilation. `rustls` support is enabled by default, otherwise can be enabled with `use_rustls` feature flag.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1685,7 +1685,6 @@ dependencies = [
  "tokio",
  "tokio-rustls",
  "url",
- "webpki",
  "ws_stream_tungstenite",
 ]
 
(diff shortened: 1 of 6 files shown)


eutampieri:
Shouldn't the `webpki` crate be optional too?

de-sh:
> Shouldn't the `webpki` crate be optional too?

Thanks for the pointer, this dependency is no longer relied upon and can be removed, but I made `rustls_pemfile` optional.

Issue #252, opened 2021-02-23
ConnAck with non-success code being transformed to ConnectionError::Io is very confusing


jplatte:
When trying to catch auth errors to add special error handling for them, the first thing I reached for was splitting a previous

```rust
Event::Incoming(Incoming::ConnAck(_)) => // ...
```

match pattern into

```rust
Event::Incoming(Incoming::ConnAck(ConnAck {
    code: ConnectReturnCode::Success,
    ..
} => // success
Event::Incoming(Incoming::ConnAck(ConnAck { code, .. })) => {
    // failed with code
}
```

but was surprised that the latter reached when authorization failed. After looking at what error got generated, I now updated the `.poll().await` return handling to

```rust
Err(ConnectionError::Io(io)) => {
    let error_msg = match io.kind() {
        ErrorKind::ConnectionRefused => {
            "Failed to connect to MQTT (connection refused)"
        }

        _ if io.to_string().ends_with("NotAuthorized") => {
            "Failed to connect to MQTT (not authorized)"
        }

        // ...
    };

    // ...
}
```

But that seems super brittle. Please just stop turning `ConnAck`s with `code != Success` into errors at all, or at least turn them into a separate `ConnectionError::Server(ConnectReturnCode)` variant or something like that.

jplatte:
I'm no longer using rumqttc. I can ping my former colleagues though, maybe sbd. wants to verify this.

FSMaxB:
Will check this out once I find the time to do so.

FSMaxB:
Sorry for taking such a long time :sweat_smile:.

I have now verified that it is fixed, thanks a lot!

PR #370, opened 2022-03-13
Reorganize `ConnectionError`

de-sh:
Create new variants for `ConnectionError` instead of repackaging as `io::Error`

Solves #252

### DIFF
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -2,7 +2,6 @@ use crate::{framed::Network, Transport};
 use crate::{tls, Incoming, MqttState, Packet, Request, StateError};
 use crate::{MqttOptions, Outgoing};
 
-use crate::mqttbytes;
 use crate::mqttbytes::v4::*;
 use async_channel::{bounded, Receiver, Sender};
 #[cfg(feature = "websocket")]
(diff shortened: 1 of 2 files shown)


rumqttc/src/eventloop.rs
```diff
@@ -29,14 +28,17 @@ pub enum ConnectionError {
     MqttState(#[from] StateError),
     #[error("Timeout")]
     Timeout(#[from] Elapsed),
-    #[error("Packet parsing error: {0}")]
-    Mqtt4Bytes(mqttbytes::Error),
```
de-sh, line 33: This variant is not being used anywhere, hence decided to get it removed.

rumqttc/src/eventloop.rs
```diff
@@ -29,14 +28,17 @@ pub enum ConnectionError {
...
-    #[error("Network: {0}")]
-    Network(#[from] tls::Error),
+    #[cfg(feature = "websocket")]
+    #[error("Websocket: {0}")]
+    Websocket(#[from] async_tungstenite::tungstenite::error::Error),
+    #[error("TLS: {0}")]
+    Tls(#[from] tls::Error),
     #[error("I/O: {0}")]
     Io(#[from] io::Error),
-    #[error("Stream done")]
```
de-sh, line 38: Same here

Issue #372, opened 2022-03-15
"attempt to subtract with overflow" error on MqttState struct.


Ticsmtc:
seems program try to subtract **inflight** when it is 0 . maybe try to make it **Atomic** ?

```rust
thread 'tokio-runtime-worker' panicked at 'attempt to subtract with overflow', /src/state.rs:288:17
```

here is the code cause the problem : 

```rust
use tokio::{task, time};
use rumqttc::{self, AsyncClient, MqttOptions, QoS};
use std::error::Error;
use std::sync::Arc;
use std::time::Duration;
use std::thread;
use log::info;

fn main() {
    pretty_env_logger::init();

    let mut mqttoptions = MqttOptions::new("test-1", "localhost", 1883);
    mqttoptions.set_keep_alive(Duration::from_secs(5));

    let (client, mut eventloop) = AsyncClient::new(mqttoptions, 10);


    let client = Arc::new(client);
    let client2 = Arc::clone(&client);

    thread::spawn(move || {
       loop {
           pollster::block_on(client.publish("hello/world",QoS::ExactlyOnce,true,"hello"));
           thread::sleep(Duration::from_millis(500));
       }
    });

    thread::spawn(move || {
        loop {
            pollster::block_on(client2.publish("hello/world",QoS::ExactlyOnce,true,"hello"));
            thread::sleep(Duration::from_millis(500));
        }
    });


    let rt = tokio::runtime::Builder::new_multi_thread().worker_threads(2).enable_all().build().unwrap();
    rt.spawn(async move {
        loop {
            let res = eventloop.poll().await;
            info!("res {:?}",res);
        }
    });

    thread::sleep(Duration::from_secs(1200));
}
```

Ticsmtc:
not sure it is network issue or the code , close by now .

PR #373, opened 2022-03-15
Fix MqttState inflight data-race

Ticsmtc:
fix MqttState inflight data-race bug
remove Clone trait on MqttState because there is no use.
modify some related tests


Ticsmtc:
related to this issue #372

PR #375, opened 2022-03-16
Limit GitHub actions to running only on PRs affecting code/dependencies

de-sh:
Only changes in files with paths `**.rs` and `*Cargo.*`


de-sh:
All uninteresting commits will still trigger CI if they come after an interesting commit

de-sh:
reverting/removing changes to `*.rs` files will stop CI getting triggered

de-sh:
The CI was not triggered most probably because of the wrong use of regex to match `Cargo.toml`s

Issue #337, opened 2022-01-20
Please include `Cargo.lock` for Nix package-ability


colemickens:
Hi,

For packaging in Nix, we need the `Cargo.lock` to be available. My understanding is that Cargo already ignores the lock files if it exists, for anyone using `mqttd` as a library, so it shouldn't be a problem.

Thanks!

PR #378, opened 2022-03-18
Add `native-tls` support

mladedav:
There are no tests involved including whether stuff compiles with or without any given feature. Docs are also missing. I will need a bit of guidance with both.

I also have not tried WebSockets with native TLS and I have reason to believe it might not work since there are features specific for TLS libraries in `async-tungstenite`.

Other than that I have tried compiling this manually with either `use-rustls`, `use-native-tls, with neither, and with both. I have also used this against Azure IoT Hub so I know that it works.


mladedav:
By the way I have also deleted Cargo.lock (hence the -2574 changes). From what I understood this file is not supposed to be part of libraries, only bin crates. I can revert that if it should be kept here for some reason.

de-sh:
Cargo.lock was included because of #337. We will start with not supporting `websocket`+`native-tls`, let them be separate for now. This looks like it's going in the right direction. I will take a look at what we can do to write documentation and tests. Thanks for opening the PR, let's merge this!

de-sh (approved):
LGTM. Made some minor changes here: https://github.com/bytebeamio/rumqtt/commit/ca41afd2ed03d0b44c9126f973dc61867864967e

mladedav:
I have checked the pipeline and the issues stem from rustls being optional now. TLS examples do not compile without rustls.

I would like to wait for #365 and then tweak anything that's left with native-tls.

de-sh (changes requested):
For secure websockets you can limit that with a `#[cfg(all(feature = "websocket", feature = "use-rustls"))]`. I'll try and devote some time to review this further tomorrow. Thanks.

mladedav:
It should now all work. I have recently discovered `cargo hack` and `cargo udeps` so I was playing with it a bit and some stuff I was changing ended up in the commit by accident.

All feature powersets should now work (although arguably websockets + native-tls does not make much sense).

rumqttc/src/tls.rs
```diff
@@ -1,34 +1,59 @@
...
     Io(#[from] io::Error),
+    #[cfg(feature = "use-rustls")]
     #[error("Web Pki")]
     WebPki(#[from] webpki::Error),
+    #[cfg(feature = "use-rustls")]
     #[error("DNS name")]
     DNSName(#[from] InvalidDnsNameError),
+    #[cfg(feature = "use-rustls")]
     #[error("TLS error")]
     TLS(#[from] rustls::Error),
```
de-sh, line 51: We could name this Rustls(...)

de-sh (approved):
LGTM 👍🏼

rumqttc/src/tls.rs
mladedav: I would be in favor of that, but I wanted to keep the features consistent with rumqttd. I'll leave to your judgement whether you want to change the names just here, in both or in neither.

mladedav:
hey @de-sh, can this be merged now? Should I make some changes or is this waiting for some internal processes?

de-sh:
Hey @mladedav, sorry for the delayed response. We have hit a roadblock while testing this PR.

There are two main issues:
- The lack of a way to define custom pkcs12 certificate and password, I have worked on adding this [here](https://github.com/bytebeamio/rumqtt/commit/ef0db8352fad362e4884d6a4faf3f5d35abd1b7c)
- Tests with and without custom TLS options are failing:
  - Using the default TLS connector, i.e implemented in this PR until now:
    ```
    Error = Tls(NativeTls(Ssl(Error { code: ErrorCode(1), cause: Some(Ssl(ErrorStack([Error { code: 337047686, library: "SSL routines", function: "tls_process_server_certificate", reason: "certificate verify failed", file: "../ssl/statem/statem_clnt.c", line: 1913 }]))) }, X509VerifyResult { code: 19, error: "self signed certificate in certificate chain" })))
    ```
  - Using custom TLS certificate that were created with the same CA as the server gives us:
    ```
    Error = Tls(NativeTls(Ssl(Error { code: ErrorCode(1), cause: Some(Ssl(ErrorStack([Error { code: 337047686, library: "SSL routines", function: "tls_process_server_certificate", reason: "certificate verify failed", file: "../ssl/statem/statem_clnt.c", line: 1913 }]))) }, X509VerifyResult { code: 19, error: "self signed certificate in certificate chain" })))
    ```

mladedav:
Ok, thanks for the response. If you need anything from me do not hesitate to ask.

mladedav:
Hi @de-sh. I have noticed that R13 was released. Can I do something to unblock this?

There were some issues with custom certificates which you have already solved, right? And then there were some issues with tests? Should I try and rebase the pull request on latest main to have this merged?

mladedav:
Here is the usage: https://gist.github.com/mladedav/490857cc3419dbf56b59589d465d42e1

I haven't manually created any certificates, I believe native-tls should use your OS store if not configured to do differently. It works out of the box both on my system and in Docker rust image with `docker run -it -v ${PWD}:/src -w /src rust cargo run` in the directory with the project.

mladedav:
@de-sh can I ask what happened here?

de-sh:
This seems to have been a mistake on the part of GitHub.

de-sh:
We will manage future developments on this in the separate branch, we are still blocked on a few issues, hopefully they will be resolved in the near future.

mladedav:
Ok, thanks. If there's anything I can do to help please let me know.

PR #385, opened 2022-04-16
Improve Error texts

EdJoPaTo:
Initially I wanted to just add details for the `mqttbytes::Error` in the `StateError`. I continued with other `thiserror::Error` implementations to improve error details.

Something that is not the same everywhere is the formatting style: `text: {0}`, `text {0}` or `text = {0}` are used currently. I could also improve that with this PR but I am not sure with format is preferred. Maybe `text: {0}`?

### DIFF
--- a/benchmarks/simplerouter/src/protocol/mod.rs
+++ b/benchmarks/simplerouter/src/protocol/mod.rs
@@ -68,7 +68,7 @@ pub enum Error {
     InvalidReason(u8),
     #[error("Invalid protocol used")]
     InvalidProtocol,
-    #[error("Invalid protocol level")]
+    #[error("Invalid protocol level = {0}")]
     InvalidProtocolLevel(u8),
     #[error("Invalid packet format")]
     IncorrectPacketFormat,
--- a/rumqttc/src/mqttbytes/mod.rs
+++ b/rumqttc/src/mqttbytes/mod.rs
@@ -4,7 +4,7 @@
 //! The [`bytes`](https://docs.rs/bytes) crate is used internally.
 
 use bytes::{Buf, BufMut, Bytes, BytesMut};
-use core::fmt::{self, Display, Formatter};
+use core::fmt;
 use std::slice::Iter;
 
 mod topic;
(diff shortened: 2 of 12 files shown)


rumqttc/src/mqttbytes/mod.rs
```diff
@@ -13,7 +13,7 @@ pub mod v4;
 pub use topic::*;
 
 /// Error during serialization and deserialization
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
```
de-sh, line 16: Adding `#[error("text: {0}")]` attributes above each variant, might be a good idea?

rumqttd/src/lib.rs
```diff
@@ -86,13 +86,14 @@ pub enum Error {
...
     #[error("Invalid server pass")]
-    InvalidServerPass(),
+    InvalidServerPass,
     #[error("Invalid server key file {0}")]
     InvalidServerKey(String),
     RustlsNotEnabled,
     NativeTlsNotEnabled,
     Disconnected,
     NetworkClosed,
+    #[error("Wrong packet {0:?}")]
```
de-sh, line 96: I agree that `text: {0}` is a better format for error messages, this could change here.

de-sh, line 96: I agree that `text: {0}` is a better format for error messages, this could change here.

de-sh (commented):
This looks great, I just have a few questions

rumqttc/src/mqttbytes/mod.rs
EdJoPaTo: I was too lazy to add the text everywhere and checked out how it looks:

`Error = PayloadTooLong` with the name from the code wasnt too bad. Its better to write it manually but this way its already an improvement.

In can add all the `#[error("text: {0}")]` descriptions tho as it looks nicer then.

rumqttd/src/lib.rs
EdJoPaTo: Alright, I will migrate all error messages towards the "text: {0}" format. Probably later today.

rumqttc/src/mqttbytes/mod.rs
EdJoPaTo: I plan on writing the messages tomorrow

EdJoPaTo: Not sure about all the messages. There might be a mistake somewhere as I tried to guess what it might mean sometimes.

de-sh (approved):
LGTM

de-sh:
Congrats and thank you on the merge @EdJoPaTo 🍾

rumqttc/src/mqttbytes/mod.rs
```diff
@@ -13,33 +13,50 @@ pub mod v4;
...
 pub enum Error {
+    #[error("Expected Connect, received: {0:?}")]
     NotConnect(PacketType),
+    #[error("Unexpected Connect")]
     UnexpectedConnect,
+    #[error("Invalid Connect return code: {0}")]
     InvalidConnectReturnCode(u8),
-    InvalidReason(u8),
```
EdJoPaTo, line 21: looking at the final diff…

Not sure what happened here but are some cases missing now? It still compiles so they weren’t used?

benchmarks/simplerouter/src/protocol/mod.rs
```diff
@@ -68,7 +68,7 @@ pub enum Error {
     InvalidReason(u8),
     #[error("Invalid protocol used")]
     InvalidProtocol,
-    #[error("Invalid protocol level")]
+    #[error("Invalid protocol level = {0}")]
```
EdJoPaTo, line 71: I think we missed noticing this `text:` vs `text =` thingy.

feel free to just change it on the default branch or create a PR yourself, that’s way quicker.

rumqttc/src/mqttbytes/mod.rs
de-sh: I removed them as they were deemed unnecessary

benchmarks/simplerouter/src/protocol/mod.rs
de-sh: Let's keep it like that

PR #391, opened 2022-04-29
Improve MqttOptions creation

EdJoPaTo:
Improvements:
- typed-builder
- ClientId struct
- deprecated old `MqttOptions::new` for easier migration

I dont think its that bad to expose the MqttOptions to the public. The types itself prevent a lot of wrong content (maybe types like NonZeroU16 could improve that even further). I refactored the client id check into `ClientId` so it is checked by itself.

There are still a few open questions.
- I set the `broker_addr` and `port` to `localhost` and `1883`. Especially the port sounds sane to me. Should this be kept?
- It would be nice to have some generator for ClientId. Like `rumqttc-<systemtime>` or so to connect easier to some broker when the client id does not need to be something fix. But this should be a separate PR.
- The content of the options is now completely public. Should be rename some of them now to something better or should we keep them like that?
- `MqttState::new` got the panic which was previously in `MqttOptions::set_inflight`. Personally I think there shouldnt be any panic in a library (only impossible ones like `unreachable!()`). Should we change that behaviour there towards `Result<MqttState, SomeError>` or keep the panic for now?

I am open for feedback here. When only part of the improvements is liked I can create another PR with only that part.

### DIFF
--- a/rumqttc/src/state.rs
+++ b/rumqttc/src/state.rs
@@ -80,6 +80,11 @@ impl MqttState {
     /// connection for persistent sessions while new state should
     /// instantiated for clean sessions
     pub fn new(max_inflight: u16, manual_acks: bool) -> Self {
+        // TODO: Maybe std::num::NonZeroU16 might be neat here?
+        if max_inflight == 0 {
+            panic!("zero inflight is not allowed");
+        }
+
         MqttState {
             await_pingresp: false,
             collision_ping_count: 0,
@@ -505,7 +510,7 @@ mod test {
     use super::{MqttState, StateError};
     use crate::mqttbytes::v4::*;
     use crate::mqttbytes::*;
-    use crate::{Event, Incoming, MqttOptions, Outgoing, Request};
+    use crate::{Event, Incoming, Outgoing, Request};
 
     fn build_outgoing_publish(qos: QoS) -> Publish {
         let topic = "hello/world".to_owned();
@@ -773,8 +778,6 @@ mod test {
     #[test]
     fn outgoing_ping_handle_should_throw_errors_for_no_pingresp() {
         let mut mqtt = build_mqttstate();
-        let mut opts = MqttOptions::new("test", "localhost", 1883);
-        opts.set_keep_alive(std::time::Duration::from_secs(10));
         mqtt.outgoing_ping().unwrap();
 
         // network activity other than pingresp
@@ -795,9 +798,6 @@ mod test {
     #[test]
     fn outgoing_ping_handle_should_succeed_if_pingresp_is_received() {
         let mut mqtt = build_mqttstate();
-
-        let mut opts = MqttOptions::new("test", "localhost", 1883);
-        opts.set_keep_alive(std::time::Duration::from_secs(10));
 
         // should ping
         mqtt.outgoing_ping().unwrap();
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1717,6 +1717,7 @@ dependencies = [
  "thiserror",
  "tokio",
  "tokio-rustls",
+ "typed-builder",
  "url",
  "ws_stream_tungstenite",
 ]
@@ -2341,6 +2342,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "59b11b2b5241ba34be09c3cc85a36e56e48f9888862e19cedf23336d35316ed1"
 dependencies = [
  "memchr",
+]
+
+[[package]]
+name = "typed-builder"
+version = "0.10.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "89851716b67b937e393b3daa8423e67ddfc4bbbf1654bcf05488e95e0828db0c"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
 ]
 
 [[package]]
--- a/benchmarks/clients/rumqttasync.rs
+++ b/benchmarks/clients/rumqttasync.rs
@@ -21,9 +21,13 @@ async fn main() {
 }
 
 pub async fn start(id: &str, payload_size: usize, count: usize) -> Result<(), Box<dyn Error>> {
-    let mut mqttoptions = MqttOptions::new(id, "localhost", 1883);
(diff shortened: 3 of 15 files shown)


de-sh:
A lot of the tests fail since the `url` dependency is optional, only enabled with `features = ["url"]`

EdJoPaTo:
> Edit after the CI reminded me: How to handle the code example in the docs which fails without the url feature?

Yeah, I know. But I am not sure how I want to handle that well and am open for ideas there.

EdJoPaTo:
> A lot of the tests fail since the `url` dependency is optional, only enabled with `features = ["url"]`

Maybe more details there: The `parse_url` method had a `#[cfg(feature = "url")]`. I migrated that towards an `impl FromStr` which also as the feature check there. The documentation that this exists was on the `parse_url` method which worked fine. As the documentation is kinda useless for the `impl FromStr` method (the user will never see it) I moved it to the main struct. But then there is no feature check there. We could mark these code examples as `ignore` but then they arent tested anymore (even when the feature is enabled). These code examples could also be added somewhere else to ensure they are working but its easy for them to get out of sync. This might be a simple way to do it but its definitely not idea.

We could also write the docs twice and use conditional compilation to decide which docs to use. Then the code will only be there when the feature is enabled and only tested in that case. But I think its only possible when the docs are in a file somewhere?

de-sh:
I tried doing this:
```
rumqttc/src/lib.rs --- Rust
367 ///                                                                          367 ///                                                                         
368 /// ```                                                                      368 /// ```                                                                     
369 /// # use rumqttc::MqttOptions;                                              369 /// # use rumqttc::MqttOptions;                                             
...                                                                              370 /// # #[cfg(feature = "url")]                                               
370 /// # fn main() -> Result<(), Box<dyn std::error::Error>> {                  371 /// # fn main() -> Result<(), Box<dyn std::error::Error>> {                 
371 /// let options = "mqtt://example.com:1883?client_id=123".parse::<MqttOption 372 /// let options = "mqtt://example.com:1883?client_id=123".parse::<MqttOption
... s>()?;                                                                       ... s>()?;                                                                      
372 /// # Ok(())                                                                 373 /// # Ok(())                                                                
...                                                                              374 /// # }                                                                     
...                                                                              375 /// /// # #[cfg(not(feature = "url"))]                                      
...                                                                              376 /// # fn main() {                                                           
...                                                                              377 /// // ENABLE feature == "url"                                              
373 /// # }                                                                      378 /// # }                                                                     
374 /// ```                                                                      379 /// ```                                                                     
375 ///                                                                          380 ///                                                                         
376 /// You can also go from an [`Url`](url::Url) directly:                      381 /// You can also go from an [`Url`](url::Url) directly:                     
377 ///                                                                          382 ///                                                                         
378 /// ```                                                                      383 /// ```                                                                     
379 /// # use rumqttc::MqttOptions;                                              384 /// # use rumqttc::MqttOptions;                                             
380 /// use std::convert::TryFrom;                                               385 /// use std::convert::TryFrom;                                              
381 /// # use url::Url;                                                          386 /// # #[cfg(feature = "url")]                                               
382 /// # fn main() -> Result<(), Box<dyn std::error::Error>> {                  387 /// # fn main() -> Result<(), Box<dyn std::error::Error>> {                 
...                                                                              388 /// # use url::Url;                                                         
383 /// let url = Url::parse("mqtt://example.com:1883?client_id=123")?;          389 /// let url = Url::parse("mqtt://example.com:1883?client_id=123")?;         
384 /// let options = MqttOptions::try_from(url)?;                               390 /// let options = MqttOptions::try_from(url)?;                              
385 /// # Ok(())                                                                 391 /// # Ok(())                                                                
...                                                                              392 /// # }                                                                     
...                                                                              393 /// # #[cfg(not(feature = "url"))]                                          
...                                                                              394 /// # fn main() {                                                           
...                                                                              395 /// // ENABLE feature == "url"                                              
386 /// # }                                                                      396 /// # }                                                                     
387 /// ```                                                                      397 /// ```                                                                     
388 ///                                                                          398 /// 
```

EdJoPaTo:
Uh, that is both smart and a simple solution! I like that!

Have you checked out the rest of the improvements? Any thoughts on them?

rumqttc/src/state.rs
```diff
@@ -78,6 +78,11 @@ impl MqttState {
     /// connection for persistent sessions while new state should
     /// instantiated for clean sessions
     pub fn new(max_inflight: u16, manual_acks: bool) -> Self {
+        // TODO: Maybe std::num::NonZeroU16 might be neat here?
```
de-sh, line 81: @tekjar what are your thoughts about this?

de-sh:
> have you checked out the rest of the improvements? Any thoughts on them?

Yeah, they look good, especially the builder pattern, sweet.

EdJoPaTo:
> But I do think we should not be deprecating methods such as `new()` and `parse_url()` right away, this is just what I think.

I like to have only one preferred way to do things. This simplifies understanding code and problems of people and simplifies debugging / performance improvements efforts.

Keeping the old methods and marking them as deprecated shows how to migrate to the new preferred way but keeps things working. Normally this shouldnt be a breaking change.

In this case it is a breaking change anyway as methods like `set_keep_alive` are removed completely. (See diff of the examples). So the deprecation hint is more like a helper to find the new way.

Having two ways of doing things is a mess in my eyes but if that's preferred I can provide PRs for the distinct changes.

EdJoPaTo:
(Tried to solve merge conflicts and accidentally merged in main instead of master → rebased my changes to get rid of the wrong merge)

EdJoPaTo:
I just realized there is another `MqttOptions` for the v5 module. Also this PR is just ignored for months now. Not sure if there is any point in keeping it up as it would also need some work in the v5 module now.

I would be happy to see improvements to the MqttOptions creation to provide easier support for multiple ways to connect to a broker especially simplifying configuration via CLI arguments but I am not sure if this helps with that. I guess with the v5 module this got even harder now?

EdJoPaTo:
Additional to my last comment `MqttOptions` seems to have even more problems which should be refactored in my opinion. (See #434)
As there need to be changes on a bigger scale than this PR, I will close this PR as it misses its point.

PR #395, opened 2022-05-13
add support for MQTT 3.1

Freax13:
This pr adds support for MQTT 3.1 to `rumqttc`.

This pr is based on https://www.oasis-open.org/committees/download.php/55095/mqtt-diffs-v1.0-wd01.doc.

This pr introduces a breaking change: https://github.com/bytebeamio/rumqtt/blob/23b23d6997d0e5b0d7ec3d9f4fc79ade6bd9015b/rumqttc/src/mqttbytes/v4/connect.rs#L22

### DIFF
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -333,7 +333,7 @@ async fn mqtt_connect(
     let clean_session = options.clean_session();
     let last_will = options.last_will();
 
-    let mut connect = Connect::new(options.client_id());
+    let mut connect = Connect::new(options.protocol, options.client_id());
     connect.keep_alive = keep_alive;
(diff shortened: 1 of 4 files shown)


Freax13:
The failed CI job seems unrelated to this pr.

de-sh:
Hey @Freax13, had a discussion about this PR with @tekjar. We are currently blocked on adding MQTT 5 support, so we are not very sure about how we could maintain support for 3.1. We are curious about where and why you use it, instead of 3.1.1 or 5.

Freax13:
> We are curious about where and why you continue using it, instead of 3.1.1 or 5.

We don't have a choice, we don't control the broker/server we're connecting to.

PR #396, opened 2022-05-14
Cleanup

EdJoPaTo:
Listed changes from the commits:

- Cargo.toml dependencies:
  - sorted alphabetically
  - use shorter version identifier with same meaning
- remove unused test variables
- add ; at last statement
- cargo fmt
- same mqttoptions variable name in all examples
- minor things found in #391

This should also simplify the diff of #391 (after merging, this will create merge conflicts)

### DIFF
--- a/benchmarks/Cargo.toml
+++ b/benchmarks/Cargo.toml
@@ -3,23 +3,22 @@ name = "benchmarks"
 version = "0.4.0"
 authors = ["tekjar <raviteja@bytebeam.io>"]
 edition = "2018"
-
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
 argh = "0.1"
-serde = { version = "1", features = ["derive"] }
-serde_json = "1"
+async-channel = "1"
+bytes = "1"
+futures = { version = "0.3", features = ["compat"] }
+itoa = "0.4"
+pprof = { version = "0.3", features = ["flamegraph", "protobuf"] }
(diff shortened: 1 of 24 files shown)


EdJoPaTo:
Anything blocking this? (Having not that much time to review is also understandable and totally valid)

tekjar:
This mostly looks good to me. I prefer to freeze versions in libraries as well, atleast critical crates like tokio because I notices some upgrades causing perf bottlenecks in the past

EdJoPaTo:
> I prefer to freeze versions in libraries as well, atleast critical crates like tokio because I notices some upgrades causing perf bottlenecks in the past

The version identifiers which changed in Cargo.toml have exactly the same meaning before and after the change. Both `1.2` and `1` will install version `1.33.7` (when without the lock file). This is counterintuitive for me so I like the shorter version more.
The lock file is still in place and unchanged.

A different idea might be to replace the version requirements with [a different strategy like tilde requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) and remove the lock file then.

EdJoPaTo:
Remove the lock file then too when the tokio version is pinned?

tekjar:
I think that lock file is related to some binary in this workspace. I think it's all cool w.r.t this library. I'll merge this tomorrow in our opensource syncup

EdJoPaTo:
Thanks @adamscybot for bringing it up!
I think this is its own topic that might be bigger than this PR. I created an issue for discussing it in its own place.

PR #405, opened 2022-06-06
fix(performance): Don't create a new runtime in drop

henil:


rumqttc/src/client.rs
```diff
@@ -398,8 +397,6 @@ impl<'a> Iterator for Iter<'a> {
 
 impl<'a> Drop for Iter<'a> {
     fn drop(&mut self) {
-        // TODO: Don't create new runtime in drop
-        let runtime = runtime::Builder::new_current_thread().build().unwrap();
-        self.connection.runtime = Some(mem::replace(&mut self.runtime, runtime));
+        self.connection.runtime = None;
```
de-sh, line 400:
```suggestion
        self.connection.runtime = self.runtime.take();
```

henil: yeah that makes more sense, it slipped by mind that mem::replace returns back the old value.

henil:
Not sure why some tests are failing seems unrelated. #402 is having a similar issue.

rumqttc/src/client.rs
```diff
@@ -398,8 +398,28 @@ impl<'a> Iterator for Iter<'a> {
...
 impl<'a> Drop for Iter<'a> {
     fn drop(&mut self) {
-        // TODO: Don't create new runtime in drop
-        let runtime = runtime::Builder::new_current_thread().build().unwrap();
-        self.connection.runtime = Some(mem::replace(&mut self.runtime, runtime));
+        self.connection.runtime = self.runtime.take();
+    }
+}
+
+#[cfg(test)]
```
de-sh, line 405: @tekjar do we keep the test?

rumqttc/src/client.rs
```diff
@@ -367,23 +367,24 @@ impl Connection {
...
+            runtime: Some(runtime),
         }
     }
 }
 
 /// Iterator which polls the `EventLoop` for connection progress
 pub struct Iter<'a> {
     connection: &'a mut Connection,
-    runtime: runtime::Runtime,
+    runtime: Option<runtime::Runtime>,
```
tekjar, line 378: Is it possible to use runtime by reference (like connection) instead of wrapping it with an option and unwrapping everytime?

henil: I think the bigger question is do we want to allow consumer having multiple `Iter` at the same time?

If yes then we can just take a `&mut Connection` in the `Iter` and remove runtime completely. And run the executor using runtime in `Connection` itself. 

If no then we can probably wrap `Connection` inside a Inner type and use a singleton like pattern to get a `Iter`.

de-sh: I believe what you mean is an `Arc<Murex<Connection>>` as `&mut` is always exclusively owned.

Issue #406, opened 2022-06-08
`conn_timeout` does not apply to `eventloop::network_connect` 


david-mcgillicuddy-moixa:
Hi, we were having a weird issue where in certain very rare cases, the mqtt event loop would get stuck and hang forever. After quite some debugging the problem has been narrowed down to the case where the TLS server replies with hello, and then nothing more. This means that `eventloop::network_connect` gets called, which waits for further packets, which never come. The timeout mechanism that exists in `eventloop` only takes place _after_ `network_connect` has completed, i.e. for `eventloop::mqtt_connect` and then the keepalive mechanism kicks in for `eventloop::select`. 
As a workaround in Moixa's code we now have `eventloop::poll` racing against a timeout to call `AsyncClient::cancel`, which works fine, but this was a subtle bug that happened so rarely it was a nightmare to get a reproducible case. I think that either `conn_timeout` should apply to `network_connect` too, or a second optional timeout field should be available that applies solely to `network_connect`.
Happy to make either change myself. I find it pretty hard to believe that this is working as intended given that there _is_ a separate timeout mechanism around every other part of the event loop.

david-mcgillicuddy-moixa:
Yep, looks good to me.

PR #408, opened 2022-06-09
Move timeout to `connect()` call

de-sh:
Solves #406 

Moving all use of `connection_timeout` up to call for `connect()` to ensure the timeout is applied to the entire connect procedure and not merely to a specific async closure.


de-sh:
@tekjar, any ideas why the discrepancy? `connect_or_cancel` is missing from v5 code:
- [v3.1.1](https://github.com/bytebeamio/rumqtt/blob/master/rumqttc/src/eventloop.rs#L236)
- [v5](https://github.com/bytebeamio/rumqtt/blob/master/rumqttc/src/v5/eventloop.rs#L118)

PR #413, opened 2022-07-11
fix(serialization): Serialize no properties as 0x0 byte

henil:
ref: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Properties


henil:
looks like its failing due to a flaky test

PR #421, opened 2022-08-01
feat: remove ability to cancel `EventLoop.poll()` mid execution

de-sh:
Removes the ability for users to cancel requests while they are being polled within the `EventLoop`. This feature was deemed to be overkill and hence good to get rid of.

CC @tekjar


de-sh:
@tekjar shall I remove code associated with `cancel()` in `EventLoop` in this PR itself?

Issue #426, opened 2022-08-10
Fix leaking `ClientConfig` struct


henil:
Due to the way we where re-exporting the `tokio_rustls::rustls::ClientConfig` struct to make it easy for library authors to import it, documentation is generated in a way which suggests its a struct from our library which is not the case.

PR #428, opened 2022-08-12
fix(exports): remove leaky `pub use`

henil:
Closes: #426 

This exports are causing `ClientConfig` to be shown as a type coming from `rumqttc` in [docs.rs](https://docs.rs/rumqttc/latest/rumqttc/struct.ClientConfig.html). Uses of this [type](https://docs.rs/rumqttc/latest/src/rumqttc/lib.rs.html#299-304) still expect the correct `tokio_rustls::rustls::ClientConfig` type.

 So this could just be a bug in generating docs still not sure if this is the right way to do it, suggestions welcome. 

re: https://github.com/thin-edge/thin-edge.io/pull/1291

### DIFF
--- a/rumqttc/src/lib.rs
+++ b/rumqttc/src/lib.rs
@@ -120,8 +120,11 @@ pub use mqttbytes::*;
 pub use state::{MqttState, StateError};
 #[cfg(feature = "use-rustls")]
 pub use tls::Error as TlsError;
+
 #[cfg(feature = "use-rustls")]
-pub use tokio_rustls::rustls::ClientConfig;
+pub use tokio_rustls;
+#[cfg(feature = "use-rustls")]
+use tokio_rustls::rustls::ClientConfig;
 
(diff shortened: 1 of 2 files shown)


de-sh:
Any ideas why this PR triggered the implement `Eq` compiler error? Doesn't seem like the changes have anything to do with it? Maybe a compiler version upgrade is the cause and the newer compiler requires this warning and our use of the `-D warning` argument is causing CI to fail?

henil:
Yeah that [lint](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq) was introduced in release 1.63.0. If implementing `Eq` for those types is not a problem I will open a PR for it.

henil:
hmm there still seems to be some problem with docs.rs linking. Now the extra struct from rumqttc is gone but now all the references are unlinked. locally they seem to be linked properly so not sure what the problem is.

Issue #430, opened 2022-08-14
Add ability to get events through polling, not only through an iterator.


SirVer:
Hey, thanks for the crate! Very useful indeed and I had very little issues with using it so far. I only use the sync interface, so I am talking about this below.

However, what would be useful for me is to get a polling interface, similar to [`std::sync::mspc::Receiver::try_recv`](https://doc.rust-lang.org/std/sync/mpsc/struct.Receiver.html#method.try_recv). Currently, all I can do to drive the event loop is calling `Connection.iter`. Quite often though, I have multiple places I receive data from: for example in some of my tools, I have a global bool that is set when it is time to clean up. In others, I also recv messages from other threads. Right now, I have to wait for the next mqtt signal to act on it, since the iter is blocking on action on mqtt - which complicates my designs significantly.

I do understand the implications that `try_recv` would need to be called regularly to keep the mqtt loop healthy, but that seems a fine interface ask to me still. Could you take this into consideration?

tekjar:
@SirVer We'll take this into consideration in the current design

de-sh:
> Hey, thanks for the crate! Very useful indeed and I had very little issues with using it so far. I only use the sync interface, so I am talking about this below.
> ...

Thank you for your inputs. Module v5 is heavily experimental at the moment. Please do not use it in production as there are possible holes(we are yet to fully test and verify our implementation), we are working out the proper way to do this and it seems like a `try_next()` that returns an Empty error is what you are pointing towards? Again, thanks for your input, will open a PR to resolve this by the weekend.

SirVer:
yes. That is what I mean. It would make it possible to combine events from other sources than just mqtt which is a common pattern in my use cases. 

I’ll post an example that currently is hard later.

SirVer:
Here is a motivating example in code. The aim here is to replace this "poor mans logging" commandline: ```mosquitto_sub -v -t "topic/of/interest"  | tee -a /srv/log/mqtt_logger.log```.

Essentially, we listen to one or more topics, and every topic received we write into a `logfile`. since the logfile is buffered and we only want complete messages in the file, we want to ensure that upon termination of the program the buffer is flushed. One way of doing so is shown below: We register a `ctrl-c` handler which sends a message to a channel. The main thread needs to terminate if this channel has a message. The code below is functional, however due to the iterator interface of rumqtt the program will only terminate upon the next event received after a ctrl-c. Some topics only see a message once per hour, so this is very unpractical.

```rust
[...]

PR #431, opened 2022-08-15
feat: Create non-blocking interface for `Notifier`

de-sh:
Add `try_recv()` which resolves to `Ok` only if it is able to immediately receive an incoming packet.
Else errors out with a special error type `TryRecvError` which has variants to signify if connection is active or dead.

Closes: #430

### DIFF
--- a/rumqttc/src/v5/notifier.rs
+++ b/rumqttc/src/v5/notifier.rs
@@ -4,9 +4,12 @@ use std::sync::{Arc, Mutex, RwLock};
 
 use crate::v5::Incoming;
 
-/// Returnes by
+/// Error type to denote if Notifier failed to resolve due to empty buffer or due to disconnection
+#[derive(Debug, Eq, PartialEq)]
 pub enum TryRecvError {
+    /// [`EventLoop`] is disconnected from broker
     Disconnected,
+    /// [`EventLoop`] is connected to broker
     Waiting,
 }
 
@@ -60,7 +63,7 @@ impl Notifier {
 
         match next {
             Some(p) => Ok(p),
-            None if self.is_disconnected() => Err(TryRecvError::Waiting),
+            None if self.is_disconnected() => Err(TryRecvError::Disconnected),
             None => Err(TryRecvError::Waiting),
         }
     }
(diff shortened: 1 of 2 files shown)


de-sh:
Some possible changes to this design could include returning a `NotifierError` object that has the variants `Closed` and `Waiting` or something similar to signify that the connection is either dead or yet to receive new packets.

rumqttc/src/v5/notifier.rs
```diff
@@ -36,30 +34,34 @@ impl Notifier {
     pub fn iter(&mut self) -> NotifierIter<'_> {
         NotifierIter(self)
     }
+
+    #[inline]
+    /// Retruns None immediately even if not disconnected, use when expecting non-blocking interface
```
SirVer, line 39: typo: Retruns

rumqttc/src/v5/notifier.rs
```diff
@@ -36,30 +34,34 @@ impl Notifier {
...
+            val => val,
+        }
+    }
 }
 
 impl Iterator for Notifier {
     type Item = Incoming;
 
     #[inline]
+    /// Retrun None only if disconnected, else block and retry till resolves as Some
```
SirVer, line 61: typo: Retrun

rumqttc/src/v5/notifier.rs
```diff
@@ -36,30 +34,34 @@ impl Notifier {
     pub fn iter(&mut self) -> NotifierIter<'_> {
         NotifierIter(self)
     }
+
+    #[inline]
+    /// Retruns None immediately even if not disconnected, use when expecting non-blocking interface
+    pub fn try_next(&mut self) -> Option<Incoming> {
```
SirVer, line 40: I suggest calling this `try_recv` instead, since `try_next` makes this sound like an iterator (which this struct is not).

SirVer:
The interface looks like the one I was hoping to get. Stylistic nit: Now that you have a `try_recv`, people will expect a blocking `recv`. Once you acknoledge that, you realize that you should probably just copy the API from `mspc::sync::Receiver`, i.e. have Notifier implement  at least `recv, try_recv, iter` and maybe even `recv_deadline` and `try_iter`. 

Unrelated to this patch, but still worth a mention I think: looking through the code I am now confused about the error cases: Why is `next` and `try_next` not returning a `Result` anymore (like they do in the current released crate)?. It seems these operations are very much fallible - for example through disconnect of the mqtt server - or are they not?

PR #436, opened 2022-08-25
feat: `MqttOptions::parse_url` uses `rustls-native-certs::load_native_certs()` for encrypted connections

### LINKED ISSUE #435: parse_url does not handle ssl/tls like document implies
The document says:
> NOTE: A url must be prefixed with one of either tcp://, mqtt://, ssl://,mqtts://, ws:// or wss:// to denote the protocol for establishing a connection with the broker.

https://docs.rs/rumqttc/0.14.0/rumqttc/struct.MqttOptions.html#method.parse_url

While `"mqtts" | "ssl"` is simply not handled:
https://github.com/bytebeamio/rumqtt/blob/2f11546b64d4ab86b2b49642d8b7068f624992a9/rumqttc/src/lib.rs#L611-L620
similar to #434 

At least the document should be updated to prevent confusion.

de-sh:
fixes #435 

## Changelog
- `MqttOptions::parse_url` when passed the url for an encrypted protocol, e.g. `mqtts://`, `ssl://` or `wss://` will from now on use `rustls-native-certs::load_native_certs()` to load the certificates from the native cert store.
- Updated documentation

### DIFF
--- a/rumqttc/Cargo.toml
+++ b/rumqttc/Cargo.toml
@@ -29,6 +29,7 @@ rustls-pemfile = { version = "0.3", optional = true }
 thiserror = "1"
 tokio = { version = "1", features = ["rt", "macros", "io-util", "net", "time"] }
 tokio-rustls = { version = "=0.23.3", optional = true }
+rustls-native-certs = "0.6"
 url = { version = "2", default-features = false, optional = true }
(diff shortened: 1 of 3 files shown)


tekjar:
On a different node, can we use system certificates to establish secure connection by default. It can always be overridden using Transport constructs?

PR #437, opened 2022-08-26
refactor!: make eventloop handle private to crate & move tests around

henil:
BREAKING CHANGE: don't expose flume types

Make the handle public only for the crate. Some tests in integration required this handle so i moved those in as a unit test.

And i needed to duplicate `Broker` (the dummy broker used in tests) because there was no way to share modules between integration tests and unit tests.


rumqttc/src/eventloop.rs
```diff
@@ -327,3 +327,341 @@ pub(crate) async fn next_pending(
     time::sleep(delay).await;
     pending.next()
 }
+
+#[allow(dead_code)]
```
henil, line 331: needed to add this because we only use this in unit tests in `eventloop` module. We might also use this in other modules so added it here instead of in unit tests module directly.

rumqttc/src/lib.rs
```diff
@@ -902,3 +901,313 @@ mod test {
         let _mqtt_opts = MqttOptions::new("", "127.0.0.1", 1883).set_clean_session(true);
     }
 }
+
+// Dummy broker used in unit tests
+#[allow(dead_code)]
```
henil, line 906: needed to add this because we only use this in unit tests in `eventloop` module. We might also use this in other modules so added it here instead of in unit tests module directly.

PR #440, opened 2022-08-28
fix: allow use of `url` feature without `use-rustls`

lautat:
Compiling the crate failed when default features were disabled and `url` feature was enabled. That happened due to missing conditinal compilation attributes in URL parsing. This was tested by building `rumqttc` with the following commands:
```sh
cargo build --no-default-features --features=url
cargo build --no-default-features --features=url --features=websocket
```


de-sh:
Thank you for submitting the fixes @lautat! 🎊

PR #442, opened 2022-09-03
Opensource beamd's broker 🎉

henil:
We are open sourcing our internally used embeddable MQTT broker. Our plan is to replace current `rumqttd` crate with this. It is available both as a library and a binary. 

We currently support MQTT 3.1.1 and plan to support MQTT 5 in future. 

For more information about how to run the broker and which MQTT 3.1.1 features we support checkout README.


msdrigg:
What are the benefits / costs of using this over the current rumqttd implementation?

henil:
We plan to write a blog about the same during our release announcement

msdrigg:
Ooh exciting! Please link it here when you announce it!

PR #453, opened 2022-09-11
fix(mqttbytes): when present wildcard must occupy entire topic level

pabigot:
Improves validation of filters by checking rule in 4.7.1.3 ('+' must occupy an entire level of the filter).

~~Improves usability by adding a `get_wildcards(topic, filter) -> Option(Vec<&str>)` function that returns `None` if the topic doesn't match the filter, and otherwise returns `Some(v)` where v contains the fragments of the topic corresponding to filter wildcards.~~ removed as requested.

### DIFF
--- a/rumqttc/src/mqttbytes/topic.rs
+++ b/rumqttc/src/mqttbytes/topic.rs
@@ -26,18 +26,25 @@ pub fn valid_filter(filter: &str) -> bool {
 
     let hirerarchy = filter.split('/').collect::<Vec<&str>>();
     if let Some((last, remaining)) = hirerarchy.split_last() {
-        // # is not allowed in filer except as a last entry
-        // invalid: sport/tennis#/player
-        // invalid: sport/tennis/#/ranking
         for entry in remaining.iter() {
+            // # is not allowed in filter except as a last entry
+            // invalid: sport/tennis#/player
+            // invalid: sport/tennis/#/ranking
             if entry.contains('#') {
+                return false;
+            }
+
+            // + must occupy an entire level of the filter
+            // invalid: sport+
+            if entry.len() > 1 && entry.contains('+') {
                 return false;
             }
         }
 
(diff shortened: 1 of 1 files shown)


pabigot:
`get_wildcards` could be named something else; `get_matches`, `get_wildcard_matches`, `wildcard_matches`, ....  That last one might be best.

henil:
Bug fix LGTM, but not sure about where the `get_wildcards` might be useful in practice. @pabigot did you have any specific usecase where that function might be useful?

pabigot:
@henil Yes.  I have services where the topic path is constructed from things like `{prefix}/+/+` where the prefix comes from a configuration file and may include multiple topic levels.  I can't just split on `'/'` because I don't know the index that gets me to the wildcards.

Also, correct handling of `#` is tricky: `foo/#` matches `foo`, `foo/`, and `foo/bar` with the match values being `""`, `"/"`, and `"/bar"`, and in some cases it's useful to distinguish all three.  Having to have code for that at every point where I'm decoding a topic would be painful when I could just do:
```
let state_filter = format!("{}/+", cfg.pfx).as_str();
let set_filter = format!("{}/+/set/#", cfg.pfx).as_str();
...
let topic = p.topic.as_str();
if let Some(wc) = get_wildcards(topic, state_filter) {
    // work with wc[0]
} else if let Some(wc) = get_wildcards(topic, set_filter) {
    // work with wc[0] and wc[1]
} else {
  ...
}
```

Though if you don't think that's useful I'll just remove the feature patch, since now that I have it I can put it in my own crate for people to use with any MQTT client implementation.

henil:
@pabigot we discussed this internally and we think having it in a separate crate is a better option, so can you remove that patch?

pabigot:
@henil done

henil:
Thank you @pabigot!

PR #458, opened 2022-09-21
feat: Implement `recv` `try_recv` `recv_timeout` for `Connection`

de-sh:
Continued from discussion on #430

Implement `recv` `try_recv` `recv_timeout` for `Connection` similar to `v5::Notifier::try_recv`

CC @sirver

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,23 @@
 ### Unreleased
 ---
 rumqttc
+-------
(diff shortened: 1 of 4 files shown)


SirVer:
This looks good to me. I cannot comment if the dependency on futures is a good or bad one as of lack on experience. It seems a lightweight dependency that does not pull a lot of other stuff in, but I have no idea how future proof it is in the Rust ecosystem.

PR #468, opened 2022-09-28
help to trace the request's status

jm-observer:


de-sh:
This would not work as expected, the place where pkids are set is over here: https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/state.rs#L321

jm-observer:
i think when i get the pkid from eventloop, it means the event has done, this is not trace. i neet to trace the event, becouse the event initiator is other thread.

de-sh:
@jm-observer in that case you will need to maintain a channel between thread that runs `EventLoop` and the thread that run `AsyncClient`.

PR #476, opened 2022-10-01
fix: hyperlinks in doc and ignore test dependent on speed of execution

henil:
- format links properly in docs
- ignore test that is sensitive to execution environment


de-sh:
@henil lets only keep the changes related to CI in here, code changes I am moving to #481

Issue #473, opened 2022-10-01
[rumqttc] Getting "ConnectionRefused(NotAuthorized)" after updating to 0.14 to 0.17 and mqtt v4 to v5


kylehendricks:
I recently updated from rumqttc 0.14 to 0.17 and also ported over to using the MQTT v5.  I'm using username/password authorization and now I'm suddenly getting `ConnectionRefused(NotAuthorized)` when attempting to connect to the broker.  I'm using `MqttOptions::set_credentials` to set the username/password.

My broker is Mosquitto 2.0.15.

kylehendricks:
Took a peek at the source and it looks like a lot of things aren't hooked up yet in v5?  (last will, credentials etc)

bjornhamra:
I stumbled upon the same and was able to tempfix this in my playground client by just slipping an mqttoptions ref into Network::connect and cloning Login and LastWill from there. I won't PR this because it's almost certainly not the right way to do it, but if you just need a temporary quickfix like I did then this might be of use to you.


```diff
diff --git a/rumqttc/src/v5/eventloop.rs b/rumqttc/src/v5/eventloop.rs
index 86b0d50..739ae10 100644
--- a/rumqttc/src/v5/eventloop.rs
+++ b/rumqttc/src/v5/eventloop.rs
@@ -298,7 +298,7 @@ async fn mqtt_connect(
     };

     // send mqtt connect packet
-    network.connect(connect).await?;
+    network.connect(connect, options).await?;

     // validate connack
     match network.read().await? {
diff --git a/rumqttc/src/v5/framed.rs b/rumqttc/src/v5/framed.rs
index cb63711..68e4620 100644
--- a/rumqttc/src/v5/framed.rs
+++ b/rumqttc/src/v5/framed.rs
@@ -2,7 +2,8 @@ use bytes::BytesMut;
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};

 use super::mqttbytes::v5::Packet;
-use super::mqttbytes::{self, Connect};
+use super::mqttbytes::{self, Connect, Login};
+use super::MqttOptions;
 use super::{Incoming, MqttState, StateError};
 use std::io;

@@ -96,9 +97,16 @@ impl Network {
         }
     }

-    pub async fn connect(&mut self, connect: Connect) -> io::Result<usize> {
+    pub async fn connect(&mut self, connect: Connect, options: &MqttOptions) -> io::Result<usize> {
[...]

PR #478, opened 2022-10-05
fix(v5): send last_will and login info with connect

henil:
Fixes: #473


brianmay:
What needs to happen to get this merged? The build failures don't seem to have anything to do with the changes.

brianmay (approved):

henil:
I was working on refactoring rumqttc so we wouldn't need this hotfix but it turned out to be alot more work than i expected (also would require some breaking changes). So that was deprioritized atm, then I was not available for a month due to personal reasons in which this PR got buried down.

PR #480, opened 2022-10-05
fix(protocol): Extract out Disconnect packet logic similar to other packets

henil:
And add optional properties to `unsubscribe`, `unsuback` and `connack` packet, so its consistent with other packets.


henil:
(ci failing for reasons that are fixed in #476)

PR #482, opened 2022-10-06
v5 AsyncClient: add fuction to tracing pkid

jm-observer:


jm-observer:
hello, do you have plan to implement this feature? i am develep a mqtt client [desktop application](https://github.com/jm-observer/for-mqtt) that needs this feature

PR #489, opened 2022-10-26
feat(rumqttd): Add support for disk persistent connections

tekjar:
- Support re-connection of clients in persistent mode
- Persist messages on disk for clients connected in persistent mode and that have temporarily disconnected


rumqttd/src/link/persistance.rs
```diff
@@ -0,0 +1,325 @@
...
+    ConnectionAck(String),
+    #[error("Channel try send error")]
+    TrySend(#[from] TrySendError<(ConnectionId, Event)>),
+    #[error("Link error = {0}")]
+    Link(#[from] LinkError),
+}
+
+/// Orchestrates between Router and Network.
+pub struct PersistanceLink<P: AsyncProtocol> {
+    _connect: Connect,
```
swanandx, line 49: why do we need `Connect` ? are we going to use it later? if not, then we should remove it.

rumqttd/src/link/persistance.rs
```diff
@@ -0,0 +1,325 @@
...
+
+pub(super) struct DiskHandler<P: AsyncProtocol> {
+    storage: Storage,
+    protocol: P,
+}
+
+impl<P: AsyncProtocol> DiskHandler<P> {
+    fn new(client_id: &str, protocol: P) -> Result<Self, Error> {
+        // TODO: take params from config
+        let path = format!("/tmp/rumqttd/{}", &client_id);
```
swanandx, line 69: to get temporary directory, we should use
```rust 
env::temp_dir();
```
  - [docs](https://doc.rust-lang.org/std/env/fn.temp_dir.html)

rumqttd/src/link/persistance.rs
```diff
@@ -0,0 +1,325 @@
...
+                    self.disk_handler.write(&mut self.notifications);
+                }
+            }
+        }
+    }
+
+    async fn write_to_client(&mut self) -> Result<(), Error> {
+        // separate publish notifications out
+        let mut publish = VecDeque::new();
+        let mut non_pulish = VecDeque::new();
```
swanandx, line 234: a typo, should be `non_publish` ig.

rumqttd/src/router/mod.rs
```diff
@@ -86,6 +86,34 @@ pub enum Notification {
...
+            Notification::DeviceAck(ack) => match ack {
+                Ack::ConnAck(_, connack) => packet = Packet::ConnAck(connack, None),
+                Ack::PubAck(puback) => packet = Packet::PubAck(puback, None),
+                Ack::SubAck(suback) => packet = Packet::SubAck(suback, None),
+                Ack::PingResp(pingresp) => packet = Packet::PingResp(pingresp),
+                Ack::PubRec(pubrec) => packet = Packet::PubRec(pubrec, None),
+                Ack::PubRel(pubrel) => packet = Packet::PubRel(pubrel, None),
+                Ack::PubComp(pubcomp) => packet = Packet::PubComp(pubcomp, None),
+                Ack::UnsubAck(unsuback) => packet = Packet::UnsubAck(unsuback, None),
+                _ => unimplemented!(),
```
swanandx, line 108: return a `ParseError` of some kind? Or as return type ( `MaybePacket` ) is an `Option`, just return `None`!

swanandx (commented):
Just a code review, I haven't verified working!

rumqttd/src/link/persistance.rs
mnpw: > On Unix, returns the value of the TMPDIR environment variable if it is set
ideally i'd want this directory to be not dependent on an environment variable since on broker restart I want to reliably find the directory we previously wrote to

rumqttd/src/router/mod.rs
mnpw: None holds special significance here (it implies that the the Link reading this MaybePacket should read again from the Router - so wouldn't want to use it). Restructuring this would take some more thought, so running with a `unimplemented()` for now

Issue #467, opened 2022-09-26
Recv_timeout does not seem to be working


SirVer:
calling the newly added `recv_timeout` gives me the following error:

```
thread 'main' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime', /cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.17.0/src/runtime/context.rs:54:26
```

Using `try_recv` instead gives the same error.

SirVer:
This is the code I am using:

```
'outer: loop {
        // To make our client name unique, we append the current time.
        let now = Utc::now().to_rfc3339();
        let mut mqttoptions =
            MqttOptions::new(format!("mqtt_logger_{now}"), &args.hostname, args.port);
        mqttoptions.set_keep_alive(Duration::from_secs(60));

        let (mut client, mut connection) = Client::new(mqttoptions, 10);
        for t in &args.topics {
            client.subscribe(t, QoS::ExactlyOnce)?;
        }

        use rumqttc::Event::*;
        use rumqttc::Packet::*;
        loop {
            loop {
                let notification = match connection.recv_timeout(MIN_REACTION_TIME) {
                    Ok(rv) => rv,
                    Err(RecvTimeoutError::Timeout) => break,
                    Err(RecvTimeoutError::Disconnected) => continue 'outer,
                };

                let n = match notification {
                    Ok(n) => n,
                    Err(e) => {
                        eprintln!("MQTT error. Will reconnect: {e}.");
                        continue 'outer;
                    }
                };

                let res = match n {
                    Incoming(Publish(data)) => {
                        handle_message(&data.topic, &data.payload, &mut logfile, args.quiet)
                    }
                    Incoming(_) | Outgoing(_) => Ok(()),
                };

                if let Err(e) = res {
                    eprintln!("error in handle_message. Will reconnect: {e}.");
                    continue 'outer;
                }
            }
        }
```

de-sh:
I am not sure what exactly triggered this, but I believe the problem stems from an error [here](https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/client.rs#L249-L252) while trying to build the runtime on the same thread where one was already built in a previous iteration? 

Maybe I don't make much sense but this issue seems to be pretty complicated for me to even comprehend, frankly.

SirVer:
@de-sh Can you reproduce this?

tekjar:
@SirVer I'll take a look at this soon

[...]

PR #492, opened 2022-10-29
fix(client): enter tokio runtime context before calling timeout in recv_timeout

swanandx:
`timeout` function uses `Sleep` internally, but as we are outside of Tokio runtime context, it will fail to spawn it and thus will panic! [reference](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html#panics).

To fix it, we can explicitly enter the runtime by using [enter](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.enter) method.

Fixes Issue: #467

### DIFF
--- a/rumqttc/src/client.rs
+++ b/rumqttc/src/client.rs
@@ -427,6 +427,9 @@ impl Connection {
         &mut self,
         duration: Duration,
     ) -> Result<Result<Event, ConnectionError>, RecvTimeoutError> {
+        // Enter the runtime so we can use Sleep, which is required by timeout.
+        // ref: https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.enter
+        let _guard = self.runtime.enter();
         let f = timeout(duration, self.eventloop.poll());
         let event = self
             .runtime
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Removed unused dependencies and updated version of some of used libraries to fix dependabots warning (#475)
 - (rumqttd) Added properties field to `Unsubscribe`, `UnsubAck`, and `Disconnect` packets so its consistent with other packets. (#480)
 - (rumqttd) Changed default segment size in demo config to 100MB (#484)
+- (rumqttc) Fixed panicking of `timeout` in `recv_timeout` by entering tokio runtime context (#492)
 
 ### R17
 ---
--- a/rumqttc/src/v5/client.rs
+++ b/rumqttc/src/v5/client.rs
@@ -430,6 +430,9 @@ impl Connection {
         &mut self,
         duration: Duration,
     ) -> Result<Result<Event, ConnectionError>, RecvTimeoutError> {
+        // Enter the runtime so we can use Sleep, which is required by timeout.
+        // ref: https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.enter
+        let _guard = self.runtime.enter();
         let f = timeout(duration, self.eventloop.poll());
         let event = self
             .runtime


aeshirey:
I just tested and found the panic still occurs. Here's my minimum, reproducible example:

Cargo.toml:
```toml
[package]
name = "my-example"
version = "0.1.0"
edition = "2021"

[dependencies]
rumqttc = { default-features = false, git = "https://github.com/bytebeamio/rumqtt" }
```

main.rs:
```rust
use rumqttc::mqttbytes::v4::Packet;
use rumqttc::{Client, Event, MqttOptions, QoS};

fn main() {
    let mqttoptions = MqttOptions::new("my-client-id", "localhost", 1883);
    let (mut client, mut connection) = Client::new(mqttoptions, 10);
    client.subscribe("sub-topic", QoS::ExactlyOnce).ok();

    loop {
        let event = connection.try_recv().unwrap().unwrap();

        if let Event::Incoming(Packet::Publish(p)) = event {
            let s = String::from_utf8_lossy(p.payload.as_ref()).to_string();
            println!("{s}");
        }
    }
}
```

Running it:

```bash
$ ~/cargo-target/release/my-example
thread 'main' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime', /home/adam/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/context.rs:56:26
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

The stacktrace includes the following:

```
   9: futures_util::future::future::FutureExt::now_or_never
  10: rumqttc::client::Connection::try_recv
```

So when I look at `Connection::try_recv`, it [calls `f.now_or_never()`](https://github.com/bytebeamio/rumqtt/blob/38786428bb1dae931073a118ef1acc635aeb04e5/rumqttc/src/client.rs#L417), where `f` is the result of [`EventLoop::poll`](https://github.com/bytebeamio/rumqtt/blob/38786428bb1dae931073a118ef1acc635aeb04e5/rumqttc/src/eventloop.rs#L112), an async function. I'm not _quite_ sure what the solution is here, but I see that in the previous PR, you had updated the very next function, [`recv_timeout` to include `self.runtime.enter()`](https://github.com/bytebeamio/rumqtt/blob/38786428bb1dae931073a118ef1acc635aeb04e5/rumqttc/src/client.rs#L432), which `try_recv` doesn't have. Possibly [`recv`](https://github.com/bytebeamio/rumqtt/blob/38786428bb1dae931073a118ef1acc635aeb04e5/rumqttc/src/client.rs#L404) should also include this call?

I don't do much with async, but is this the optimal way of supporting fully synchronous APIs?

rumqttc/src/client.rs
```diff
@@ -427,6 +427,9 @@ impl Connection {
         &mut self,
         duration: Duration,
     ) -> Result<Result<Event, ConnectionError>, RecvTimeoutError> {
+        // Enter the runtime so we can use Sleep, which is required by timeout.
+        // ref: https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.enter
+        let _guard = self.runtime.enter();
```
de-sh, line 432: Seem like this could have been fixed by ensuring the construction of the timeout happened within an async block passed to `block_on` without having to explicitly enter the runtime.

```
.block_on(async { timeout(duration, f).await })
```

swanandx: I did try it out, but for me it didn't work. That is why I ended up choosing the solution of manually entering runtime.

de-sh: I believe since the `timeout` would be created outside the runtime context before being passed into `block_on()`, the reactor would be deemed missing whereas within `async { }` the timeout is only created once the external future gets polled from within the runtime.

Thus why `.block_on(timeout(duration, f))` behaves different to `.block_on(async { timeout(duration, f).await })`.

swanandx: Yeah you are right, I was creating timeout future first and then passing it down in that async block inside block_on 😅. Good catch! If entering guard isn't required, can we remove it in your PR?

PR #497, opened 2022-11-09
fix: enter runtime context for `now_or_never()` within `try_recv`

### LINKED ISSUE #496: Can the sync API be used without a tokio reactor?
I'd like to use `rumqttc` in a strictly synchronous context, but when I do so, my app panics with:

> thread 'main' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime'

Is there a way to run the client without any async runtime/scaffolding? If not, do you have plans to allow this?

de-sh:
Closes #496 

Enter the runtime context within `Connection::try_recv()` in order to allow the creation and passing of a noop waker for the instantaneous polling of associated future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,16 @@
 ### Unreleased
-
+---
+rumqttc
+-------
+- Fixed panicking in `recv_timeout` and `try_recv` by entering tokio runtime context (#492, #497)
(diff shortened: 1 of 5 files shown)


de-sh:
The errors seem to be caused by new linting rule additions to clippy

PR #507, opened 2022-11-19
refactor(router): Refactor forward_device_data()

mnpw:


henil:
changes in this PR where merged along with #503

Issue #515, opened 2022-11-30
Get notifications for new `Subscribe` or `Unsubscribe` packets


Bergmann89:
Is there any possibility to get informed about new subscriptions/unsubscriptions of any connected client?

If not can we implement some simple channel to receive such events?

PR #516, opened 2022-12-01
Add channel to receive subscription events

Bergmann89:
Related to issue #515


tekjar:
@Bergmann89 We are planning for `EventsLink` in R19 release where user can register for interested events link new connections, disconnections, subscriptions, message drops etc.  I'll tag you in that PR soon when it's public

henil:
https://github.com/bytebeamio/rumqtt/issues/515#issuecomment-1384054748

PR #523, opened 2022-12-05
feat(cli): Make configuration file optional

henil:
- also add a subcommand to generate a config file on demand. 

for example

```
 rumqttd generate-config > rumqttd.toml
```

### DIFF
--- a/rumqttd/src/main.rs
+++ b/rumqttd/src/main.rs
@@ -1,6 +1,11 @@
-use rumqttd::{Broker, Config};
+use std::io::Write;
+
+use config::FileFormat;
+use rumqttd::Broker;
 
 use structopt::StructOpt;
+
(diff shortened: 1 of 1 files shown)


rumqttd/src/main.rs
```diff
@@ -1,7 +1,12 @@
-use rumqttd::{Broker, Config};
+use std::io::Write;
+
+use config::FileFormat;
+use rumqttd::Broker;
 
 use structopt::StructOpt;
 
+pub static RUMQTTD_DEFAULT_CONFIG: &str = include_str!("../demo.toml");
```
henil, line 8: used `include_str!` macro instead of copying the config in source code directly was to avoid need to change config file in 2  places whenever we want to update the default config file.

in this case we just need to remember to update this path whenever configuration file name is update or path is changed.

Issue #487, opened 2022-10-22
mqtt v5 client most subscribe options not implemented


brianmay:
Looking at https://github.com/bytebeamio/rumqtt/blob/29f5b1e8a6e479a4454e0b0feb6142942f39b15e/rumqttc/src/v5/mqttbytes/v5/subscribe.rs I think QoS is the only subscribe option that is sent to the server.

Others, nolocal, reserve_retain, and retain_forward_rule do not get incorporated into the options value.

brianmay:
I think this can be closed now that #536 was merged.

PR #536, opened 2022-12-11
fix: add processing of missing Mqtt subscribe options

brianmay:
Fix for #487.

### DIFF
--- a/rumqttd/src/protocol/v5/subscribe.rs
+++ b/rumqttd/src/protocol/v5/subscribe.rs
@@ -140,6 +140,20 @@ mod filter {
         let mut options = 0;
         options |= filter.qos as u8;
 
+        if filter.nolocal {
+            options |= 0b0000_0100;
+        }
+
+        if filter.preserve_retain {
+            options |= 0b0000_1000;
+        }
+
+        options |= match filter.retain_forward_rule {
+            RetainForwardRule::OnEverySubscribe => 0b0000_0000,
+            RetainForwardRule::OnNewSubscribe => 0b0001_0000,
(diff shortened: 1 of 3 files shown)


brianmay:
Seems to work. Note that initially I wasn't getting retained messages. This is because the default for `retain_forward_rule` is `Never`:

https://github.com/bytebeamio/rumqtt/blob/29f5b1e8a6e479a4454e0b0feb6142942f39b15e/rumqttc/src/v5/mqttbytes/mod.rs#L336

This default was somewhat unexpected, and because the value was previously ignored it had no affect.

henil:
ported the changes to `rumqttd`'s protocol. and updated the default for `RetainForwardRule` to `OnEverySubscribe` because `Never` by default is not very intuitive.

rumqttd/src/protocol/v5/subscribe.rs
```diff
@@ -140,6 +140,20 @@ mod filter {
         let mut options = 0;
         options |= filter.qos as u8;
 
+        if filter.nolocal {
```
de-sh, line 143: This might be unnecessary since the broker isn't expected to write a subscribe onto network if I am not wrong...

de-sh (approved):
LGTM 👍🏼

rumqttd/src/protocol/v5/subscribe.rs
henil: True, but `protocol` module doesn't make assumption about where it is being used, in future we can extract it out to its own crate like we had `mqttbytes` previously, so its good to keep it updated

PR #538, opened 2022-12-13
feat(link): Add AlertLink to get alerts about router events

henil:
Alerts of different kind are published on topic mentioned below. MQTT wildcards can be used to subscribe to multiple types of alerts at the same time. This alerts are stored seperately from normal MQTT messages so you will need to use `AlertLink` to get `Alerts`
```
Connect => "/alerts/connect"
Disconnect => "/alerts/disconnect"
Subscribe => "/alerts/subscribe"
Unsubscribe => "/alerts/unsubscribe"
``` 
A basic example of how to use this can be found in `rumqttd/examples/alerts.rs`

TLDR:
```rust
// Create a AlertLink from broker using `.alerts` method on it. And pass a MQTT like filter
let alerts = broker.alerts("/alerts/#".to_string()).unwrap();
```

Other changes:
- Adds a validation for `client_id`. Only allowing client_ids with doesn't have `+#/$` in them.


rumqttd/src/router/routing.rs
```diff
@@ -57,6 +58,8 @@ pub struct Router {
     /// List of connections
     meters: Slab<Sender<(ConnectionId, Meter)>>,
     /// List of connections
+    alerts: Slab<(Filter, Sender<(ConnectionId, Alert)>)>,
```
mnpw, line 61: Update desc comment

rumqttd/src/router/routing.rs
```diff
@@ -68,6 +71,8 @@ pub struct Router {
     obufs: Slab<Outgoing>,
     /// Data log of all the subscriptions
     datalog: DataLog,
+    /// Data log of all the subscriptions
```
mnpw, line 74: ++

rumqttd/src/router/routing.rs
```diff
@@ -891,6 +973,51 @@ impl Router {
...
+fn append_to_alertlog(alert: Alert, alertlog: &mut AlertLog) -> Result<Offset, RouterError> {
+    let topic = alert.topic();
+
+    let filter_idxs = alertlog.matches(&topic);
+
+    // Create a filter dynamically
+    let filter_idxs = match filter_idxs {
+        Some(v) => v,
+        None => {
+            let (idx, _cursor) = alertlog.next_native_offset(&topic);
```
mnpw, line 1002: We should not be creating an AlertData for `topic` if it does not already exist in AlertLog. 

Although this will currently not happen since `alerts.topic()` always matches with some filter in AlertsLog, we should enforce otherwise also by omitting None branch altogether. 

Also, `filter_idxs` would never be `None` anyway since `alertlog.matches(..)` always returns a `Some(..)` so we can use .unwrap() on `filter_idxs`.

henil: fixed, also removed creation of unused filter

rumqttd/src/router/alertlog.rs
```diff
@@ -0,0 +1,244 @@
...
+            Self::Disconnect(_) => AlertTopic::Disconnect.to_string(),
+            Self::Subscribe(_, _) => AlertTopic::Subscribe.to_string(),
+            Self::Unsubscribe(_, _) => AlertTopic::Unsubscribe.to_string(),
+        }
+    }
+
+    pub fn topic(&self) -> String {
+        match self {
+            Self::Connect(client_id) => {
+                format!("{}/{}", AlertTopic::Connect.to_string(), client_id)
```
mnpw, line 52: `Alert::topic()` may produce a filter of unexpected behavior for an unusual value of `client_id` instead of a topic

example: for `client_id = "+"`, `Alert::Connect(client_id).topic()` will be `alerts/connect/+`

rumqttd/src/router/routing.rs
```diff
@@ -705,6 +776,17 @@ impl Router {
...
 
+    /// Apply filter and prepare this connection to receive subscription data
+    fn prepare_alert_filter(&mut self, id: ConnectionId, filter: String) {
+        let _ = self.alertlog.next_native_offset(&filter);
+        let offsets = self
+            .alertlog
+            .offsets
+            .entry(filter)
+            .or_insert_with(HashMap::new);
+        offsets.entry(id).or_insert((0, 0));
```
mnpw, line 787: why not use the offset from line 781?

rumqttd/src/server/broker.rs
```diff
@@ -129,6 +130,11 @@ impl Broker {
         Ok(link)
     }
 
+    pub fn alerts(&self, filter: Filter) -> Result<alerts::AlertsLink, alerts::LinkError> {
```
mnpw, line 133: documentation so that a user knows what type of filters are allowed
/// alerts/<alert_kind>/<connection_id>
    /// alert_kind belongs to 
    /// - "connect"
    /// - "disconnect"
    /// - "subscribe"
    /// - "unsubscribe"

PR #541, opened 2022-12-14
refactor!: Remove `Connect` from `ConnectionError::StateError`

henil:
`ConnectionError` has a variant named `ConnectionRefused` which should be used instead.


de-sh:
This seems like the opposite of what we should be doing. Since connection refusals happen only during the initial parts and has nothing to do with MQTT state, we have to remove the `Connect` variant from there.

PR #545, opened 2022-12-28
fix(rumqttc): Add timeouts to network flushes and clean state properly on network failures

henil:
changes made:
- connection_timeout is moved to `NetworkOptions`
- connection_timeout is removed from `Url` feature.
- `ConnectionError` now has specific variant for type of Timeout, `FlushTimeout` and `NetworkTimeout` instead of generic `Timeout` for both.

### DIFF
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -238,13 +238,11 @@ async fn network_connect(options: &MqttOptions) -> Result<Network, ConnectionErr
         Transport::Tcp => {
             let addr = format!("{}:{}", options.broker_addr, options.port);
             let socket = TcpSocket::new_v4()?;
-            if let Some(network_option) = options.network_option {
-                if let Some(send_buff_size) = network_option.tcp_send_buffer_size {
-                    socket.set_send_buffer_size(send_buff_size).unwrap();
-                }
-                if let Some(recv_buffer_size) = network_option.tcp_recv_buffer_size {
-                    socket.set_recv_buffer_size(recv_buffer_size).unwrap();
-                }
+            if let Some(send_buff_size) = options.network_option.tcp_send_buffer_size {
+                socket.set_send_buffer_size(send_buff_size).unwrap();
+            }
+            if let Some(recv_buffer_size) = options.network_option.tcp_recv_buffer_size {
+                socket.set_recv_buffer_size(recv_buffer_size).unwrap();
             }
 
             let stream = socket.connect(addr.to_string().parse().unwrap()).await?;
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -236,13 +236,11 @@ async fn connect(options: &MqttOptions) -> Result<(Network, Incoming), Connectio
 async fn network_connect(options: &MqttOptions) -> Result<Network, ConnectionError> {
     let network = match options.transport() {
         Transport::Tcp => {
-            let addr = options.broker_addr.as_str();
-            let port = options.port;
-            let addr2 = format!("{}:{}", addr, port);
+            let addr = format!("{}:{}", options.broker_addr, options.port);
             let socket = TcpSocket::new_v4()?;
-            socket.set_send_buffer_size(1024).unwrap();
+            // socket.set_send_buffer_size(1024).unwrap();
 
-            let stream = socket.connect(addr2.to_string().parse().unwrap()).await?;
+            let stream = socket.connect(addr.to_string().parse().unwrap()).await?;
             Network::new(stream, options.max_incoming_packet_size)
         }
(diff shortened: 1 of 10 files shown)


rumqttc/src/state.rs
```diff
@@ -52,6 +52,8 @@ pub struct MqttState {
     last_outgoing: Instant,
     /// Packet id of the last outgoing packet
     pub(crate) last_pkid: u16,
+    /// Packet id of the last acked publish
+    pub(crate) last_acked_publish_pkid: u16,
```
tekjar, line 56: Rename this to `last_puback`

rumqttc/src/state.rs
```diff
@@ -124,6 +130,8 @@ impl MqttState {
         self.await_pingresp = false;
         self.collision_ping_count = 0;
         self.inflight = 0;
+        self.write.clear();
+        self.events.clear();
```
tekjar, line 134: Don't clear events

rumqttc/src/eventloop.rs
```diff
@@ -156,7 +157,7 @@ impl EventLoop {
             o = network.readb(&mut self.state) => {
                 o?;
                 // flush all the acks and return first incoming packet
-                network.flush(&mut self.state.write).await?;
+                time::timeout(connection_timeout, network.flush(&mut self.state.write)).await??;
```
tekjar, line 160: Rename `connection_timeout` to `network_timeout` (including error message)

rumqttc/src/eventloop.rs
```diff
@@ -156,7 +158,7 @@ impl EventLoop {
             o = network.readb(&mut self.state) => {
                 o?;
                 // flush all the acks and return first incoming packet
-                network.flush(&mut self.state.write).await?;
+                time::timeout(network_timeout, network.flush(&mut self.state.write)).await??;
```
tekjar, line 161: See if we can have more descriptive errors for flush timeouts

rumqttc/src/eventloop.rs
```diff
@@ -232,13 +234,48 @@ async fn connect(options: &MqttOptions) -> Result<(Network, Incoming), Connectio
     Ok((network, packet))
 }
 
+pub(crate) async fn get_tcp_stream(
+    addrs: String,
```
tekjar, line 238: Rename this to `address` or `host`

rumqttc/src/eventloop.rs
```diff
@@ -232,13 +234,48 @@ async fn connect(options: &MqttOptions) -> Result<(Network, Incoming), Connectio
...
+    network_options: &NetworkOptions,
+) -> io::Result<TcpStream> {
+    let addrs = lookup_host(addrs).await?;
```
tekjar, line 241: Add timeout to `lookup_host`

rumqttc/src/lib.rs
```diff
@@ -432,6 +457,7 @@ impl MqttOptions {
             last_will: None,
             conn_timeout: 5,
             manual_acks: false,
+            network_option: NetworkOptions::new(),
```
tekjar: Move `conn_timetout` to `NetworkOptions` and move `NetworkOptions` as an argument to `Eventloop`

rumqttc/src/eventloop.rs
henil: from its description: https://docs.rs/tokio/1.23.0/tokio/net/fn.lookup_host.html, it doesn't look like it would requires a timeout. should i still add it?

rumqttc/src/eventloop.rs
henil: i have added a variant for `network timeout` and other for `flush timeout`, https://github.com/bytebeamio/rumqtt/pull/545/files#diff-2d1780a6ddafae5dabfb4d74dea0621a46d8505f493a46981589205fffb4faedR34-R37. 

Do you think there can be any other kind of timeout's that we might need to add in future? 

If so then i think having a `Timeout` variant that contains an enum of different kind of variants (i.e. `Timeout(TimeoutType)`) which we can add to in future would be a better idea.

rumqttc/src/eventloop.rs
```diff
@@ -81,7 +85,7 @@ impl EventLoop {
     ///
     /// When connection encounters critical errors (like auth failure), user has a choice to
     /// access and update `options`, `state` and `requests`.
-    pub fn new(options: MqttOptions, cap: usize) -> EventLoop {
+    pub fn new(options: MqttOptions, network_options: NetworkOptions, cap: usize) -> EventLoop {
```
tekjar, line 88: rename `options` to `mqtt_options`

henil: i suppose you meant in eventloop as well, right?

henil:
Changelog was updated after this branch got created so changing it here will cause merge conflict, will add the appropriate entries in a new PR after this is merged. Its ready to merge now.

PR #546, opened 2023-01-03
fix(router): Disconnect existing client on a new connection with same client_id and don't allow empty client_id

henil:


rumqttd/src/router/routing.rs
```diff
@@ -232,6 +232,17 @@ impl Router {
         outgoing: Outgoing,
     ) {
         let client_id = outgoing.client_id.clone();
+        // Check if same client_id already exists and if so, replace it with this new connection
+        // ref: https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718032
+
+        let connection_id = self.connection_map.iter().find(|(id, _)| id == &&client_id);
```
mnpw, line 238: this check should be done after we enter into tracing span (line 247)
also the debug message can omit use of `client_id` variable as it will already be in tracing span's fields

mnpw:
One suggestion, lgtm otherwise

PR #547, opened 2023-01-04
fix: change panic to warn in debug mode

henil:


rumqttd/src/router/routing.rs
```diff
@@ -193,7 +193,10 @@ impl Router {
         }
 
         // A connection should not be scheduled multiple times
-        debug_assert!(self.scheduler.check_readyqueue_duplicates());
+        #[cfg(debug_assertions)]
+        if !self.scheduler.check_readyqueue_duplicates() {
+            warn!("Connection was scheduled multiple times in readyqueue, this shouldn't happen.");
```
mnpw, line 198: Better to log the connection_id's that are duplicate in readyqueue as well so that we can check logs of those connection ids then and there and try to debug

mnpw:
One suggestion, lgtm otherwise

rumqttd/src/router/routing.rs
henil: changed it to print `readyqueue` when duplicates where found, instead of just the duplicates which might be helpful

PR #556, opened 2023-01-16
refactor: custom `NetworkError` and move code into relevant modules

de-sh:
### Changes
- Create custom `NetworkError` type
- Move `readb` from `Network` into `MqttState`
- Move timeouts into `Network::flush()`
- Reorganize imports


tekjar:
It's a good idea to pass `Network` to `MqttState` instead of passing `MqttState` to network. This allows us to easily separate networking implementation from MQTT state management. I'll discuss these changes with @yatinmaan and merge this.

PR #585, opened 2023-03-02
feat: rumqttc Made it possible to convert async MQTT client to sync client using Client::new_from_async

Archipel:
In some cases it is useful to convert an AsyncClient to a sync Client - e.g. to keep the code DRY. This PR makes this possible

Signed off: Thomas Weyn

Attribute: Thomas Weyn

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - `NetworkOptions` added to provide a way to configure low level network configurations (#545)
+- Added `Client::new_from_async` to make it possible to convert from `AsyncClient` to `Client`
 
 ### Changed
 - `options` in `Eventloop` now is called `mqtt_options` (#545)
--- a/rumqttc/src/client.rs
+++ b/rumqttc/src/client.rs
@@ -245,7 +245,11 @@ impl Client {
     /// `cap` specifies the capacity of the bounded async channel.
     pub fn new(options: MqttOptions, cap: usize) -> (Client, Connection) {
         let (client, eventloop) = AsyncClient::new(options, cap);
-        let client = Client { client };
+        Self::new_from_async(client, eventloop)
+    }
+
+    pub fn new_from_async(async_client: AsyncClient, eventloop: EventLoop) -> (Client, Connection) {
+        let client = Client { client: async_client };
         let runtime = runtime::Builder::new_current_thread()
             .enable_all()
             .build()
--- a/rumqttc/src/client.rs
+++ b/rumqttc/src/client.rs
@@ -249,7 +249,9 @@ impl Client {
     }
 
     pub fn new_from_async(async_client: AsyncClient, eventloop: EventLoop) -> (Client, Connection) {
-        let client = Client { client: async_client };
+        let client = Client {
+            client: async_client,
+        };
         let runtime = runtime::Builder::new_current_thread()
             .enable_all()
             .build()
--- a/rumqttd/src/link/mod.rs
+++ b/rumqttd/src/link/mod.rs
@@ -8,4 +8,3 @@ pub mod remote;
 #[cfg(feature = "websockets")]
 pub mod shadow;
 pub mod timer;
-


de-sh:
Hi @Archipel, this PR has more changes than what the title mentions. With regards to async and DRY, I feel the proper way to handle this would be to enable the user to construct the message and make the request separately, thus enabling even more control, what do you think?:
```
let client = AsyncClient::new(...);
let publish = Publish::new(...);
client.request(publish)?;
client.try_request(publish)?;
client.request_timeout(publish, duration)?;
```

Archipel:
Hi @de-sh you mean because of the pkid that is added? You are right about that of course.

I agree that it would be best to separate the message creation & the request itself completely. It should be easy to obtain an available pkid though, so I would still add a function like "next_pkid" to the eventloop or synchronous client too, that can then be used to create the Publish message. Knowing the pkid is important to make it possible to followup on sent packages.

Archipel:
@tekjar to be clear, I agree that the eventloop should generate the pkid to avoid duplicates, but knowing the pkid is important for tracing. Without knowing which pkid corresponds to which message, it is impossible to know which message has been acknowledged and which one hasn't.

To be clear: also in my use case and code changes, the eventloop remains the one that sets the pkid, though this is not enforced indeed:

```
let pkid = mqtt_event_loop.next_pkid();
mqtt_client.publish_with_pkid("topic", mqtt::QoS::AtLeastOnce, false, "payload", pkid).await?;
```

Perhaps a better idea would be to let `publish(...)` return te pkid, but is not trivially possible since the pkid is only set by the eventloop once the Publish is actually handled.

tekjar:
I think it's better to leave `pkid` as an internal detail and develop `tagged` publishes where you can assign a unique id to the publish message to identify it and eventloop can return this id on ack

tekjar:
@Archipel Can you help me understand (preferably with a use case) why you would like to trace the packet until broker with packet id? Packet ids are internal details of the protocol and consumer packet ids are anyway decoupled from publisher packet ids. In that sense, you can never trace a packet end to end. 

Assuming your goal is to find out if a publish has reached the broker, isn't publish+ack with user tokens/tags a good enough proxy for that?

tekjar:
Close enough to user tokens, paho go is returning synchronisation tokens on publish messages instead of packet ids

https://github.com/eclipse/paho.mqtt.golang#error-handling

Archipel:
Sorry for the mate reply, but I guess these synchronisation tokens would be OK as well. I don't have a actual use case requiring the pkid to be followed until the broker.

I'm just not sure why not to use the pkid instead of a new kind of token. It seems like a lot less work. Currently, PubAck also contains the pkid, so it seems logical to just check this with the pkid that has been used to send the message.

Archipel:
Hi @yatinmaan Thanks for the headsup. I completely understand! I have no use case for controlling the pkid, it was just a quick way to solve my problem at the time.

Actually, I had a problem with my implementation of `publish_with_pkid` today as well, namely the fact that I needed to encapsulate the eventloop in an `Arc<Mutex>` to be able to generate the next pkid in a thread safe manner. The locking time could be very long though, when simultaneously polling the eventloop (`eventloop.poll()`) which can block for a while.

For my use case I solved this by separating the pkid-management from the rest of the MqttState and encapsulating this in `Arc<Mutex>`. This way, `next_pkid()` can be called from `client.publish` and the package id can be returned directly:

```
#[derive(Clone, Debug)]
pub struct AsyncClient {
    request_tx: Sender<Request>,
    pkid_manager: Option<ThreadSafeMqttPkidManager>
}

...

    /// Sends a MQTT Publish to the `EventLoop`.
    pub async fn publish<S, V>(
        &self,
        topic: S,
        qos: QoS,
        retain: bool,
        payload: V,
    ) -> Result<u16, ClientError>
    where
        S: Into<String>,
        V: Into<Vec<u8>>,
    {
        let mut pkid = 0;
        if let Some(pm) = &self.pkid_manager {
            pkid = pm.next_pkid()?;
        }
        let publish = create_request_publish(topic, qos, retain, payload, pkid)?;
        self.request_tx.send_async(publish).await?;
        Ok(pkid)
    }
```

You can find the latest changes in commit https://github.com/ASAsense/rumqtt/commit/bae855585c8b701591469efd2ce176ff93dcd708 on my fork. Not sure if this would fit the other use cases, but it sure worked great for mine :-).

PR #588, opened 2023-03-07
fix: remove unnecessary box

### LINKED ISSUE #581: v5 incoming event is boxed?
So in the non-v5 client, he Event struct has a stack allocated Incoming, Outgoing.
In v5, Incoming (but not Outgoing) is boxed.

This makes pattern matching the incoming event in v5 more annoying (and incompatible with the non-v5 one) , because box patterns are not available (and will never be, it looks like rust will try to stabilize deref patterns instead).
As a result, you have to split the match.

IE in non-v5 you can have a match arm of Event::Incoming(Incoming::ConnAck(ack)) 
in v5, you can't, because it would have to be Event::Incoming(box Incoming::ConnAck(ack))

As far as i can tell, there is no totally obvious reason why Incoming is boxed in v5.
Removing the box and updating 3 lines of code to not use boxing still passes all tests, and works fine in my (admittedly simple) app.

Is there a particular reason it's boxed?
If not, i'll send along a pull request to unbox it.

de-sh:
Please read `CONTRIBUTING.md` if you are contributing for the first time.

# Description

Remove use of box for enclosing `Incoming` in `mod v5`

Fixes #581 

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

- [x] Make an entry to `CHANGELOG.md`.

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Remove `Box` on `Event::Incoming`
 
 ### Deprecated
 
--- a/rumqttc/examples/async_manual_acks_v5.rs
+++ b/rumqttc/examples/async_manual_acks_v5.rs
@@ -58,7 +58,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
         println!("{event:?}");
 
         if let Event::Incoming(packet) = event {
-            let publish = match *packet {
+            let publish = match packet {
                 Packet::Publish(publish, _) => publish,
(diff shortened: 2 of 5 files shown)


rumqttd/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Remove `Box` on `Event::Incoming`
```
henil, line 13: This should be in `rumqttc/CHANGELOG.md` instead

PR #589, opened 2023-03-09
MQTT5 Features

yatinmaan:
Depends on #577


rumqttc/examples/async_manual_acks_v5.rs
```diff
@@ -77,7 +77,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
 async fn requests(client: &AsyncClient) {
     for i in 1..=10 {
         client
-            .publish("hello/world", QoS::AtLeastOnce, false, vec![1; i])
```
tekjar, line 80: Split client APIs to `publish` and `publish_with_properties`, `subscribe`, `subscribe_with_properties` etc

rumqttc/examples/async_manual_acks_v5.rs
```diff
@@ -12,7 +12,7 @@ fn create_conn() -> (AsyncClient, EventLoop) {
     mqttoptions
         .set_keep_alive(Duration::from_secs(5))
         .set_manual_acks(true)
-        .set_clean_session(false);
+        .set_clean_start(false);
```
tekjar, line 15: Can we rename this back to `clean_session` as it's the correct term in MQTT's context?

rumqttc/src/v5/framed.rs
```diff
@@ -15,13 +15,13 @@ pub struct Network {
     /// Buffered reads
     read: BytesMut,
     /// Maximum packet size
-    max_incoming_size: usize,
+    max_incoming_size: Option<usize>,
```
tekjar, line 18: This shouldn't be optional. There are always default values for `max_incoming_size`. Either via config or server connack. Does the spec have any restrictions about making this an optional filed?

rumqttc/examples/async_manual_acks_v5.rs
yatinmaan: `clean_start` is a new slightly different field in v5.

rumqttc/src/v5/framed.rs
yatinmaan: Added `default_max_incoming_size` to mqtttoptions

PR #592, opened 2023-03-17
feat(prometheus): add a field in config file to specify listen socket address

henil:
# Description

- When `port` is specified `listen` is ignored.
- When both are not specific default socket address used is `127.0.0.1:9042`
- `port` is deprecated and will be removed in future breaking release.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

# Checklist:

- [x] Make an entry to `CHANGELOG.md`.


mnpw:
LGTM

PR #593, opened 2023-03-22
fix(segments): Calculate `absolute_offset` and `Position` properly 

h3nill:
## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:


rumqttd/src/segments/mod.rs
```diff
@@ -274,7 +276,7 @@ where
                 // debug!("start: {:?}, end: ({}, {}) done", orig_cursor, cursor.0, absolute_offset);
                 Ok(Position::Done {
                     start,
-                    end: (cursor.0, absolute_offset),
+                    end: (cursor.0 + 1, absolute_offset),
```
mnpw, line 279: This seems incorrect. Added failing case [here](https://github.com/bytebeamio/rumqtt/commit/2149d949a1f7ddfd6483a19a813ef2d99edd8762).

In test `active_segment_appends_and_reads_works` [we should get](https://github.com/bytebeamio/rumqtt/pull/593/files#diff-de9a4fd1a5e1abe08ac05206b8f03bc90d77d8d2ab40dc6d8fa903690d63628cR391-R392) `Done(0, 10)` instead of `Done(1, 10)`. This is because first segment in the CommitLog is still 90% empty. So our next read on the CommitLog should start from (0, 10). `cursor.0` should only increment when a segment is read fully. This is done [automatically](https://github.com/bytebeamio/rumqtt/pull/593/files#diff-de9a4fd1a5e1abe08ac05206b8f03bc90d77d8d2ab40dc6d8fa903690d63628cL199-L207) within `CommitLog::readv()`.

henil: good catch, from reading comments at https://github.com/bytebeamio/rumqtt/pull/593/files#diff-de9a4fd1a5e1abe08ac05206b8f03bc90d77d8d2ab40dc6d8fa903690d63628cL226-R236 i thought we only return `SegementPosition::Done` when we reach to the end of an segment. 

But when segment we are reading from is the active segment its not true because we return based on the `len` which is different from capacity of the segment. And when i saw `Done` inside `active_segment_appends_and_reads_works` I thought it was because of segment getting full and missed that segment has capacity to hold 100 elements not 10.

But the previous code is not correct as well because the invariant on `end` is that it is the cursor from where next read should begin and in few cases (see the test i added) it was breaking.

henil: fixed this by checking if current segment can hold more data or not.

rumqttd/src/segments/mod.rs
```diff
@@ -272,10 +274,17 @@ where
...
+                    Ok(Position::Done {
+                        start,
+                        end: (cursor.0, absolute_offset),
+                    })
+                } else {
+                    Ok(Position::Done {
+                        start,
+                        end: (cursor.0 + 1, absolute_offset),
+                    })
+                }
```
henil, line 287: if `active_segment().size()` >= `self.max_segment_size` on the next [append to commitlog](https://github.com/bytebeamio/rumqtt/blob/3f388b00f2e4eb8d7e710673bbc0dcb137244bbc/rumqttd/src/segments/mod.rs#L161) we will call `apply_retention` and new segment would be created.

swanandx:
@henil is the bug fix still required? most of the changes already existed in main branch, I resolved the merge conflicts, can you have a look at current state of PR and lmk if the changes are still required?

if not required, can we close the PR?

PR #600, opened 2023-03-31
Added support for MQTT5 features to v5 client

yatinmaan:
## Type of change

 - Added support for MQTT5 features to v5 client

## Checklist:


rumqttc/src/v5/mod.rs
```diff
@@ -53,31 +54,27 @@ pub struct MqttOptions {
...
     /// while retransmitting pending packets
     pending_throttle: Duration,
-    /// maximum number of outgoing inflight messages
-    inflight: u16,
     /// Last will that will be issued on unexpected disconnect
     last_will: Option<LastWill>,
     /// Connection timeout
     conn_timeout: u64,
+    /// Default value of for maximum incoming packet size.
+    /// Used when `max_incomming_size` in `connect_properties` is NOT available.
```
thienandangthanh, line 74: `The ConnectionProperties` struct doesn't have `max_incomming_size`.
Instead, it has `max_packet_size`.

To avoid confusion, this comment should be updated to reflect this information.

PR #602, opened 2023-04-05
feat: Filter based segment

mnpw:
## Changes:

- Add option to override segment config based on a filter

**Breaking:** 
- Remove unused config fields
- Rename `router.max_read_len` to `router.max_outgoing_packet_count` in config and set its value from 10240 to 200

## Checklist:


rumqttd/src/lib.rs
```diff
@@ -102,10 +102,8 @@ pub struct BridgeConfig {
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct ConnectionSettings {
     pub connection_timeout_ms: u16,
-    pub throttle_delay_ms: u64,
     pub max_payload_size: usize,
-    pub max_inflight_count: u16,
-    pub max_inflight_size: usize,
+    pub max_inflight_count: usize,
```
henil, line 106: I think we should update this variable name to a more descriptive `max_connection_buffer_len` that we use internally in `Network`, so it is not confused with max allowed inflight publish messages which is currently a constant set to 100 (i.e. MAX_INFLIGHT in `src/router/iobufs.rs`).

rumqttd/rumqttd.toml
```diff
@@ -5,11 +5,13 @@ id = 0
...
 [router]
 id = 0
-instant_ack = true
+max_connections = 10010
+max_outgoing_packet_count = 200
 max_segment_size = 104857600
 max_segment_count = 10
-max_read_len = 10240
-max_connections = 10010
+    [router.custom_segment.'/tenants/+/+/+/actions']
```
henil, line 12: do we want to also limit the size of segment when subscription is on filter `/tenants/+/+/+/actions`

rumqttd/src/router/logs.rs
```diff
@@ -234,7 +230,22 @@ impl<T> Data<T>
...
+
+        // Override segment config for selected filter
+        if let Some(config) = &router_config.custom_segment {
+            for (f, segment_config) in config {
+                if matches(filter, f) {
+                    info!("Overriding segment config for filter: {}", filter);
+                    max_segment_size = segment_config.max_segment_size;
+                    max_mem_segments = segment_config.max_segment_count;
+                }
+            }
```
henil, line 245: nitpick: we can break out of this loop once we found a match

rumqttd/rumqttd.toml
mnpw: Yep, smaller segments and less in number for stream that contains device actions

henil:
We have merged this internally, we will create a rumqttd release with this changes soon.

PR #604, opened 2023-04-06
Add qos2 support in Broker

swanandx:
- Enable QoS2 support
- Fix a bug in scheduler: We had to wait until a new packet is there so the connection will get scheduled and then we send the pending ack ( PubRec ) of older packet. 

Even if the scheduler's behavior is changed for a particular case, it shouldn't affect the existing users. Therefore, this isn't a breaking change.

## Type of change

## Checklist:


rumqttd/src/router/scheduler.rs
```diff
@@ -160,7 +160,9 @@ impl Tracker {
                 self.status = Status::Ready;
                 Some(previous)
             }
-            ScheduleReason::IncomingAck if previous == PauseReason::InflightFull => {
+            // There might be a case where we got IncomingAck but our inflight wasn't full
+            // So we set status to Ready in case of IncomingAck unless reason in Busy
+            ScheduleReason::IncomingAck if previous != PauseReason::Busy => {
                 self.status = Status::Ready;
                 Some(previous)
             }
```
henil, line 168: i am not sure big off an affect this would be but, this would cause status to be `Ready` even for incoming `PubAck`s which is not really necessary (which could lead to `connection_id` being in readyqueue multiple times). 

we could split `IncomingAck` reason to 2 parts `IncomingAck` and `IncomingAckNeedsReply` (ok no really this name xP) and only reschedule tracker when its really required. 

@mnpw wdyt?

rumqttd/src/router/scheduler.rs
```diff
@@ -160,7 +160,9 @@ impl Tracker {
...
-            ScheduleReason::IncomingAck if previous == PauseReason::InflightFull => {
+            // There might be a case where we got IncomingAck but our inflight wasn't full
+            // So we set status to Ready in case of IncomingAck unless reason in Busy
```
henil, line 164:
```suggestion
            // There might be a case where we got IncomingAck but our inflight wasn't full, and incase of QoS2 we are expected to send PubRel as soon as possible
```

rumqttd/src/router/routing.rs
```diff
@@ -703,8 +699,13 @@ impl Router {
                             break;
                         }
                     };
+                    self.scheduler.reschedule(id, ScheduleReason::IncomingAck);
+                }
+                Packet::PubComp(pubcomp, _) => {
+                    // Should we just log with info! instead of creating a span?
+                    let span = tracing::info_span!("pubcomp", pkid = pubcomp.pkid);
+                    let _guard = span.enter();
                 }
```
henil, line 708: I assume current creating this span is redundant because we are not logging any `Event` after we create the span. I also think we should add a `info!` here.

rumqttd/src/router/routing.rs
```diff
@@ -719,7 +720,7 @@ impl Router {
                     break;
                 }
                 incoming => {
-                    warn!(packet=?incoming, "Packet not supported by router yet" );
+                    warn!(packet=?incoming, "Unexpected packet received" );
```
henil, line 723:
```suggestion
                    warn!(packet=?incoming, "Unexpected packet received, ignoring the packet." );
```

rumqttd/src/router/scheduler.rs
swanandx: how does `Ready` status after getting `PubAck` cause `connection_id` to be in readyqueue multiple times?

henil: so where we [call `try_ready`](https://github.com/bytebeamio/rumqtt/blob/fe8a8e3c12274846965f3bbcf504706cac654d6e/rumqttd/src/router/scheduler.rs#L64-L67) we are adding the connection_id to `readyqueue` if call returned a `Some(T)`. 

And there can be a case where this `reschedule` is called multiple times in [this loop](https://github.com/bytebeamio/rumqtt/blob/fe8a8e3c12274846965f3bbcf504706cac654d6e/rumqttd/src/router/routing.rs#LL645C1-L645C21)

henil: yup

rumqttd/src/router/routing.rs
```diff
@@ -761,8 +785,22 @@ impl Router {
...
+                    let outgoing = self.obufs.get_mut(id).unwrap();
+                    let pkid = pubcomp.pkid;
+                    if outgoing.register_pubcomp(pkid) {
+                        error!(
+                            pkid,
+                            "ack received for pkid {}, but the pkid didn't exists!", pkid
+                        );
+                        // shall we ignore it or diconnect the client?
+                        // disconnect = true;
+                        // break;
```
henil, line 801: yeah to be consistent with other places, we should disconnect the client as there might be some state mismatch if we just ignore it.

rumqttd/src/router/routing.rs
```diff
@@ -761,8 +785,22 @@ impl Router {
                             break;
                         }
                     };
+                    self.scheduler.reschedule(id, ScheduleReason::IncomingAck);
+                }
+                Packet::PubComp(pubcomp, _) => {
+                    info!(pkid = pubcomp.pkid, "received pubcomp");
```
henil, line 791: make it a `info_span` to make it consistent with other acks

rumqttd/src/router/iobufs.rs
```diff
@@ -175,6 +179,18 @@ impl Outgoing {
...
     }
 
+    pub fn register_pubrec(&mut self, pkid: u16) {
+        // NOTE: we can return true of false
+        // to indicate whether this is duplicate or not
+        self.pending_acks.insert(pkid);
+    }
+
+    // PubComp can be out of order!
+    // return false if pkid was not present
```
henil, line 189: because we don't support out of order acks for other types of acks i would prefer if we keep the behaviour consistent, wdyt?

swanandx: But [standards](https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718105) don't have any rules for ordering on pubcomp, therefore I thought it implies they can be out or order as well.

rumqttd/src/router/routing.rs
swanandx: okie!

rumqttd/src/router/iobufs.rs
```diff
@@ -72,6 +74,7 @@ impl Outgoing {
         let (handle, rx) = flume::bounded(MAX_CHANNEL_CAPACITY);
         let data_buffer = VecDeque::with_capacity(MAX_CHANNEL_CAPACITY);
         let inflight_buffer = VecDeque::with_capacity(MAX_INFLIGHT);
+        let pending_acks = VecDeque::with_capacity(100); // cuz acklog also has 100
```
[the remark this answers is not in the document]
swanandx: `pending_acks` will only store PubRels, so I think AckLog size is more relevant than Inflight?

rumqttd/src/router/iobufs.rs
henil: yes they can be out of order, but [we already don't allow out of order puback's and pubrec's](https://github.com/search?q=repo%3Abytebeamio%2Frumqtt%20ooo&type=code) so to keep things consistent i think we should have same behaviour here as well.

rumqttd/src/router/iobufs.rs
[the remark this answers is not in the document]
henil: in the worst cast there can be `MAX_INFLIGHT` qos2 publishes which means at max `MAX_INFLIGHT` `pending_acks`. Maybe acklogs size is should be `MAX_INFLIGHT` as well.

anyways it doesn't really matter because all of them are 100, its just for better readability.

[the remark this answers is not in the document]
swanandx: If there are `Publish`s in Inflight, that means we haven't got `PubRec` yet, which implies there won't `PubRel`, so nothing in pending acks. 
You can think of it like, we add `PubRel` to pending_ack, once we remove `Publish` from Inflight. Anyways, will change it to `MAX_INFLIGHT` 🫡

rumqttd/src/router/iobufs.rs
swanandx: I believe we should adhere to standards and let them be out of order, otherwise there is chance someone facing an issue as they though it's okie to send PubComp out of order. Therefore, it seems to be a better choice for long term, wdyt?

swanandx: Anyway, for now, I made the changes to check for order. We will see how it goes later.

henil (commented):
LGTM

PR #611, opened 2023-04-20
feat: Add support for HTTP(s) proxy

### LINKED ISSUE #598: Guidance on using rumqttc websockets over http proxy
I do realize that the library does not contain http proxy support per design, although I don't agree with it (rumqttc is advertised as a high level interface, and therefore it should expose a proxy configuration). But anyway.

I'm still trying to understand how I achieve a HTTP proxy connection given the current API, when using WS or WSS. Do I follow the tungstenite approach, where I connect to the proxy, build a "connect" http request, and connect to the final server?  How do I wire something like this into the client?

### LINKED ISSUE #608: Expose `Connection` on eventloop initialization
This would allow calling different initialization routines (e.g. as using http proxy).

I am willing to implement it.

sushi-shi:
Issue: #608

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,6 +68,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
+name = "async-http-proxy"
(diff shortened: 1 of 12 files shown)


tekjar:
@sushi-shi Apologies for the delay. We'll close this this week

PR #619, opened 2023-05-17
docs(rumqttc/examples): (de)serialization example

### LINKED ISSUE #617: rumqttc examples | Payload (de)serialization
I've been struggling to serialize (`Vec<u8>`) and deserialize (`Bytes`) a struct into a valid MQTT `payload` since I just recently started using Rust and this library (great, but I'm not an expert).

Can we include an easy example demonstrating how to serialize and deserialize a struct into a valid MQTT payload?

What follows can be a good place to start ([`bincode`](https://crates.io/crates/bincode) + [`serde`](https://crates.io/crates/serde))?

```rust
use rumqttc::Incoming;
use rumqttc::{Client, Event, MqttOptions, QoS};
use serde::{Deserialize, Serialize};
use std::thread;
use std::time::{Duration, SystemTime};

#[derive(Serialize, Deserialize, Debug)]
struct Message {
    i: usize,
    time: SystemTime,
}

impl From<&[u8]> for Message {
    fn from(value: &[u8]) -> Self {
        bincode::deserialize(value).unwrap()
    }
}

impl From<Message> for Vec<u8> {
    fn from(value: Message) -> Self {
        bincode::serialize(&value).unwrap()
    }
}

fn main() {
    let mqqt_opts = MqttOptions::new("ciao", "localhost", 1883);

    let (mut client, mut connection) = Client::new(mqqt_opts, 10);
    client.subscribe("hello/rumqtt", QoS::AtMostOnce).unwrap();
    thread::spawn(move || {
        for i in 0..10 {
            let message = Message {
                i,
                time: SystemTime::now(),
            };

            client
                .publish("hello/rumqtt", QoS::AtLeastOnce, false, message)
                .unwrap();
            thread::sleep(Duration::from_millis(100));
        }
    });

    // Iterate to poll the eventloop for connection progress
    for notification in connection.iter() {
        if let Event::Incoming(Incoming::Publish(message)) = notification.unwrap() {
            let message: Message = message.payload.as_ref().into();
            println!("Payload = {:?}", message);
        }
    }
}
```

carlocorradini:
Fix #617

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -272,6 +272,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "bincode"
+version = "1.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b1f45e9417d87227c7a56d22e471c6206462cba514c7590c09aff4cf6d1ddcad"
+dependencies = [
+ "serde",
+]
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1956,6 +1965,7 @@ version = "0.21.0"
 dependencies = [
  "async-http-proxy",
  "async-tungstenite",
+ "bincode",
  "bytes",
  "color-backtrace",
  "flume",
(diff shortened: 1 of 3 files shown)


carlocorradini:
@yatinmaan Let me know what you think 😊

PR #622, opened 2023-05-22
rumqttc: Enforce max packet size

yatinmaan:
## Type of change

   - ~~Stop `.poll()` from reconnecting after an `disconnect()` call~~
   - Enforce `max_outgoing_packet_size` on v3 client

## Checklist:

### DIFF
--- a/rumqttc/src/state.rs
+++ b/rumqttc/src/state.rs
@@ -30,6 +30,8 @@ pub enum StateError {
     EmptySubscription,
     #[error("Mqtt serialization/deserialization error: {0}")]
     Deserialization(#[from] mqttbytes::Error),
+    #[error("Cannot recieve packet of size '{pkt_size:?}'. It's greater than the client's maximum packet size of: '{max:?}'")]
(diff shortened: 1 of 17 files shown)


rumqttc/src/state.rs
```diff
@@ -72,13 +74,15 @@ pub struct MqttState {
...
+    /// Maximum outgoing packet size, set via MqttOptions
+    pub max_outgoing_packet_size: usize,
 }
 
 impl MqttState {
     /// Creates new mqtt state. Same state should be used during a
     /// connection for persistent sessions while new state should
     /// instantiated for clean sessions
-    pub fn new(max_inflight: u16, manual_acks: bool) -> Self {
+    pub fn new(max_inflight: u16, manual_acks: bool, max_outgoing_packet_size: usize) -> Self {
```
henil, line 85: we are exposing `MqttState`, so is also considered a breaking change. Its very unlikely that someone is using this but can you still mention this in CHANGELOG.

henil (commented):
one comment, else LGTM

PR #632, opened 2023-06-01
Subscription IDs support in broker

swanandx:
Client can specify subscription identifier while subscribing to a topic. While forwarding the Publish packets that matches particular subscription, broker will add the respective identifier ( if exists ) in publish properties.
## Type of change

## Checklist:


henil (commented):
LGTM

Issue #634, opened 2023-06-06
Rumqttd: Support for v3.1.1 & v5 on the same port


Nickztar:
# Expected Behavior

Most brokers support both protocols on the same port. Allowing the connect packet determine which protocol is used (If both are supported...).

# Current Behavior

This currently is not supported and you are required to start two ports with one protocol on each.

I've tried my best at getting a prototype to work for this but having some trouble. Also don't really want to rewrite large parts of the current implementation of this is not something that is being planned or wanted.

swanandx:
Hey, is there any significant advantage of having them on same port?

 Currently, as `Network<P>`  has the `P: Protocol`, it knows how to read the packets ( including connect ), but if we want to change it to read one packet first to determine the protocol, that _might_ affect us in future in case we want to add support for different protocol. Because then we won't know how to read the very first packet ( e.g. connect  )

Nickztar:
Hi. 
It does somewhat complicate the configuration process. Having one port allows ease in configuration and easier support for the clients connecting. It is not a deal breaker and more of a nice to have.

As for reading the packets. Since both 3.1.1 and 5 have similar header and start of connect I tried to have P: Protocol be changed after reading as far but my rust knowledge did not get me very far and struggled to get a understandable and robust solution. Maybe there is a smarter way, will try to look into how other brokers handles this 👍🏼

swanandx:
> As for reading the packets. Since both 3.1.1 and 5 have similar header and start of connect I tried to have P: Protocol be changed after reading

Agreed. I wanted to say, if we plan to add other protocol to the broker ( as broker is _generic_ for any pub/sub protocol ) which might have a different start of connect, then it would add up some complexity 

If it doesn't provide much benefit, then let's keep it on different ports, atleast for now. Will see how it goes later 😁

PR #641, opened 2023-06-10
feat: Support both v4 & v5 on same port

Nickztar:
This PR adds the ability to specify a single port and have that port support both v.3.1.1 and v5. It also allows a simpler configuration for tcp connection. The implementation works by adding a new "protocol" which reads the connect packet and determines which protocol the client is using. For the next packets and the continuation of the connect packet they are forwarded to the determined protocol (If it is supported as per the configuration). 

As the comment `dynamic.rs` indicates. This does add a small bit of overhead in reading the fixed header twice, will try to figure out a smarter way to peek at the packet. 

I decided to keep the previous way of configuring to not break existing configurations. It could be possible and possibly cleaner to remove the old way (v4.1 & v5.1) in favor of the dynamic tcp configuration. 

Open to feedback on the solution and this might not be something of interest which is fine. 

This solves #634 and the discussion in #633.

## Type of change

# Checklist:

### DIFF
--- a/rumqttd/rumqttd.toml
+++ b/rumqttd/rumqttd.toml
@@ -78,6 +78,22 @@ next_connection_delay_ms = 1
     max_inflight_count = 500
     max_inflight_size = 1024
 
+[tcp.1]
+name = "tcp-1"
+listen = "0.0.0.0:8883"
+sub_protocol = "all"
+#sub_protocol = "v3.1.1"
+#sub_protocol = "v5"
+next_connection_delay_ms = 1
+    [tcp.1.connections]
+    connection_timeout_ms = 60000
+    max_client_id_len = 256
+    throttle_delay_ms = 0
+    max_payload_size = 20480
+    max_inflight_count = 500
+    max_inflight_size = 1024
+    dynamic_filters = true
+
 [prometheus]
 listen = "127.0.0.1:9042"
 interval = 1
--- a/rumqttd/src/lib.rs
+++ b/rumqttd/src/lib.rs
@@ -42,7 +42,8 @@ pub use server::Broker;
 pub struct Config {
     pub id: usize,
     pub router: RouterConfig,
-    pub v4: HashMap<String, ServerSettings>,
+    pub tcp: Option<HashMap<String, ServerSettings>>,
+    pub v4: Option<HashMap<String, ServerSettings>>,
     pub v5: Option<HashMap<String, ServerSettings>>,
     pub ws: Option<HashMap<String, ServerSettings>>,
     pub cluster: Option<ClusterSettings>,
@@ -101,6 +102,8 @@ pub struct ServerSettings {
     pub tls: Option<TlsConfig>,
     pub next_connection_delay_ms: u64,
     pub connections: ConnectionSettings,
+    #[serde(default)]
+    pub sub_protocol: SubProtocol,
 }
 
 #[derive(Debug, Serialize, Deserialize, Clone)]
@@ -191,4 +194,15 @@ pub enum MetricType {
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct MetricSettings {
     push_interval: u64,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+pub enum SubProtocol {
+    #[serde(rename = "all")]
+    #[default]
+    All,
+    #[serde(rename = "v3.1.1")]
+    V4,
+    #[serde(rename = "v5")]
+    V5,
 }
--- a/rumqttd/src/main.rs
+++ b/rumqttd/src/main.rs
@@ -88,10 +88,12 @@ fn main() {
 
 // Do any extra validation that needs to be done before starting the broker here.
 fn validate_config(configs: &rumqttd::Config) {
-    for (name, server_setting) in &configs.v4 {
-        if let Some(tls_config) = &server_setting.tls {
-            if !tls_config.validate_paths() {
-                panic!("Certificate path not valid for server v4.{name}.")
+    if let Some(v4) = &configs.v4 {
+        for (name, server_setting) in v4 {
+            if let Some(tls_config) = &server_setting.tls {
+                if !tls_config.validate_paths() {
+                    panic!("Certificate path not valid for server v4.{name}.")
+                }
             }
         }
     }
@@ -111,6 +113,16 @@ fn validate_config(configs: &rumqttd::Config) {
             if let Some(tls_config) = &server_setting.tls {
                 if !tls_config.validate_paths() {
                     panic!("Certificate path not valid for server ws.{name}.")
+                }
+            }
+        }
+    }
+
+    if let Some(tcp) = &configs.tcp {
+        for (name, server_setting) in tcp {
+            if let Some(tls_config) = &server_setting.tls {
+                if !tls_config.validate_paths() {
+                    panic!("Certificate path not valid for server tcp.{name}.")
                 }
             }
         }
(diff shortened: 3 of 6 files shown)


swanandx:
I just glanced over the changes, this is really clever way to so it haha, glad that you figured it out! :smile: 

The idea of **_adding_** the ability to specify we want to use single port for both v4/5 is noicee! :rocket: 
But, `DynamicProtocol` _might_ be confusing for other users. I'm curious what are the issues if we try to replace the `Network<P>`? 

I will go over the PR in depth later, awesome work :100:

Nickztar:
Hi! Though about the problem when going to bed last night and had to get up to try it. 😄 

Agree that `DynamicProtocol` is not a optimal name. Might have to think a bit about that. 

From my understanding, `Network<P>` should be non the wiser since Dynamic implements `P: Protocol` in this case. If in the future another protocol is added it can either be added to Dynamic to have it support that (If possible?) or just have the other Protocol run on a different port. Since I am guessing both v4 & v5 still should still be supported in that case, any other methods added to Protocol trait should be able to be supported by the Dynamic protocol.

swanandx:
> Hi! Though about the problem when going to bed last night and had to get up to try it.

<br>


Oh btw, i remember you mentioning about taking look at how other brokers implement it, would love to know if you got any insights

Nickztar:
Exactly like that! 😆 

Ended up just trying to connect to different brokers and changing my protocol version. Most of them just worked. Tried to go though some HiveMQ code but brain was hurting from trying to find anything... But it seems that most brokers just have one large protocol that just checks what protocol level it is (most by doing something like my solution with determining by connect) and then takes different code paths depending on the level.

In my opinion my solution feels very optimal, apart from having a lot of code duplication (fixed header etc which is shared between v4 & v5 if im not mistaken).

swanandx:
I did went through HiveMQ & emqx code ( not that i know java / erlang ) and figured out that they store protocol level with client and pass it down when serializing/deserializing, as you mentioned. ( PS: I used GitHub android app just to increase my pain :rofl: )
But in our case, we don't store the protocol level at client, instead out `Network` has something that implements `Protocol`, so your proposed solution seems to be great choice. Though the implementation has scope for improvements, including what to name the protocol haha.

btw, will the following be better than `sub_protocol` field?
- `tcp.1` -> dynamically figure out protocol
- `tcp.v4.1` / `tcp.v5.1` -> specify protocol
( same can be done for `ws` )

what do you think @tekjar ?

Nickztar:
Agree that there is room for improvement.

As for your example there, does that really make it simpler? In my opinion that feels a bit more difficult? Maybe my toml knowledge is just lacking 😀

`sub_protocol` thing is something inspired by how our HiveMQ config at work looks like. But in that case it's a list of protocols. So might need a rename.

rumqttd/src/lib.rs
```diff
@@ -192,3 +195,14 @@ pub enum MetricType {
 pub struct MetricSettings {
     push_interval: u64,
 }
+
+#[derive(Debug, Serialize, Deserialize, Clone, Default)]
+pub enum SubProtocol {
```
henil, line 200: i think we are using this in context of Supported protocols only, so the name `SubProtocol` doesnt make sense. Maybe we can name it `SupportedProtocol` itself?

rumqttd/rumqttd.toml
```diff
@@ -78,6 +78,22 @@ next_connection_delay_ms = 1
     max_inflight_count = 500
     max_inflight_size = 1024
 
+[tcp.1]
```
henil, line 81: `tcp` is a bit confusing here and doesn't fit in with `protocol.server` convention, maybe we should name it `dynamic` as well?

Nickztar: Well I would say that tcp makes a lot more sense than `protocol.server`. Especially if you look at the other types. There WS, Prometeus, Bridge etc. My initial plan was to get rid of protocol.server but to maintain backwards compatible they are left in.

Consider the following config to replace the one that is current default:
```id = 0

[router]
id = 0
instant_ack = true
max_segment_size = 104857600
max_segment_count = 10
max_read_len = 10240
max_connections = 10010

[tcp.1]
name = "tcp-1"
listen = "0.0.0.0:1884"
sub_protocol = "all"
#sub_protocol = "v3.1.1"
#sub_protocol = "v5"
next_connection_delay_ms = 1
    [tcp.1.connections]
    connection_timeout_ms = 60000
    max_client_id_len = 256
    throttle_delay_ms = 0
    max_payload_size = 20480
    max_inflight_count = 500
    max_inflight_size = 1024
    dynamic_filters = true

[ws]
name = "ws-1"
listen = "0.0.0.0:8080"
sub_protocol = "all"
#sub_protocol = "v3.1.1"
#sub_protocol = "v5"
next_connection_delay_ms = 1
    [tcp.1.connections]
    connection_timeout_ms = 60000
    max_client_id_len = 256
    throttle_delay_ms = 0
    max_payload_size = 20480
    max_inflight_count = 500
    max_inflight_size = 1024
    dynamic_filters = true

[prometheus]
listen = "127.0.0.1:9042"
interval = 1

[console]
listen = "0.0.0.0:3030"```

This to me feels a lot more obvious about which parts does what. Maybe I am alone in thinking this? 😄

Comment (henil) [reply rumqttd/rumqttd.toml:81]: yeah actually i just realized about the `ws` part, with that it does make more sense. 

But i am not sure whats more appropriate `tcp`, `ws`, etc... or having `v4`, `v5` etc... and specifying transport layer inside. 

will get back you after discussing with team. Marking this as resolved for now.

File: rumqttd/src/lib.rs (hunk 1 above)
Comment (Nickztar) [reply rumqttd/src/lib.rs:200]: SupportedProtocol makes sense. Will change that later today 👍🏼

**Comment by henil** [comment]
so i discussed this with team and we think having the protocol version on top is more intuitive, for example:

```toml
[v4]
[v4.tcp.1]
name = "v4-tcp-1"
...

[v4.ws.1]
name = "v4-ws-1"
...

[v5]
[v5.tcp.1]
name = "v5-tcp-1"
...

[v5.ws.1]
name = "v5-ws-1"
...

[dynamic]
[dynamic.tcp.1]
name = "dynamic-tcp-1"
...

[dynamic.ws.1]
name = "dynamic-ws-1"
...
```

Also I think we can remove the `SubProtocol` logic, because we use dynamic only when we want to support multiple protocol at the same time.

It would make things more consistent related to `ws` and would make it trivial to add `MQTTv5` for ws in #633

wdyt @Nickztar @swanandx ?

Nickztar:
Not bad at all. This would be a breaking change right? 

Will look into implementing it later today! Any idea about how the type for this would look?

henil:
Actually because that would be a breaking change, lets merge this like this only. 

We can take that up in a different PR where we can experiment more on this.

Nickztar:
I'll change it to SupportedProtocol instead of SubProtocol to keep it clean until we actually change it. 👍🏼

henil:
is this working for you @Nickztar?

trying to connect using mosquitto is failing for me with error:
```log
  2023-06-17T07:47:47.737667Z ERROR rumqttd::server::broker: Remote link error, error: Network(Protocol(BoundaryCrossed(257)))
    in rumqttd::server::broker::remote_link with tenant_id: None
```

and its originating from call to [`read_mqtt_string`](https://github.com/bytebeamio/rumqtt/pull/641/files#diff-754f12cac98cdccab33960c5de1f217df5f0ce28b708e4573c3179e0591b2a2aR281)

Nickztar:
@henil It is working great for me. That error looks like when you try to connect with TLS without having server configured to TLS. Which is exactly what mosquitto does when you run on port 8883, which is a terrible example port chosen by me. If you change the port in rumqtt.toml to something else, like 8884, and then try with mosquitto it should work.

Nickztar:
@henil @swanandx Would you like me to merge main and add support for configuring websockets in the same way you can for tcp here?

PR #642, opened 2023-06-11
Allow modifying websocket requests.

reitermarkus:
I am trying to connect to an AWS IoT Shadow via MQTT and SigV4 but couldn't find any way to add additional headers to the request. This adds a way to modify the request before connecting.

Example usage:

```rust
    mqtt_options.set_request_modifier(move |request| {
      let credential_provider = credential_provider.clone();
      let region = region.clone();

      async move {
          let request_config = RequestConfig {
              request_ts: SystemTime::now(),
              region: &SigningRegion::from(region.clone()),
              service: &SigningService::from_static("iotdata"),
              payload_override: None,
          };

        let (parts, body) = request.into_parts();
        let mut request: http::Request<SdkBody> = http::Request::from_parts(parts, SdkBody::empty());

        let signer = SigV4Signer::new();
        signer.sign(
          &OperationSigningConfig::default_config(),
          &request_config,
          &credential_provider.provide_credentials().await.unwrap(),
          &mut request,
        ).unwrap();

        let (parts, _) = request.into_parts();
        http::Request::from_parts(parts, body)
      }
    });
```

## Type of change

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Added `MqttOptions::set_request_modifier` for setting a handler for modifying a websocket request before sending it.
+
 ### Changed
 
 ### Deprecated
--- a/rumqttc/src/v5/eventloop.rs
+++ b/rumqttc/src/v5/eventloop.rs
@@ -332,6 +332,10 @@ async fn network_connect(options: &MqttOptions) -> Result<Network, ConnectionErr
                 .headers_mut()
                 .insert("Sec-WebSocket-Protocol", "mqtt".parse().unwrap());
 
+            if let Some(request_modifier) = options.request_modifier() {
+                request = request_modifier(request).await;
+            }
+
             let (socket, _) = async_tungstenite::tokio::client_async(request, tcp_stream).await?;
 
             Network::new(WsStream::new(socket), max_incoming_pkt_size)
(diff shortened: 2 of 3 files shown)


henil:
@reitermarkus can you check CI failures? 

Also I am not sure if the updated `Cargo.lock` file is required here, remove it if its not relevant with this changes.

reitermarkus:
Added `#[cfg(feature = "websocket")]` for the added parts and removed the `Cargo.lock` changes.

reitermarkus:
Added a `RequestModifierFn` type alias to fix the complex type warning.

reitermarkus:
Well, I only need to set the auth/signature headers, so I don't want to also have to set all MQTT protocol-specific headers.

Also, the AWS `ProvideCredentials` trait may be implemented in a such way that credentials are automatically refreshed, so no additional changes would be necessary when reconnecting.

reitermarkus:
That seems very unergonomic and unintuitive to use. I'd have to create an empty request with the URL I already passed to `MqttOptions::new`.

reitermarkus:
To get the headers I would still have to create a request, since signing is done on a fully formed request. There is no way to only get signature headers without a request in the first place.

yatinmaan:
> To get the headers I would still have to create a request, since signing is done on a fully formed request. There is no way to only get signature headers without a request in the first place.

@reitermarkus Makes sense. That's the sort of stuff I was look for when I asked.

>> Unless there's a use case I missed which can't be serviced like this?

Sorry for the delay and thanks for your contribution :rocket:

reitermarkus:
@yatinmaan, could you release a new version with this included? Thanks.

yatinmaan:
> @yatinmaan, could you release a new version with this included? Thanks.

@swanandx ^^

swanandx:
thnx for the ping @yatinmaan ! We will release rumqttc this week @reitermarkus  :100:

swanandx:
update: rumqttc v0.23.0 is released, which includes this change!

PR #646, opened 2023-06-21
feat: Synchronous Client: Remove mut from methods

BKDaugherty:
I realized that the synchronous client might not actually need to take `&mut` self instead of `&self` here. I think reducing the requirements on this would make it more ergonomic. 

Also with respect to a breaking change, I'm not actually sure that this is one. I don't think it is, but am not really sure how to tell.

## Type of change

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Expose `EventLoop::clean` to allow triggering shutdown and subsequent storage of pending requests
 
 ### Changed
+- Synchronous client methods take `&self` instead of `&mut self` (#646)
 
 ### Deprecated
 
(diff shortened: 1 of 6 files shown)


rumqttc/CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `MqttState::new` takes `max_outgoing_packet_size` which was set in `MqttOptions` but not used (#622)
+- Synchronous client methods take `&self` instead of `&mut self` (#646)
```
swanandx, line 38: move this to unreleased changes section

swanandx (changes requested):
please sync up the with latest main branch to resolve the merge conflicts, rest LGTM! Thank you for the PR :smile:

BKDaugherty:
Thanks for the review @swanandx! Rebased on main, and updated the changelog! Would love another look!

swanandx (approved):

PR #648, opened 2023-06-22
feat: remove unnecessary features to avoid conflict

swanandx:
Executing `cargo test --all-features` from `rumqtt` directory was causing some issue in `rumqttc` like:

```
error[E0308]: mismatched types
   --> rumqttc/src/eventloop.rs:405:22
    |
405 |                 Some(connector),
    |                 ---- ^^^^^^^^^ expected `tokio_native_tls::TlsConnector`, found `tokio_rustls::TlsConnector`
    |                 |
    |                 arguments to this enum variant are incorrect
    |
    = note: `tokio_rustls::TlsConnector` and `tokio_native_tls::TlsConnector` have similar names, but are actually distinct types
```

NOTE: Testing `rumqttc` and `rumqttd` separately doesn't cause this issue.

This issue came up when we enabled some features of `async_tungestine`, if we disable them ( cuz we _most likely_ don't need them ), this issue is resoved!

## Type of change

## Checklist:

- [ ] Make an entry to `CHANGELOG.md` if its relevant of user of the library. If its not relevant mention why. </br> - Just a fix


henil (commented):
i am not sure if this fixes the root of the issue. i think its problem is related to us only enabling on [`tokio-rustls-native-certs` in rumqttc](https://github.com/bytebeamio/rumqtt/blob/056a5e5df31a1bbe4f652048a2ef72e027560fea/rumqttc/Cargo.toml#L38). 

And when we run it without specifying a package, cargo build `async-tungstenite` [as union of both repos requirement](https://doc.rust-lang.org/cargo/reference/features.html?highlight=union#feature-unification).

Nickztar:
This will not prevent something like this from happening in the future and might be worth changing the tests/build to be run in individual packages. As it turns out we don't actually need these features so might be worth to remove them anyways.

henil:
from comment in discord channel, looks like this features are not required because we already handle tls certificates before handling websockets, so merging.

PR #649, opened 2023-06-23
The server should not send DISCONNECT packets to MQTT v4 clients

belltoy:
The server should not send DISCONNECT packets to MQTT v4 clients.

The MQTT v4 (3.1.1) specification only mentions that DISCONNECT packages are sent from the client to the server (see [section 3.14](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718090) and [section 2.1.1](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc353481061)).

> | Name | Value | Direction of flow | Description |
> |------|-------|-------------------|-------------|
> | ...  | ...   | ...               | ...         |
> | DISCONNECT | 14 | **Client to Server** | Client is disconnecting |

## Type of change

## Checklist:

### DIFF
--- a/rumqttd/CHANGELOG.md
+++ b/rumqttd/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Include reason code for UnsubAck in v5
+- The server should not send DISCONNECT packets to MQTT v4 clients
 
 ### Security
 
--- a/rumqttd/src/protocol/v4/disconnect.rs
+++ b/rumqttd/src/protocol/v4/disconnect.rs
@@ -2,7 +2,8 @@ use super::*;
 use bytes::{BufMut, BytesMut};
 
 // In v4 there are no Reason Code and properties
-pub fn write(_disconnect: &Disconnect, payload: &mut BytesMut) -> Result<usize, Error> {
-    payload.put_slice(&[0xE0, 0x00]);
-    Ok(2)
+pub fn write(_disconnect: &Disconnect, _payload: &mut BytesMut) -> Result<usize, Error> {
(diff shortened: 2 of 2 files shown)


henil:
Thanks for taking time to open the PR!

But the part that you mention is handled internally, broker makes sure that disconnect packet are not sent to client.

We have that there because `protocol` independent of broker, it can be used by anyone. Think of it as an independent mqtt packet serialization/deserialization crate. If its used by MQTTv4 client it would need this functionality.

henil:
please feel free to reopen if you have any further questions.

belltoy:
Sorry I missed that the router won't push an outgoing disconnect Notification if the `reason` is `None`.

> We have that there because protocol independent of broker, it can be used by anyone. Think of it as an independent mqtt packet serialization/deserialization crate. If its used by MQTTv4 client it would need this functionality.

BTW, do you have a plan to seperate the `procotol` or the MQTT serialization/deserialization out of both rumqttd and rumqttc as an independent crate? If so, the `mqttbytes` in rumqttc will be replaced by it. Right?

henil:
Yes we want to do that. Also there is already https://crates.io/crates/mqttbytes but it has not been maintained for sometime. 

There are some differences in `mqttbytes` and `protocol` because broker expects packets in certain format as a result there is alot of duplicated code between them.

We there exploring having all ser/deserialization parts in a common crate and provide a translation layer for either broker or client so it can get packets in the format it expects. and in future we might refactor them so they doesn't require translation. But we haven't gotten around to doing it yet.

If you need this and would like to contribute, that would be great!

CC: @yatinmaan

belltoy:
Thank you for replying.

I'm interested in these crates. I'll look deep into them to find what I can contribute.

henil:
you can join our discord if you have any questions, link is in README.

Issue #650, opened 2023-06-23
${topic} subscriptions are invalid and a session is disconnected


gechelberger:
# Expected Behavior

A subscription should be registered for any messages sent to a topic with a $ prefix.

Section 4.7.2 of both v3.1.1 and v5: 

```
The Server MUST NOT match Topic Filters starting with a wildcard character (# or +) with Topic Names 
beginning with a $ character [MQTT-4.7.2-1]. The Server SHOULD prevent Clients from using such 
Topic Names to exchange messages with other Clients. Server implementations MAY use Topic Names 
that start with a leading $ character for other purposes. 
```

It seems like it may be ok to block messages published to such a topic but subscriptions should be allowed.

# Current Behavior

The subscription is refused and the connection terminated.

## Failure Logs

```
mqttplay-broker-1  |   2023-06-23T22:39:22.186826Z  WARN rumqttd::router::routing: Subscription cannot be validated: Invalid filter prefix $SYS/#, reason: InvalidFilterPrefix("$SYS/#")
mqttplay-broker-1  |     in rumqttd::router::routing::incoming_payload with client_id: "mqtt-explorer-ad980b3f"
mqttplay-broker-1  |     in rumqttd::router::routing::[>] incoming with connection_id: 1
mqttplay-broker-1  | 
mqttplay-broker-1  |   2023-06-23T22:39:22.186908Z ERROR rumqttd::server::broker: router-drop, error: Recv(Disconnected)
mqttplay-broker-1  |     in rumqttd::server::broker::remote_link with tenant_id: None, client_id: "mqtt-explorer-ad980b3f", connection_id: 1
```

swanandx:
Hey, One use of $ is to have [shared subscription](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901250), like `$share/group/topic` in v5. This will be added by #628 PR.

Other than that, I can't think of a reason why client should be able to _subscribe_ to $topic.

[Docs](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901246) also say:
```
Applications cannot use a topic with a leading $ character for their own purposes
```

Currently, `$` isn't used for any special purpose by the Server, so I don't think it matters? Or do you want to say, let Client register the subscription, disconnect only those who try to publish to such topic names?

swanandx:
> The Server SHOULD prevent Clients from using such 
[...]

PR #651, opened 2023-06-25
Feature/filter and topic validation

gechelberger:
## Type of change

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -117,6 +117,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
 
 [[package]]
+name = "assert_matches"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
+
+[[package]]
 name = "async-http-proxy"
 version = "1.2.5"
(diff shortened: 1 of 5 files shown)


gechelberger:
fixes #650 

Looking at the diff, I'm noticing that the RouterError is not marked as non_exhaustive which means this isn't technically a non breaking change :/

Thoughts?

swanandx:
Breaking changes are those changes which affect/change the behavior, in this case, think of it like previously we would disconnect/disallow subscriptions to topic with `$` prefix, but with this PR, we allow them, which is change in behavior. So this should be a breaking change right? 

Please correct me if I got the wrong idea haha :sweat_smile:

gechelberger:
I generally think of breaking changes as having 2 criteria:

1) Is the change of behavior going to effect the behavior of existing applications?

In this case the behavior is more permissive or will reject requests which do not conform to spec. So from a behavioral perspective I do not consider this a breaking change. Only applications which has malformed requests should see any differences.

2) Is the change going to effect compilation of existing codebases for which this is a dependency?

In this case, since the RouterError is not marked as #[non_exhaustive], any application which matches on a RouterError but does not provide a catch-all match arm ```_ => { return foo(); } ``` will fail to compile since we have added a new RouterError::InvalidTopicFilter to the enum that must be handled. So, in that sense this is definitely a breaking change.

In general, I think it would be useful to look at any enums in the library's public interface that we expect could require additional cases and mark them as non-exhaustive to avoid this situation in the future.

jarvjani:
Hi,
This would be great so we could use MQTT explorer. 

What is the status of this? looks like it is already implemented but this stays in draft?

jarvjani:
Anyway, here's workaround for developers using MQTT exporer

image

PR #653, opened 2023-06-29
Disable unsused features from `flume` dependency

domenicquirl:
This PR modifies the dependency on `flume` to only enable its `async` feature instead of all features that are enabled by default. The motivation for this change is that neither `rumqttc` nor `rumqttd` need to select between (i.e., wait for any one of multiple) messages from multiple channels. However, the `select` feature of `flume` is enabled by default, together with its `eventual-fairness` feature that ensures that channels in a selector cannot make other ones starve. To do so, a source of randomness is required which is obtained through additional dependencies that this PR aims to avoid for both users and developers of `rumqtt`.

## Type of change

Miscellaneous (related to maintanance) 

## Checklist:

- [x] Formatted with `cargo fmt` (only `Cargo.toml`s are changed)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -664,7 +664,6 @@ checksum = "1657b4441c3403d9f7b3409e47575237dac27b1b5726df654a6ecbf92f0f7577"
 dependencies = [
  "futures-core",
  "futures-sink",
- "nanorand",
  "pin-project",
  "spin 0.9.8",
 ]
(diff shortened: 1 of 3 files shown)


domenicquirl:
Is a `CHANGELOG` entry required for this change? The change should be transparent for any users, except if they were unknowingly relying on `rumqtt` enabling the features while also depending on `flume` without specifying them. 

Based on the previous entries it looks to me like this wouldn't appear in the list, but I can include a mention if you prefer.

henil:
no changelog entry not required for this

henil:
Thank you @domenicquirl!

PR #654, opened 2023-07-03
rumqttc: Add bind_device to NetworkOptions to enable TCPSocket.bind_device()

seimonw:
Allows the network connection to be optionally bound to a specific interface.
See also: https://docs.rs/tokio/latest/tokio/net/struct.TcpSocket.html#method.bind_device

## Type of change

## Checklist:

### DIFF
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -131,7 +131,7 @@ impl EventLoop {
         if self.network.is_none() {
             let (network, connack) = match time::timeout(
                 Duration::from_secs(self.network_options.connection_timeout()),
-                connect(&self.mqtt_options, self.network_options),
+                connect(&self.mqtt_options, self.network_options.clone()),
             )
             .await
             {
@@ -242,7 +242,7 @@ impl EventLoop {
     }
 
     pub fn network_options(&self) -> NetworkOptions {
-        self.network_options
+        self.network_options.clone()
     }
 
     pub fn set_network_options(&mut self, network_options: NetworkOptions) -> &mut Self {
@@ -311,6 +311,16 @@ pub(crate) async fn socket_connect(
         if let Some(recv_buffer_size) = network_options.tcp_recv_buffer_size {
             socket.set_recv_buffer_size(recv_buffer_size).unwrap();
         }
+
+        #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
+        socket
+            .bind_device(
+                network_options
+                    .bind_device
+                    .as_ref()
+                    .map(|bind_device| bind_device.as_bytes()),
+            )
+            .unwrap();
 
         match socket.connect(addr).await {
             Ok(s) => return Ok(s),
--- a/rumqttc/src/lib.rs
+++ b/rumqttc/src/lib.rs
@@ -378,11 +378,13 @@ impl From<ClientConfig> for TlsConfiguration {
 }
 
 /// Provides a way to configure low level network connection configurations
-#[derive(Clone, Copy, Default)]
+#[derive(Clone, Default)]
 pub struct NetworkOptions {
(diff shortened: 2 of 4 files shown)


rumqttc/src/eventloop.rs
```diff
@@ -312,6 +312,15 @@ pub(crate) async fn socket_connect(
...
         }
 
+        socket
+            .bind_device(
+                network_options
+                    .bind_device
+                    .as_ref()
+                    .map(|bind_device| bind_device.as_bytes()),
+            )
+            .unwrap();
```
henil, line 322: https://docs.rs/tokio/latest/tokio/net/struct.TcpSocket.html#method.bind_device this method is available for specific platforms only, so we should only include this if we are building for that target.

henil (commented):
else LGTM, but can you explain the use case for this?

seimonw:
> else LGTM, but can you explain the use case for this?

It can be used to force the connection to go via a specific network which may not be the default route. I would like to use it to maintain two independent connections to a server. I can verify the backup network route's connectivity while the primary (default) route is still active.

rumqttc/src/eventloop.rs
seimonw: Also added docs to the set function

rumqttc/src/eventloop.rs
```diff
@@ -312,6 +312,16 @@ pub(crate) async fn socket_connect(
             socket.set_recv_buffer_size(recv_buffer_size).unwrap();
         }
 
+        #[cfg(all(any(target_os = "android", target_os = "fuchsia", target_os = "linux")))]
```
henil, line 315:
```suggestion
        #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
```
is `all` required here?

rumqttc/src/lib.rs
```diff
@@ -412,6 +414,13 @@ impl NetworkOptions {
     pub fn connection_timeout(&self) -> u64 {
         self.conn_timeout
     }
+
+    /// bind connection to a specific network device by name
+    /// only supported on linux, android and fuchsia
+    pub fn set_bind_device(&mut self, bind_device: &str) -> &mut Self {
+        self.bind_device = Some(bind_device.to_string());
+        self
+    }
```
henil, line 423: would it make sense to set this value if `target_os` is not one of the supported OS?
 
i would prefer to add `cfg` in this method as well as in `bind_device` member in the struct

rumqttc/src/eventloop.rs
henil: fair enough

rumqttc/src/lib.rs
seimonw: Yeah, I considered that also, but decided that keeping it in it would make any examples that use the function more portable albeit non-functional.
Feels messy putting `#[cfg(all(any(target_os = "android", target_os = "fuchsia", target_os = "linux")))]` all over the place. What do you think about if I add a `bind-device` feature, that `compile_error!`s if it's not the supported os like: https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features ?

seimonw: Yeah, and also at `NetworkOptions::new()`. I'll add that in.

seimonw:
What's the latest with this PR @henil - it seems to have stalled. Is there anything more I can do?

henil:
Hey, sorry for the delay. we have backlog of things to review and merge but we dont have enough bandwidth. I will try to get this reviewed and merged by this month.

seimonw:
No worries - I noticed that clippy didn't like the `#cfg(all(any` from tokio so have pushed without the all in there

henil:
Thank you @seimonw :tada:

henil:
Thank you @seimonw! 🎉

Issue #643, opened 2023-06-15
Keep Alive has an arbitrary limit


danieldougherty:
# Expected Behavior

According to both MQTT specs, the Keep Alive value should be an integer from 0 to 65,535. A non-zero value indicates the number of seconds between messages before the client must send a pingreq; a value of 0 disables the keep-alive mechanism.

# Current Behavior

Attempting to set a keep-alive value of less than 5 seconds fails with the output 'Keep alives should be >= 5 secs'. This makes it impossible to set a 0 value, or to use shorter intervals for testing/time-sensitive applications.

henil:
currently [pings are sent](https://github.com/bytebeamio/rumqtt/blob/714f71c9bc3369fa186353201dc2e8197d07594c/rumqttc/src/eventloop.rs#L228-L240) regardless of if there is activity on the network or not. 

So to avoid user from setting it something very small and generate alot of unnecessary traffic we had put that limit. Also according to spec:
> The actual value of the Keep Alive is application specific; typically this is a few minutes.

But I agree it should be `>= 1 second`.


> it impossible to set a 0 value

i think this can be implemented by disabling the `select!` branch that sends the pings (cc: @yatinmaan)

Both of this should be an easy fix, do you want to open a PR for this @danieldougherty ?

PR #655, opened 2023-07-03
fix(Event Loop): Allow keep-alives <= 5 secs

danieldougherty:
This PR removes the requirement that the keep alive timeout be at
least 5 seconds. This is in keeping with the MQTT spec, which
states that the keep alive value should be between 0 and 65535.
A value of 0 indicates that no keep-alive pings are sent to the
broker.

Issue: #643 - Keep Alive has an arbitrary limit

## Type of change

## Checklist:

- [ X ] Formatted with `cargo fmt`
- [ X ] Make an entry to `CHANGELOG.md` if its relevant of user of the library. If its not relevant mention why.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ rumqttc v0.19.0
 - Fix examples to stop printing error in loop (#540)
 - MQTTv5!: Remove `Connect` from `ConnectionError::StateError` (#541)
 - MQTTv5: Send last_will and login info with connect (#478)
+- Allow keep alive values <= 5 seconds (#643)
 
 rumqttd v0.12.1
 -------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ rumqttc v0.19.0
 - Fix examples to stop printing error in loop (#540)
 - MQTTv5!: Remove `Connect` from `ConnectionError::StateError` (#541)
 - MQTTv5: Send last_will and login info with connect (#478)
-- Allow keep alive values <= 5 seconds (#643)
 
 rumqttd v0.12.1
 -------
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -140,7 +140,7 @@ impl EventLoop {
             };
             self.network = Some(network);
 
-            if self.keepalive_timeout.is_none() {
+            if self.keepalive_timeout.is_none() && !self.mqtt_options.keep_alive.is_zero() {
                 self.keepalive_timeout = Some(Box::pin(time::sleep(self.mqtt_options.keep_alive)));
             }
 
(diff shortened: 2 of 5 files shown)


rumqttc/src/eventloop.rs
```diff
@@ -227,7 +228,8 @@ impl EventLoop {
             },
             // We generate pings irrespective of network activity. This keeps the ping logic
             // simple. We can change this behavior in future if necessary (to prevent extra pings)
-            _ = self.keepalive_timeout.as_mut().unwrap() => {
+            _ = self.keepalive_timeout.as_mut().unwrap_or(&mut no_sleep),
+                if self.keepalive_timeout.is_some() && !self.mqtt_options.keep_alive.is_zero() => {
```
henil, line 232: i dont think `unwrap_or` is required here because of `self.keepalive_timeout.is_some()` guard but keeping that is also fine.

rumqttc/src/lib.rs
```diff
@@ -562,8 +562,6 @@ impl MqttOptions {
     /// Set number of seconds after which client should ping the broker
     /// if there is no other data exchange
     pub fn set_keep_alive(&mut self, duration: Duration) -> &mut Self {
-        assert!(duration.as_secs() >= 5, "Keep alives should be >= 5 secs");
-
         self.keep_alive = duration;
         self
     }
```
henil, line 567: according to mqtt spec:
> The Keep Alive is a time interval measured in seconds. Expressed as a 16-bit word, it is the maximum time interval that is permitted to elapse between the point at which the Client finishes transmitting one Control Packet and the point it starts sending the next. It is the responsibility of the Client to ensure that the interval between Control Packets being sent does not exceed the Keep Alive value. In the absence of sending any other Control Packets, the Client MUST send a PINGREQ Packet [MQTT-3.1.2-23].

keep alive should be in seconds but because we take `Duration` as a argument, which means it can technically contain nanoseconds, so ideally we should take `usize` as argument.

But to not break API for now we can verify that duration doesn't contain nanoseconds inside this method.

wdyt? @yatinmaan

henil (commented):
few comments

rumqttc/src/eventloop.rs
danieldougherty: I just double-checked this to be sure and it is required: The assignment expression is evaluated before the guard, so even though we check for `is_some()` and then skip the handling code that follows, it's not until after the system has called `unwrap()`. By making this `unwrap_or()`, we remove the possibility of a panic due to calling `unwrap()` on `None`.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ rumqttc v0.19.0
 - Fix examples to stop printing error in loop (#540)
 - MQTTv5!: Remove `Connect` from `ConnectionError::StateError` (#541)
 - MQTTv5: Send last_will and login info with connect (#478)
+- Allow keep alive values <= 5 seconds (#643)
```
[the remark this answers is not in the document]
danieldougherty: Done.

danieldougherty:
Friendly ping. Is this still a viable change?

henil:
Hey, sorry for the delay. we have backlog of things to review and merge but we dont have enough bandwidth. I will try to get this merged by this month.

henil:
thank you @danieldougherty! :tada:

PR #659, opened 2023-07-11
Builder pattern for constructing Link

swanandx:
Support for constructing `LinkRx`/`LinkTx` using `LinkBuidler`.

```rust
let (link_tx, link_rx, notification) = LinkBuilder::new(&client_id, router_tx)
    .tenant_id(tenant_id) // None by default
    .clean_session(clean_session) // true by default
    .last_will(lastwill) // None by default
    .dynamic_filters(dynamic_filters) // false by default
    .topic_alias_max(topic_alias_max) // 0 by default
    .build()?;
```
## Type of change

## Checklist:

### DIFF
--- a/rumqttd/src/link/local.rs
+++ b/rumqttd/src/link/local.rs
@@ -129,6 +129,7 @@ impl<'a> LinkBuilder<'a> {
 
 pub struct Link;
 
+#[deprecated = "Link will be removed soon, please consider using LinkBuilder"]
 impl Link {
     #[allow(clippy::type_complexity)]
     fn prepare(
--- a/rumqttd/src/link/local.rs
+++ b/rumqttd/src/link/local.rs
@@ -35,6 +35,98 @@ pub enum LinkError {
     Elapsed(#[from] tokio::time::error::Elapsed),
 }
 
+// used to build LinkTx and LinkRx
+pub struct LinkBuilder<'a> {
+    tenant_id: Option<String>,
+    client_id: &'a str,
+    router_tx: Sender<(ConnectionId, Event)>,
+    // true by default
+    clean_session: bool,
+    last_will: Option<LastWill>,
+    // false by default
+    dynamic_filters: bool,
+    // default to 0, indicating to not use topic alias
+    topic_alias_max: u16,
+}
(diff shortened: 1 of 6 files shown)


rumqttd/src/link/local.rs
```diff
@@ -35,6 +35,98 @@ pub enum LinkError {
     Elapsed(#[from] tokio::time::error::Elapsed),
 }
 
+// used to build LinkTx and LinkRx
+pub struct LinkBuilder<'a> {
+    tenant_id: Option<String>,
+    client_id: &'a str,
```
mnpw, line 41: we clone this while creating connection anyway so let's just make it `String`

rumqttd/src/link/local.rs
```diff
@@ -35,6 +35,98 @@ pub enum LinkError {
...
+            _message => return Err(LinkError::NotConnectionAck),
+        };
+
+        let tx = LinkTx::new(id, self.router_tx.clone(), incoming_data_buffer);
+        let rx = LinkRx::new(id, self.router_tx, link_rx, outgoing_data_buffer);
+        Ok((tx, rx, notification))
+    }
+}
+
 pub struct Link;
```
mnpw, line 130: what's the strategy on deprecating `Link::(prepare, new)` methods?

mnpw (commented):
LGMT, thanks 🚀🚀🚀

please merge after addressing the comments

rumqttd/src/link/local.rs
swanandx: in upcoming release, we will mark it as [deprecated](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute), and after that, we will just remove it.

swanandx: I am not sure if they are used by someone or not, that is why we are planning this deprecation strategy, otherwise we can just remove them and mention it in changelog.

rumqttd/src/link/local.rs
swanandx: keeping it unchanged! resolving the conversation.

PR #668, opened 2023-07-30
Shared subscriptions with configurable strategy

Nickztar:
This PR adds support for shared subscriptions. This PR adds support for shared subscriptions in broker with the ability to use multiple strategies for it. The available strategies are as follows:

- round_robin (Default): 	In order of subscription, moving from first to last to first etc.
- random: 	Random selection among all subscribers, using `rand` crate to select a random client.
- sticky: 	Always send to the last selected subscriber, only moving when a client is no longer available.

Strategies are specified in the router config and by default commented but easily changeable.

This PR supersedes #628 and was made by me and @swanandx. 👍🏼 

## Type of change

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1968,6 +1968,7 @@ dependencies = [
  "parking_lot 0.11.2",
  "pretty_assertions",
  "pretty_env_logger",
+ "rand",
  "rustls-pemfile",
  "rustls-webpki",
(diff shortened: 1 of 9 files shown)


swanandx (approved):
LGTM! great work @Nickztar & @swanandx :rocket:

PR #671, opened 2023-08-03
Maintainability Improvements

carlocorradini:
This PR tries to enhance the overall maintainability of `rumqtt`.
Note that this PR is not ready to be merged because the code style (`clippy`) must be applied to the code.
What has been changed:
- `Stale` (bot) is a GitHub action
- Better CI/CD with enhanced checks and automation
- `VSCode` compatibility
- All files must respect a format/coding standard for better consistency (e.g. `.rs` has `fmt/clippy`, `.sh` has `prettier/shellcheck`, etc...)
- Added `EditorConfig`
- Markdown linter via `markdownlint`
- `Prettier` file formatter
- Code spell checker via `cspell`. A lot of typos have been found and fixed
- `rumqttd` Docker image is now based on `gcr.io/distroless/cc` (from Google) to enhance overall security since _"it does not contain package managers, shells or any other programs you would expect to find in a standard Linux distribution_"
- Multi-manifest Docker image using `Docker buildx`. The new Docker image is `arm64` (aarch64) and `amd64` (x86_64)
- More robust and fully configurable scripts (see `scripts` directory). Also `POSIX` compliant
- `rumqttverifier` sources are now based on Typescript for better maintainability and more. Simply run `npx ts-node path/to/script.ts` instead of old `node path/to/script.js`
- Cross compilation:
    - `aarch64-unknown-linux-gnu` -> `rumqttd-1.0.0-aarch64-unknown-linux-gnu`
    - `aarch64-unknown-linux-musl` -> `rumqttd-1.0.0-aarch64-unknown-linux-musl`
    - `x86_64-unknown-linux-gnu` -> `rumqttd-1.0.0-x86_64-unknown-linux-gnu`
    - `x86_64-unknown-linux-musl` -> `rumqttd-1.0.0-x86_64-unknown-linux-musl`
    I need to test `Cross` a little bit more but it seems to work without any problem. We could add Windows!
- Updated release profile
    ```toml
    [profile.release]
    codegen-units = 1 # Better optimization
    lto = true # Link time optimization
    strip = true # Stripping
    ```
- Minimum supported `Rust` version. I've set it to `1.64.0` due to `Cross` requirement but it can be changed to better suit your needs
- `Rust` edition is set to `2021`
- Much more...

*Please, let me know what you think and what you want to change/update/add/delete* 🥳

PS: Thanks for developing `rumqtt*`

PR #673, opened 2023-08-07
chore: docker image based on distroless & build and docker scripts

carlocorradini:
Related to PR #671 

What has been changed:

- `rumqttd` Docker image is based on distroless rather than `Alpine`
   Docker image is now based on `gcr.io/distroless/cc` (from Google) to enhance overall security since _"it does not contain package managers, shells or any other programs you would expect to find in a standard Linux distribution"_

- More robust and fully configurable scripts (see `scripts` and `rumqttd/scripts` directories). Also `POSIX` compliant
    1. `rumqttd/scripts/docker.sh`
        Build and push `rumqttd` Docker image
    2. `rumqttd/scripts/build.sh`
        Build `rumqttd` using `Cross`

   Note that both scripts are configurable and also have an helper flag (`--help`)


carlocorradini:
@de-sh Any update? 🤗🥳

carlocorradini:
@swanandx @henil  Any thoughts on this? I apologize for pinging

henil:
i think the script is very complicated and i am not sure that its really necessary. and for distroless docker image what are the advantage to use those instead of what we have currently?

carlocorradini:
I believe the numerous papers on Distroless vs Alpine can explain it far better than I can.😂😅
See [this](https://www.sumup.com/careers/techstories/stop-using-alpine-docker-images) or [this](https://www.reddit.com/r/devops/comments/dtd2mu/distroless_docker_images_thoughts) or ...

carlocorradini:
These scripts are quite user-friendly (with `--help` and args configuration) and `POSIX` compliant

carlocorradini:
The `main` parts are a few lines of code:

`build.sh`
image

`docker.sh`
image

Easy peasy

PR #678, opened 2023-08-07
chore: added gitattributes and editorconfig & better gitignore

carlocorradini:
Related to PR #671 

What has been changed:

- Added `.gitattributes`
- Added `.editorconfig`
- Modified (better) `.gitignore`


.editorconfig
```diff
@@ -0,0 +1,14 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+max_line_length = 120
```
henil, line 10: i think this configs could conflict with rustfmt, so editior might format with something else during autosave and rustfmt might do something else. in general i think rustfmt does a good job of handling this so i don't think this is necessary.

carlocorradini: This is not for Rust only, but a general way for configuring editors.

We can add an entry to match rust files to match rustfmt but I think it's only one line (max lines?)

henil: > This is not for Rust only, but a general way for configuring editors.

yes i know but i don't want multiple ways to configure same thing.

is there any way to apply this settings for anything other than `*.rs` files?

carlocorradini: Yup, use `unset`. I think the only option to disable is `max_line_length`. The others must be the same

henil: problem is not that they are same, its just bad practice to have two conflicting configurations. when we want to update something in rustfmt we would need to update here as well. i would suggest we remove `.editorconfig` completely because code is mostly in rust and only some scripts in js

.gitattributes
henil, line 1: i tried reading git docs but didn't understand what this is used for, can you give tldr?

.editorconfig
carlocorradini: If you want I can unset everything for Rust only. A `.editorconfig` file is rarely (very rarely) touched again in a project so we are pretty safe.
But removing it completely is not a good practice, especially if you don't want any "ide" related files such as .VSCode or .idea

.gitattributes
carlocorradini: I think an article (like (this)[https://dev.to/deadlybyte/please-add-gitattributes-to-your-git-repository-1jld]) explain mu h much better than me 😂

swanandx: btw, we don't need all these entries here, we should only include the filetypes which are in our codebase!

carlocorradini: We can use https://github.com/alexkaratarakis/gitattributes for Rust.
Moreover I think adding the following simple script to CI could be very helpful:
```sh
missing_attributes=$(git ls-files | git check-attr -a --stdin | grep "text: auto") if [[ "$missing_attributes" ]]; then echo ".gitattributes rule missing for the following files:"; echo "$missing_attributes"; else echo "All files have a corresponding rule in .gitattributes"; fi
```
.
My `.gitattributes` is a file that I've been using for years (copy-paste) :)

swanandx: > We can use https://github.com/alexkaratarakis/gitattributes for Rust.

whatever works best for our repo haha, we can also add some minor modifications.

> I think adding the following simple script to CI could be very helpful

Maybe, but _personally_, I don't feel the need. Having all this is nice, but might not be that valuable/impact-full, wdyt? cc: @henil

carlocorradini: This and the other PRs are for maintainability improvements. After adding these files they are rarely updated :/

swanandx: I was talking about the CI for gitattributes 😅 not about all the PRs haha

carlocorradini: Even my brain is on vacation 🍹🏖️😎

.editorconfig
henil: > If you want I can unset everything for Rust only.

yes please do

.gitattributes
henil: yeah i agree with @swanandx, not really necessary in CI. 

And we can use `.gitattributes` by merging both rust's and alexkaratarakis's repo

carlocorradini: Done

.editorconfig
carlocorradini: Done

carlocorradini:
I've checked that all files have an entry in `.gitattributes`:

```sh
#!/usr/bin/env sh

# Fail on error
set -o errexit
# Disable wildcard character expansion
set -o noglob
# Disable uninitialized variable usage
set -o nounset
# Disable error masking in pipe
# shellcheck disable=SC3040
if (set -o pipefail 2> /dev/null); then
    set -o pipefail
fi

missing_attributes=$(git ls-files | git check-attr -a --stdin | grep 'text: auto' || printf '\n')

if [ -n "$missing_attributes" ]; then
  printf '%s\n%s\n' '.gitattributes rule missing for the following files:' "$missing_attributes"
else
  printf '%s\n' 'All files have a corresponding rule in .gitattributes'
fi
```

.gitignore
swanandx, line 1: `.gitignore` still have lots on unnecessary entries!
- we should not ignore `Cargo.lock`
- Entries for Gatsby, Next, Nuxt, Vue etc. aren't needed.
- if we are ignore `.idea` as whole, i don't think we need to ignore files inside it individually as well.

with respect to current `.gitignore` file, which entries do you think are missing and cause trouble in future from perspective of new contributor? Only need the bare minimum, wdyt?

carlocorradini: You are right! Give me 2 mins

carlocorradini: Do you know what are these? A comment could be helpful for others 😅`
immagine_2023-08-17_133907248

carlocorradini: @swanandx Done

carlocorradini:
@swanandx I've add `.vscode` in `.gitignore`
Do you want to keep continue ignore it or not (only 2 simple json files will be added)?

.gitignore
henil: pem and der are used to store certificates and its keys

perf.data is probably generated by `perf` which is a cpu profiling tool

other i don't think are useful

swanandx: > Do you know what are these? A comment could be helpful for others

I believe this files are generated when we profile rumqttd. @henil / @tekjar can give better context than me here haha.

swanandx:
> @swanandx I've add `.vscode` in `.gitignore` Do you want to keep continue ignore it or not (only 2 simple json files will be added)?

I would go with keeping it ignored for now.

.gitignore
carlocorradini: Comments added

carlocorradini:
I think now is ready! 🥳🤗

henil:
Thank you @carlocorradini! :tada:

PR #680, opened 2023-08-08
chore: enhance cargo workspace and release build

carlocorradini:
Related to PR #671

What has been changed:

- Enhance `cargo workspace` utilization while avoiding repetitions
- Updated release profile
    ```toml
    [profile.release]
    codegen-units = 1 # Better optimization
    lto = true # Link time optimization
    strip = true # Stripping
    ```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,8 +1,22 @@
 [workspace]
-
 members = [
   "rumqttc",
   "rumqttd",
   "benchmarks",
   "benchmarks/simplerouter",
 ]
+
+[workspace.package]
+rust-version = "1.70.0"
+edition = "2021"
+repository = "https://github.com/bytebeamio/rumqtt"
+license = "Apache-2.0"
+authors = ["tekjar <raviteja@bytebeam.io>"]
+
+[profile.dev.build-override]
+debug = true
+
+[profile.release]
+codegen-units = 1
+lto = true
+strip = true
--- a/rumqttd/Cargo.toml
+++ b/rumqttd/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "rumqttd"
 version = "0.17.0"
-publish = false
+publish = true
 description = "rumqttd is a high performance MQTT broker written in Rust which is light weight and embeddable"
 keywords = ["mqtt", "broker", "iot", "kafka", "nats"]
 categories = ["network-programming"]
--- a/rumqttd/Cargo.toml
+++ b/rumqttd/Cargo.toml
@@ -1,13 +1,15 @@
 [package]
 name = "rumqttd"
 version = "0.16.0"
-license = "Apache-2.0"
+publish = false
 description = "rumqttd is a high performance MQTT broker written in Rust which is light weight and embeddable"
-authors = ["tekjar <raviteja@bytebeam.io>"]
-edition = "2021"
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -1,8 +1,22 @@
...
-
 members = [
   "rumqttc",
   "rumqttd",
   "benchmarks",
   "benchmarks/simplerouter",
 ]
+
+[workspace.package]
+rust-version = "1.70.0"
```
henil, line 10: according to docs[1] this is minimal supported rust version, and i don't think you need `1.70` to compile this project. lets just remove it?

1: https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

henil (commented):
one nitpick else LGTM

swanandx:
also does `codegen-units = 1` really help in our case? Setting it to 1 doesn't guarantee better optimization, therefore just asking haha

carlocorradini:
@swanandx From [official documentation](https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units):

> This flag controls how many code generation units the crate is split into. It takes an integer greater than 0.
>
> When a crate is split into multiple codegen units, LLVM is able to process them in parallel. Increasing parallelism ?may speed up compile times, but may also produce slower code. Setting this to 1 may improve the performance of generated code, but may be slower to compile.

Cargo.toml
carlocorradini: I've set Rust 1.70 because a dependency of `rumqttd` requires it to compile (you can try this using a Dockerfile and specifying a rust version < 1.70).
Since `rumqttc` is a library and does not depend on `rumqttd`, the only change that needs to be made is to override `rust-version` just for `rumqttc`. Do you know what the minimum version is?

swanandx: Can we use this to find out the msrv? https://github.com/foresterre/cargo-msrv

swanandx: and as per docs quoted by @henil :

> The Rust version must be a bare version number with two or three components; it cannot include semver operators or pre-release identifiers. Compiler pre-release identifiers such as -nightly will be ignored while checking the Rust version. 

just in case it's the correct version, we can't specify it like `1.70.0`, we have to use `1.70` ( according to their example )

carlocorradini: Yeah nice! 
Awesome 😎

henil: @swanandx it says there can be 
> with two or three components

so `1.70.0` also works, and the next release is `1.71.0`

henil (approved):
LGTM, thank you @carlocorradini :tada:

rumqttd/Cargo.toml
```diff
@@ -1,13 +1,15 @@
 [package]
 name = "rumqttd"
 version = "0.17.0"
-license = "Apache-2.0"
+publish = false
```
swanandx, line 4: why is this `publish = false` ? we do publish rumqttd

Cargo.toml
swanandx: btw, can we once verify it its the correct msrv using cargo-msrv ?

henil: yeah i verified with `cargo-msrv`

rumqttd/Cargo.toml
henil: good catch

PR #683, opened 2023-08-15
feat(rumqttd): retained messages

swanandx:
This PR aims to bring support for retained messages in rumqttd.

Previously, we were appending the retained messages to `datalog` everytime a subscription was made. But as every client shares same datalog, this lead to client receiving multiple messages!

To fix this, we don't append messages to datalog, instead of that, we use a flag `forward_retained_messages` to indicate we still haven't forwarded the messages for a particular subscription. Once we read retained messages, we set this flag to false, which assures we only forward retained messages once!
 
## Type of change

## Checklist:


rumqttd/src/router/routing.rs
```diff
@@ -946,6 +941,9 @@ impl Router {
                 .insert(filter_path.clone(), subscription_id);
         }
 
+        let forward_retained_msg =
+            (retain_forward_rule != &protocol::RetainForwardRule::Never) && group.is_none();
```
henil, line 945: why do we check `group.is_none()` here?

rumqttd/src/router/routing.rs
```diff
@@ -954,12 +952,25 @@ impl Router {
...
+                // set true for new subscriptions
+                forward_retained_msg,
                 group,
             };
 
             self.scheduler.track(id, request);
             self.scheduler.reschedule(id, ScheduleReason::NewFilter);
             debug_assert!(self.scheduler.check_tracker_duplicates(id).is_none())
+        } else if retain_forward_rule == &protocol::RetainForwardRule::OnEverySubscribe
+            && group.is_none()
```
henil, line 964:
```suggestion
       // If subscription was already present and `RetainForwardRule` is `OnEverySubscribe` set `forward_retained_msg` to true again
        } else if retain_forward_rule == &protocol::RetainForwardRule::OnEverySubscribe
            && group.is_none()
```

nit: its non trivial to know that `HashSet::insert` returns a bool signifying if element was already present can we make it more descriptive by adding a comment there as well

rumqttd/src/router/routing.rs
```diff
@@ -1193,11 +1204,12 @@ fn append_to_commitlog(
...
         datalog.remove_from_retained_publishes(topic.to_owned());
     } else if publish.retain {
-        error!("Unexpected: retain field was not unset");
         datalog.insert_to_retained_publishes(publish.clone(), properties.clone(), topic.to_owned());
+        return Ok(None);
     }
 
-    publish.retain = false;
+    // NOTE(swanandx): why we unset the retain field?
+    // publish.retain = false;
```
henil, line 1212:
```suggestion

```

rumqttd/src/router/routing.rs
```diff
@@ -1358,7 +1370,23 @@ fn forward_device_data(
         inflight_slots = 1;
     }
 
-    let (next, publishes) =
+    let mut publishes = Vec::new();
+
+    if request.forward_retained_msg {
+        // NOTE: limit the number of read messages
+        // and skip the messages previously read.
```
henil, line 1377: how do we skip previously read messages here?

henil (commented):
few questions and suggestions

rumqttd/src/router/routing.rs
swanandx: AFAIK, retained messages are only sent for normal subscriptions, not shared subscriptions

rumqttd/src/router/routing.rs
swanandx: as of now, we don't / can't.

rumqttd/src/router/routing.rs
swanandx: turns out, it was required. as we do need to append that message to commitlog for existing subscribers!

PR #688, opened 2023-08-22
feat(rumqttc): Replace multi line debug log with single line

flxo:
## Type of change

## Checklist:

Change the ping timing debug log from a multi line string to a single line with the use of the `Debug` impl of `Duration`. Multi line logs screw a log stream in regards of readability. This PR doesn't contain any API change and just affect debug logs - no need to add a noisy changelog entry.


tekjar:
Thanks @flxo  :)

Issue #697, opened 2023-08-27
rumqttc: unable to specify subscription options in v5


swanandx:
The current API of v5 client doesn't let us specify subscription options while subscribing.

* rumqttc version: "0.22.0"

### Context
MQTTv5 introduces [subscription options](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901169). This options like `preserve_retain`, `retain_forward_rule` are preset in `Filter` defined [here](https://github.com/bytebeamio/rumqtt/blob/705b1b6d32be69edf7b1df768f857a6442b83c03/rumqttc/src/v5/mqttbytes/v5/subscribe.rs#L102):
```rust
pub struct Filter {
    pub path: String,
    pub qos: QoS,
    pub nolocal: bool,
    pub preserve_retain: bool,
    pub retain_forward_rule: RetainForwardRule,
}
```
but currently we can't pass this options to `Filter::new(_)`, instead it uses default values:
```rust

impl Filter {
    pub fn new<T: Into<String>>(topic: T, qos: QoS) -> Self {
        Self {
            path: topic.into(),
            qos,
            ..Default::default()
        }
    }
    //...
}
```
we should be able to pass down those options while creating `Filter`.

### Possible solutions ideas / challenges

We can either just add more arguments to `new(_)` or use a builder pattern for creating `Filter`, whichever is better.

We also need to think upon how the users will specify them:
- Do we pass this subscription options as arguments to `subscribe(_)`? ( would make API very verbose unnecessary as most users won't use these ) 
- Do we introduce a separate function? ( the we will have way too many fns just for subscribing haha ).

<br />
What I would prefer is: Instead of passing `topic`, `qos` to `subscribe(_)`, user will build `Filter` using builder pattern and the pass it to `subscribe(filter: Filter)`. Something like:

```diff
- pub async fn subscribe<S: Into<String>>(&self, topic: S, qos: QoS) -> Result<(), ClientError> {
+ pub async fn subscribe(&self, filter: Filter) -> Result<(), ClientError> {
```

<br />

Please feel free to comment your ideas / suggestions :rocket:

ecarrara:
@swanandx I made a PR that:

* Create the `FilterBuilder`
* Refactor the subscribe fns to receive a instance of `Filter`

It looks like a big breaking change. I'm not sure if this is the best way to do that, but I liked how the API looks receiving a Filter instance built with the builder pattern:

```rust
[...]

PR #698, opened 2023-08-28
refactor: Allow to specify MQTTv5 subscription options (#697)

ecarrara:
This PR adds support to specify subscription options in MQTTv5 client while subscribing. See #697.

BREAKING: v5 client subscribe methods:

```diff
- pub async fn subscribe<S: Into<String>>(&self, topic: S, qos: QoS) -> Result<(), ClientError> {
+ pub async fn subscribe(&self, filter: Filter) -> Result<(), ClientError> {
```

## Type of change

## Checklist:


brianmay:
I think this PR has the same issue I mentioned in #697. It is not possible to call `retain_forward_rule` because `RetainForwardRule` is in a private mod, and it is not possible to construct a value outside of this crate.

PR #700, opened 2023-08-29
refactor(rumqttd): let Broker derive Clone and remove start mut self

crimx:
fixes #699

### DIFF
--- a/rumqttd/src/server/broker.rs
+++ b/rumqttd/src/server/broker.rs
@@ -59,6 +59,7 @@ pub enum Error {
     Remote(#[from] remote::Error),
 }
 
+#[derive(Clone)]
 pub struct Broker {
     config: Arc<Config>,
     router_tx: Sender<(ConnectionId, Event)>,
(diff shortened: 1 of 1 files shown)


swanandx:
Hey, I'm curious how does this PR solves the mentioned discussion? can you elaborate more on this :sweat_smile:

crimx:
With the current implementation, mut broker is consumed by `start` and there is no way to access the broker afterwards.

Since both `Broker.config` and `Broker.router_tx` are thread safe, we can just move a `broker.clone()` to a thread and `start`, move another clone to other thread and `link`.

henil:
while sharing it would be possible, it would also allow `start` to be called multiple times which shouldn't happen.

Can you share why do you want to share the broker to another thread?

crimx:
You are right this could be an issue.

Mu use case is that I want to link and subscribe to dynamic topics that are created after broker starts.

swanandx:
> I want to link and subscribe to dynamic topics that are created after broker starts.

You can create link using `LinkBuilder` with just client id and router_tx like [here](https://github.com/bytebeamio/rumqtt/blob/main/rumqttd/src/server/broker.rs#L150) .

So you might not need whole `Broker`, just `router_tx` should be enough!

crimx:
Yes that'll do. I've read this but the`router_tx` and `LinkBuilder` are private so I assumed you want to hide the details.

swanandx:
> I've read this but the`router_tx` and `LinkBuilder` are private
> 

Good catch, i somehow managed to miss this detail 😬😅

PR #704, opened 2023-09-02
use patched `ws_stream_tungstenite` to remove `webpki`

### LINKED ISSUE #689: Vulnerability in `webpki` 0.22.0
The vulnerability https://rustsec.org/advisories/RUSTSEC-2023-0052 has been issued yesterday and the crate is present in your dependency tree, making our CI to fail.

Updating your dependency on `async-tungstenite` from `0.22` to `0.23` should fix the issue.

Would you be so kind as to update it and release a new version please 🙏🏻 ?

Thanks in advance.

swanandx:
Fixes #689 .

Once https://github.com/najamelan/ws_stream_tungstenite/pull/11 is merged, and `ws_strem_tungstenite` is released with the patch, we will switch back to using it from crates.io ( most likely in `ws_stream_tungstenite` version `0.11.0` )

Until then, as this is security issue, we should not block on it and use the patched version.
## Type of change

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,7 +185,20 @@ dependencies = [
  "rustls-native-certs",
  "tokio",
  "tokio-rustls",
- "tungstenite",
+ "tungstenite 0.19.0",
+]
+
+[[package]]
+name = "async-tungstenite"
+version = "0.23.0"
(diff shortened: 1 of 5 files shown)


henil (approved):
if updating other dependencies is not requires, lets avoid that in this patch.

PR #708, opened 2023-09-05
feat: non-consuming builder for Connection

swanandx:
Construct Connection using builder pattern to avoid passing too many arguments to `Connection::new`

## Type of change

## Checklist:


henil (commented):
lgtm

PR #709, opened 2023-09-06
feat(rumqttd): assign random client identifier to clients

swanandx:
In case client tries to connect with client identifier of 0 length, i.e. empty client id, rumqttd assigns it a random client ID to use.

## Type of change

## Checklist:


rumqttd/src/server/broker.rs
```diff
@@ -478,6 +480,16 @@ async fn remote<P: Protocol>(
...
 
+    let mut assigned_client_id = None;
+    if client_id.is_empty() {
+        client_id = thread_rng()
+            .sample_iter(&Alphanumeric)
+            .take(10)
+            .map(char::from)
+            .collect();
+        assigned_client_id = Some(client_id.clone());
+    }
```
henil, line 491: this is completely random right, how about doing `mqtt-` prefix on this? how does other broker do it?

rumqttd/src/link/remote.rs
```diff
@@ -103,8 +104,13 @@ impl<P: Protocol> RemoteLink<P> {
...
 
-        if let Some(packet) = notification.into() {
-            network.write(packet).await?;
+        if let Some(mut packet) = notification.into() {
+            if let Packet::ConnAck(_ack, props) = &mut packet {
+                let mut new_props = props.clone().unwrap_or_default();
+                new_props.assigned_client_identifier = assigned_client_id;
+                *props = Some(new_props);
+                network.write(packet).await?;
+            }
```
henil, line 113: in v4, this props will be ignored right, so is there any way for client to know which client_id was it assigned?

rumqttd/src/protocol/v4/connack.rs
```diff
@@ -51,7 +51,7 @@ fn connect_code(return_code: ConnectReturnCode) -> u8 {
     match return_code {
         ConnectReturnCode::Success => 0,
         ConnectReturnCode::RefusedProtocolVersion => 1,
-        ConnectReturnCode::BadClientId => 2,
+        ConnectReturnCode::ClientIdentifierNotValid => 2,
```
henil, line 54: unrelated to this PR, but in v5 this connect reason codes are not valid, see:

https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901079
https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901079

can you open an issue for this

rumqttd/src/link/remote.rs
swanandx: in v4, nope.

rumqttd/src/protocol/v4/connack.rs
swanandx: wdym by they are wrong? also the quoted line only affects v4.

rumqttd/src/server/broker.rs
swanandx: don't know how other brokers do it, if we really want, we can add `rumqttd-` perfix

swanandx: I tested other broker:
- HiveMQ =>  `hmq_7IEiN_7079065_e26502d9e183e8866897f10aed5168d`
- Emqx => `MzEyNTU0NzgyMTc1MzI5NzU4ODg2MDQ0MDE2MDQ2MjQzODE`

shall we still append some prefix like `rumqttd_` ?

benjamin-nw: Hi,

To add more examples:
* `Mosquitto` =>  prefix is configurable in the configuration file and defaults to `auto-` for empty `cliend_id`.

Since the field is the `client id`, should we use `rumqttc-` as a prefix ?

And for the `client_id` generation, have you considered using `UUID`s ?

swanandx: > Since the field is the client_id, should we use rumqttc- as a prefix
>
Nope, cuz it is assigned by the broker, not client, client can be any.

> using UUID
>
Initially I did think of using it, but it felt like an overkill in this case, wdyt?

benjamin-nw: From the MQTTv5 [spec](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901059):
> A Server MAY allow a Client to supply a ClientID that has a length of zero bytes, however if it does so the Server MUST treat this as a special case and assign a **unique** ClientID to that Client

For me the `UUID` is a good choice because it has the property of uniqueness (well considering you don't have [trillions](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions) of clients) and also considering that it already does what you are doing by using `rng` for version 4.

With the crate `uuid`, you can also choose the [speed](https://docs.rs/uuid/1.5.0/src/uuid/rng.rs.html) for the `rng` generation with the `fast-rng` feature flag (i.e call `get_random()` by default, and call `rand::random()` for `fast-rng`).

If we want to keep the random generation of the `client_id`, we can choose `UUID` v4, that generates all bytes randomly and only sets the `version` and `variant` bits.

For me it seems simpler in the code to see a `UUID` and go: "oh so you generate a unique id to identify this client".

swanandx: > For me it seems simpler in the code to see a UUID and go: "oh so you generate a unique id to identify this client".
>
Agree 💯 it's like a trade-off between this and adding extra dependencies ( just for this small task )

imho, there _should_ not be conflicts with 10 random characters right? ( we can increase length if we will to be more sure )

swanandx: this maths went over my head ngl 😭😂 regarding the size, I thought it would be better to not have it so long ( reduces packet size ) and that is the reason I didn't even add a prefix. But the longer it is, the less chances of conflict right? Gotta find the sweet spot haha

also, what padding you want to mention?

benjamin-nw: yeah i agree that the size shouldn't be too high to reduce the packet size.

> But the longer it is, the less chances of conflict right? Gotta find the sweet spot haha
Yup, exactly. Well 10 seems good enough, `mosquitto` uses 16 bytes. I don't know about the others.

Forget about the padding, I've seen the spec and there is none. I was just afraid that some padding was put in the packet for some reasons, but it's not the case.

swanandx: updated to use 16 bytes long client id :+1: 

for prefix, as it isn't that important, sticking with not adding any prefix for now, in future we might add it if required. 

resolving this comment as done, please feel free to reopen / comment if you think otherwise, thanks :)

rumqttd/src/link/remote.rs
```diff
@@ -196,16 +202,18 @@ where
...
+        let ack = ConnAck {
+            session_present: false,
+            code: ConnectReturnCode::ClientIdentifierNotValid,
+        };
+
+        let packet = Packet::ConnAck(ack, None);
+        network.write(packet).await?;
+
         return Err(Error::InvalidClientId);
     }
```
henil, line 218: also lets keep the other logic same in the PR

swanandx: which logic?

henil: the config flag

swanandx: can you explain the old logic a bit, like why `allow-duplicate-client-id` need that check?

I can't figure out the use of that condition 😅

swanandx: I think the update logic is more or less the same:

previously in that cfg! flag:

if client_Id is empty and clean session is false, we return error ( that is what we do rn as well )

else:
return error if empty client id ( we don't return error as we are generating random client id! as per specs empty client id can't be specified with clean session false, hence we return error in that case. )

so I think this is good to go? wdyt?

swanandx: we can remove that feature completely btw, as this is the only place it is getting used :eyes:

henil: yeah make sense

henil (approved):

PR #711, opened 2023-09-07
Make request_modifier usable from async trait

frank-hivewatch:
This patch decorates the request_modifier with Send + Sync markers.
The request_modifier must be thread safe when called from async trait.

## Type of change

## Checklist:


Siddius:
can we merge this, because this is actual a breaking change from 0.22 to 0.23. Would be nice to get it fixed again.

swanandx:
Hey @frank-hivewatch , apologies for super late reply!

as of now, the changes done in this PR are incorporated in fn signature!: https://github.com/bytebeamio/rumqtt/blob/f7c77930ce11df0b7cffd61708de4b76b41dcb86/rumqttc/src/v5/mod.rs#L53 

so I will close the PR. Thank you so much for the contribution and sorry once again for the PR going unnoticed :sweat_smile: 

@Siddius this changes will be available in 0.24 ( expect it to be released next week ).

Thank you so much.

PR #713, opened 2023-09-08
Call the bind_device function only if the bind_device network option …

frank-hivewatch:
Call the bind_device function only if the bind_device network option is defined.

If binding device is None or an empty string it removes the binding,
which is causing PermissionDenied errors in AWS environment (lambda function).

## Type of change

## Checklist:


henil:
Thank you!

PR #719, opened 2023-09-27
feat: drop messages for which datalog doesn't exists

swanandx:
## Type of change

## Checklist:

### DIFF
--- a/rumqttd/src/router/routing.rs
+++ b/rumqttd/src/router/routing.rs
@@ -43,8 +43,6 @@ pub enum RouterError {
     #[cfg(feature = "validate-tenant-prefix")]
     #[error("Bad Tenant")]
     BadTenant(String, String),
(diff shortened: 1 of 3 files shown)


rumqttd/src/router/routing.rs
```diff
@@ -1229,17 +1227,17 @@ fn append_to_commitlog(
...
-        Some(v) => v,
-        None if connection.dynamic_filters => {
+    if filter_idxs.is_empty() {
+        // Create a dynamic filter if dynamic_filters are enabled for this connection
+        if connection.dynamic_filters {
             let (idx, _cursor) = datalog.next_native_offset(topic);
-            vec![idx]
+            filter_idxs = vec![idx];
+        } else {
+            warn!(topic, "no datalog found for topic, dropping message");
```
henil, line 1238: hmm with this we would be returning `(0, 0)` would that be valid?

PR #724, opened 2023-10-05
fix: validate subprotocol mqtt

### LINKED ISSUE #640: Rumqttc: Validate response header
# Expected Behavior

Rumqttc should validate that the broker that it connects to has the correct websocket-protocol header set. As per the [spec](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Using_WebSocket_as):
` The WebSocket Subprotocol name selected and returned by the Server MUST be “mqtt”`

# Current Behavior

Currently it ignores it and continues not matter the response.

Should be a very simple fix. [The discarded parameter here](https://github.com/bytebeamio/rumqtt/blob/714f71c9bc3369fa186353201dc2e8197d07594c/rumqttc/src/eventloop.rs#L389) includes the response headers, just needs to validate that it is there, otherwise return a error.

Vilayat-Ali:
Fixes #640 

Added validation to check for response headers for websocket subprotocol for "mqtt" and returns a new error `WebsocketSubprotocol` variant if absent.

## Type of change

## Checklist:

### DIFF
--- a/mosquitto.conf
+++ b/mosquitto.conf
@@ -1,0 +1,3 @@
+listener 9001
+protocol websockets
+allow_anonymous true
--- a/mosquitto.conf
+++ b/mosquitto.conf
@@ -1,3 +1,0 @@
-listener 9001
-protocol websockets
-allow_anonymous true
--- a/rumqttc/examples/websocket.rs
+++ b/rumqttc/examples/websocket.rs
@@ -11,11 +11,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     pretty_env_logger::init();
 
     // port parameter is ignored when scheme is websocket
-    let mut mqttoptions = MqttOptions::new(
-        "clientId-aSziq39Bp3",
-        "ws://broker.mqttdashboard.com:8000/mqtt",
-        8000,
-    );
+    let mut mqttoptions: MqttOptions = MqttOptions::new("clientId-1", "ws://localhost:9001", 9001);
     mqttoptions.set_transport(Transport::Ws);
     mqttoptions.set_keep_alive(Duration::from_secs(60));
 
--- a/rumqttc/examples/websocket.rs
+++ b/rumqttc/examples/websocket.rs
@@ -11,7 +11,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
     pretty_env_logger::init();
 
     // port parameter is ignored when scheme is websocket
-    let mut mqttoptions: MqttOptions = MqttOptions::new("clientId-1", "ws://localhost:9001", 9001);
+    let mut mqttoptions = MqttOptions::new(
+        "clientId-aSziq39Bp3",
+        "ws://broker.mqttdashboard.com:8000/mqtt",
+        8000,
+    );
     mqttoptions.set_transport(Transport::Ws);
     mqttoptions.set_keep_alive(Duration::from_secs(60));
 
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -403,7 +403,6 @@ async fn network_connect(
 
             let (socket, response) =
                 async_tungstenite::tokio::client_async(request, tcp_stream).await?;
-
             validate_response_headers(response)?;
 
             Network::new(WsStream::new(socket), options.max_incoming_packet_size)
@@ -423,7 +422,6 @@ async fn network_connect(
                 Some(connector),
             )
             .await?;
-
             validate_response_headers(response)?;
 
             Network::new(WsStream::new(socket), options.max_incoming_packet_size)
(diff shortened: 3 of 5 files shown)


mosquitto.conf
de-sh, line 1: Remove this file

rumqttc/examples/websocket.rs
de-sh, line 1: Undo these changes

rumqttc/src/eventloop.rs
```diff
@@ -398,7 +401,14 @@ async fn network_connect(
...
 
-            let (socket, _) = async_tungstenite::tokio::client_async(request, tcp_stream).await?;
+            let (socket, response) =
+                async_tungstenite::tokio::client_async(request, tcp_stream).await?;
+
+            response
+                .headers()
+                .get("Sec-WebSocket-Protocol")
+                .filter(|x| x == "mqtt")
+                .ok_or(Err(ConnectionError::WebsocketSubprotocol));
```
de-sh, line 411:
```suggestion
                .ok_or(Err(ConnectionError::WebsocketSubprotocol))?;
```

rumqttc/src/eventloop.rs
```diff
@@ -411,13 +421,19 @@ async fn network_connect(
...
                 tcp_stream,
                 Some(connector),
             )
             .await?;
 
+            response
+                .headers()
+                .get("Sec-WebSocket-Protocol")
+                .filter(|x| x == "mqtt")
+                .ok_or(Err(ConnectionError::WebsocketSubprotocol));
```
de-sh, line 435:
```suggestion
                .ok_or(Err(ConnectionError::WebsocketSubprotocol))?;
```

de-sh (changes requested):

rumqttc/src/eventloop.rs
Vilayat-Ali: done

rumqttc/src/eventloop.rs
Vilayat-Ali: done

de-sh:
@swanandx any plans on fixing clippy suggestions?

swanandx:
> any plans on fixing clippy suggestions?
>
me must haha, should we open separate PR for it ( would need to sync up this branch after merging it )? or just fix it in this one?

> Test against a broker that supports websocket
>
will do!

rumqttc/src/websockets.rs
```diff
@@ -8,6 +10,33 @@ pub enum UrlError {
     Parse(#[from] http::uri::InvalidUri),
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum ValidationError {
+    #[error("Websocket response does not contain subprotocol header")]
+    SubprotocolHeaderMissing,
+    #[error("Websocket subprotocol header: {0}")]
+    SubprotocolMqtt(String),
```
swanandx, line 18: can we rename this variant to something that will indicate that we can't find "mqtt" in subprotocols list?

rumqttc/src/websockets.rs
```diff
@@ -8,6 +10,33 @@ pub enum UrlError {
...
+        .get("Sec-WebSocket-Protocol")
+        .ok_or(ValidationError::SubprotocolHeaderMissing)?;
+
+    let sub_protocol = val
+        .to_str()
+        .map_err(|_| ValidationError::SubprotocolHeaderMissing)?;
+
+    if sub_protocol != "mqtt" {
+        return Err(ValidationError::SubprotocolMqtt(sub_protocol.to_owned()));
+    }
```
swanandx, line 35: as per specs [ref](https://http.dev/ws#sec-websocket-protocol):
> This header can appear multiple times within the request, or instead, multiple subprotocols can be listed in a comma-delimited format
> 

here, we aren't covering cases where multiple protocols are listed. e.g.
`Sec-WebSocket-Protocol: something-idk, mqtt`

suggestions: we can split the header at `,` and then check if `mqtt` exists!

rumqttc/src/websockets.rs
```diff
@@ -8,6 +10,33 @@ pub enum UrlError {
...
+    response: Response<Option<Vec<u8>>>,
+) -> Result<(), ValidationError> {
+    let val = response
+        .headers()
+        .get("Sec-WebSocket-Protocol")
+        .ok_or(ValidationError::SubprotocolHeaderMissing)?;
+
+    let sub_protocol = val
+        .to_str()
+        .map_err(|_| ValidationError::SubprotocolHeaderMissing)?;
```
swanandx, line 31: if we can't convert it to string ( i.e. it isn't utf8 ) shouldn't mean it was missing right? we can probably use the other variant which shows it wasn't "mqtt"?

swanandx (commented):
Tested with rumqttd when it sends "mqtt" and when doesn't ( in which case it closed the connection as expected ). Though this PR isn't covering all the cases ( explained in review comments ), please address them!

Otherwise, great work with PR :100:

de-sh:
@Vilayat-Ali please go through the changes suggested by @swanandx and the last two commits made to fix the PR along the repo code style. Please note the reorg of code and the naming conventions used.

PR #726, opened 2023-10-07
Validate "mqtt" exists in comma delimited subprotocol header 

swanandx:
This PR adds up on #724 and fixes behavior incase the header contains multiple subprotocols listed in a comma-delimited format.

## Type of change

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Allow keep alive values <= 5 seconds (#643)
+- Verify "mqtt" is present in websocket subprotocol header.
 
 ### Security
 - Remove dependency on webpki. [CVE](https://rustsec.org/advisories/RUSTSEC-2023-0052)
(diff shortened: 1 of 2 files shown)


de-sh:
Here we are verifying the response from the websocket server, we should expect only one protocol in the response.

> The server will respond using the Sec-WebSocket-Protocol header to confirm which subprotocol will be used. This server will include this header only once in the response.

swanandx:
Header will be included only once, but it's valid to specify multiple subprotocols in it right?

e.g. if we want to specify two protocols, one option is to add that header multiple times

```
Sec-WebSocket-Protocol: mqtt
Sec-WebSocket-Protocol: chat
```

Or we can only include it once

```
Sec-WebSocket-Protocol: mqtt, chat
```

I think the quoted text is referring to later case.

de-sh:
The main point of the quote is:
> The server will respond using the Sec-WebSocket-Protocol header to confirm which subprotocol will be used
i.e. the server must only respond with one chosen protocol, the first one from the list provided as header in the connection request

swanandx:
Okie! thanks for pointing out, the standards should be more explicit ig, causing confusion in interpretations haha. Anyways, i tested different clients:
paho ( python ) - doesn't verify the header at all!
mqttx - accepts only when it is "mqtt" as you mentioned.
mosquitto - can't connect over ws 

So it would be fine if we just match "mqtt". Can you please change the CHANGELOG entry as well in PR #729 ( cuz CHANGELOG says we accept comma-delimated values! ) :smile:

PR #728, opened 2023-10-07
update tungstenite dependencies for security purpose

Arend-Jan:
As described in [rustsec](https://rustsec.org/advisories/RUSTSEC-2023-0065), there is a vulnerability in older versions of tungstenite. This library is used in the chain of dependencies of rumqtt. This change up-steps those dependencies.

## Type of change

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `MqttOptions::set_request_modifier` for setting a handler for modifying a websocket request before sending it.
 
 ### Changed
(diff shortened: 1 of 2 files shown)


de-sh (approved):
LGTM, thanks for the contribution!

rumqttc/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Added `MqttOptions::set_request_modifier` for setting a handler for modifying a websocket request before sending it.
 
 ### Changed
+- Removed dependency vulnerability, see [rustsec](https://rustsec.org/advisories/RUSTSEC-2023-0065). Update of `tungstenite` dependency.
```
swanandx, line 16: Can we move this entry under `### Security` section please, so it would follow previous changes regarding webpki?

PR #734, opened 2023-10-17
Enable websocket feature by default

swanandx:
This PR does following:
- enable websocket feature by default
- remove old "websockets" feature
- log warning if ws config is getting ignored

## Type of change

## Checklist:


henil (commented):
changes LGTM, but i rememver last time having a feature with same name between `rumqttc` and `rumqttd` caused some problem. Can you check if its still the case?

more context here: https://github.com/bytebeamio/rumqtt/pull/633

henil (approved):

Issue #672, opened 2023-08-06
rumqttd | make console config optional


crimx:
# Expected Behavior

I am not sure but the `console` seems to be a debugging feature and can be turned off right? I want to only register port for v4.

# Current Behavior

The `console` config of rumqttd is required.

P.S. Thanks for providing this great project!

swanandx:
Yup, we can make console optional and it should be pretty easy to do so. Thank you so much 😁

PR #736, opened 2023-10-19
Make v4 config optional

swanandx:
This PR makes v4 config optional so that users can specify v4 and/or v5 config ( or ws ) as per their use case.

This will allow user to use only v5 if they wish.

## Type of change

## Checklist:


henil (approved):

id4vip:
Hello, thanks for making these changes. just want to clarify. is console portion of rumqttd toml file also become optional in this upgrade ? or it will remain mandatory part.

swanandx:
It's not done in this PR, but we do plan to make it optional. I wished some new contributor would tackle #672 . Let's wait for some time, else I will resolve it as well!

PR #737, opened 2023-10-22
console configuration should be optional

### LINKED ISSUE #672: rumqttd | make console config optional

swanandx:
Fixes #672 . 

Changes:
- make console settings optional
- instead of blocking on console in `start()`, we block on joining thread handle of spawned servers.

## Type of change

## Checklist:

### DIFF
--- a/rumqttd/src/server/broker.rs
+++ b/rumqttd/src/server/broker.rs
@@ -166,6 +166,10 @@ impl Broker {
             ));
         }
 
+        // we don't know which servers (v4/v5/ws) user will spawn
+        // so we collect handles for all of the spawned servers
+        let mut server_thread_handles = Vec::new();
+
         if let Some(metrics_config) = self.config.metrics.clone() {
             let timer_thread = thread::Builder::new().name("timer".to_owned());
             let router_tx = self.router_tx.clone();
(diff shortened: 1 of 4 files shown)


rumqttd/src/server/broker.rs
```diff
@@ -297,12 +304,26 @@ impl Broker {
...
-        let runtime = runtime.enable_all().build().unwrap();
-        runtime.block_on(console::start(console_link));
+        // in ideal case, where server doesn't crash, join() will never resolve
+        // we still try to join threads so that we don't return from function
+        // unless everything crashes.
+        server_thread_handles.into_iter().for_each(|handle| {
+            // join() might panic in case the thread panics
+            // we just ignore it
+            let _ = handle.join();
+        });
```
henil, line 326: are panic messages still printed if anyone the threads panic?

asking since docs of join say it returns panic error message as `Err`.

if its not printed we should print them here.

swanandx: yes, the panic messages are printed. Tested by adding `panic!(..)` in `consume()` method of router so it can panic easily lol

PR #738, opened 2023-10-22
style: remove unused parts

de-sh:
## Type of change
Code improvements, removing unused parts such as empty file and impl serialize derives

## Checklist:

### DIFF
--- a/rumqttd/CHANGELOG.md
+++ b/rumqttd/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 ### Changed
+- Cleanup code of empty `connections.rs` and `derive(Serialize)`
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


swanandx:
to fix the CI failure caused due to clippy warning, we can put `ClientAuth` under `#[cfg(feature = "use-rustls")]`.
Doing so will also mean we have to put declaration of `Transport::Tls{..}` variant under same feature flag ( which is good thing tbh, cuz the usage of this variant is feature gated already ).

If this is fine, shall I push the commit which does so and fixes this warnings? @de-sh

swanandx:
btw, considering users can use `rumqttd` as library and build things on top of it, I believe we should keep `derive(Serialize)`.

one example I can think from top of my head is when users wants to return the config on request to some endpoint. we would need to _serialize_ config in such cases right?

de-sh:
> we would need to _serialize_ config in such cases right?

Let's include the derive when we implement that, will contain better context on why it is there, then.

PR #739, opened 2023-10-25
fix(rumqttc/examples): Fix rumqttc async_manual_acks.rs

jliklik:
- Allow code to continue after first loop finishes

## Type of change

## Checklist:

### DIFF
--- a/rumqttc/examples/async_manual_acks.rs
+++ b/rumqttc/examples/async_manual_acks.rs
@@ -1,6 +1,6 @@
 use tokio::{task, time};
 
-use rumqttc::{self, AsyncClient, Event, EventLoop, Incoming, MqttOptions, QoS};
+use rumqttc::{self, AsyncClient, Event, EventLoop, Incoming, MqttOptions, Outgoing, QoS};
 use std::error::Error;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


rumqttc/examples/async_manual_acks.rs
```diff
@@ -66,6 +62,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
                 return Ok(());
             }
         }
+        if let Ok(Event::Outgoing(Outgoing::PubAck(puback))) = event {
+            if puback == 10 {
+                return Ok(());
+            }
+        }
```
swanandx, line 69: > Allow program to terminate after all pubacks have been sent
>
we don't need to terminate the program, we can just let it run normally. It will keep sending ping req and getting ping resp like other examples.

swanandx (commented):
Thanks for the PR, please go through review comment and feel free to let me know what do you think.

rumqttc/examples/async_manual_acks.rs
jliklik: Sounds good. Will remove this section.

swanandx (approved):

PR #741, opened 2023-10-31
feat: expose `EventLoop::clean`

de-sh:
## Type of change

Make public the `EventLoop::clean` method to allow for storage of pending requests onto disk in bytebeamio/uplink

## Checklist:

### DIFF
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -120,7 +120,10 @@ impl EventLoop {
     }
 
     /// Last session might contain packets which aren't acked. MQTT says these packets should be
-    /// republished in the next session. Move pending messages from state to eventloop.
+    /// republished in the next session. Move pending messages from state to eventloop, drops the
+    /// underlying network connection and clears the keepalive timeout if any.
+    /// 
+    /// NOTE: Use only when EventLoop is blocked on network and unable to immediately handle disconnect
     pub fn clean(&mut self) {
         self.network = None;
         self.keepalive_timeout = None;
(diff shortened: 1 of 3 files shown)


rumqttc/src/eventloop.rs
```diff
@@ -119,11 +119,12 @@ impl EventLoop {
         }
     }
 
-    fn clean(&mut self) {
+    /// Last session might contain packets which aren't acked. MQTT says these packets should be
+    /// republished in the next session. Move pending messages from state to eventloop.
+    pub fn clean(&mut self) {
```
swanandx, line 124: the docs comments on this fn should also mention it would drop the network and keepalive, and thus disconnect the connection with broker ( but won't  send disconnect packet like `client.disconnect()` does )

PS: would be nice if we can specify when we should/ shouldn't use this instead of disconnect ( not that imp though )

swanandx (changes requested):
Please also copy this changes in `v5/eventloop.rs` so we stay consistent with the API

PS: what is the benefit or in which situations would someone call `clean()`? If user wants to disconnect client, they can just call `client.disconnect()` and it would call the `clean()` method internally. That would be much more cleaner right? does is has something to do with who closes the network connection first?

de-sh:
> what is the benefit or in which situations would someone call clean()? If user wants to disconnect client, they can just call client.disconnect() and it would call the clean() method internally. That would be much more cleaner right? does is has something to do with who closes the network connection first?

swanandx (approved):

PR #743, opened 2023-11-03
Allow ECC Key Configuration with rustls

AlexandreCassagne:
## Type of change

## Checklist:

#### Possible remaining actions 
- [ ] I have not included any tests yet. If you think this PR may be accepted I will look at that Monday
- [ ] I think that we should add a mention in the documentation for this feature
- [ ] I have made the choice of adding the `ecc_` prefix. Perhaps it makes more sense to detect the header or algorithm values found in the certificate (` ----- BEGIN EC PRIVATE KEY ----- `). Also, I didn't want to add a breaking change, but this assumes the RSA certs are the defaults.
- [ ] Consider implementing the above breaking change (maybe in future PR), with the following approach:
-   `rsa_` prefix allows current certs (such as the ones generated by provision, which I think are PKCS#1), i.e `BEGIN RSA PRIVATE KEY`
-  `ec_` or `ecc_` prefix with the `BEGIN EC PRIVATE KEY`
- no prefix for universal private key files, of the form `BEGIN PRIVATE KEY`

Anyway, someone more senior in this project can help make those decisions. I don't fully understand if there was a reason for hardcoding RSA in the code, but my use case requires ECC (embedded environments)

### DIFF
--- a/rumqttd/src/lib.rs
+++ b/rumqttd/src/lib.rs
@@ -3,7 +3,6 @@ use std::path::PathBuf;
 use std::{collections::HashMap, path::Path};
 
 use serde::{Deserialize, Serialize};
-use tracing::{debug, trace};
 use tracing_subscriber::{
     filter::EnvFilter,
     fmt::{
--- a/rumqttd/src/lib.rs
+++ b/rumqttd/src/lib.rs
@@ -82,7 +82,7 @@ impl TlsConfig {
     // Returns true only if all of the file paths inside `TlsConfig` actually exists on file system.
     // NOTE: This doesn't verify if certificate files are in required format or not.
     pub fn validate_paths(&self) -> bool {
-        let result = match self {
+        match self {
             TlsConfig::Rustls {
                 capath,
                 certpath,
@@ -91,9 +91,7 @@ impl TlsConfig {
                 .iter()
                 .all(|v| Path::new(v).exists()),
             TlsConfig::NativeTls { pkcs12path, .. } => Path::new(pkcs12path).exists(),
-        };
-        trace!("Validating TLS config paths with result: {:?}", result);
-        result
+        }
     }
 }
 
--- a/rumqttd/src/lib.rs
+++ b/rumqttd/src/lib.rs
@@ -1,6 +1,6 @@
-use std::{collections::HashMap, path::Path};
 use std::net::SocketAddr;
 use std::path::PathBuf;
+use std::{collections::HashMap, path::Path};
 
 use serde::{Deserialize, Serialize};
 use tracing::{debug, trace};
@@ -11,8 +11,8 @@ use tracing_subscriber::{
         Layer,
     },
     layer::Layered,
-    Registry,
     reload::Handle,
+    Registry,
 };
 
 pub use link::alerts;
--- a/rumqttd/src/lib.rs
+++ b/rumqttd/src/lib.rs
@@ -72,11 +72,6 @@ pub enum TlsConfig {
         certpath: String,
         keypath: String,
     },
-    RustlsWithECC {
-        capath: String,
-        ecc_certpath: String,
-        ecc_keypath: String,
-    },
     NativeTls {
         pkcs12path: String,
         pkcs12pass: String,
@@ -93,13 +88,6 @@ impl TlsConfig {
                 certpath,
                 keypath,
             } => [capath, certpath, keypath]
-                .iter()
-                .all(|v| Path::new(v).exists()),
-            TlsConfig::RustlsWithECC {
-                capath,
-                ecc_certpath,
-                ecc_keypath,
-            } => [capath, ecc_certpath, ecc_keypath]
                 .iter()
                 .all(|v| Path::new(v).exists()),
             TlsConfig::NativeTls { pkcs12path, .. } => Path::new(pkcs12path).exists(),
--- a/rumqttd/src/lib.rs
+++ b/rumqttd/src/lib.rs
@@ -3,6 +3,7 @@ use std::net::SocketAddr;
 use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
+use tracing::{debug, trace};
 use tracing_subscriber::{
     filter::EnvFilter,
     fmt::{
@@ -86,7 +87,7 @@ impl TlsConfig {
     // Returns true only if all of the file paths inside `TlsConfig` actually exists on file system.
     // NOTE: This doesn't verify if certificate files are in required format or not.
     pub fn validate_paths(&self) -> bool {
-        match self {
+        let result = match self {
             TlsConfig::Rustls {
                 capath,
                 certpath,
@@ -102,7 +103,9 @@ impl TlsConfig {
                 .iter()
                 .all(|v| Path::new(v).exists()),
             TlsConfig::NativeTls { pkcs12path, .. } => Path::new(pkcs12path).exists(),
-        }
+        };
+        trace!("Validating TLS config paths with result: {:?}", result);
+        result
     }
 }
 
--- a/rumqttd/src/lib.rs
+++ b/rumqttd/src/lib.rs
@@ -1,7 +1,7 @@
-use std::path::PathBuf;
 use std::{collections::HashMap, path::Path};
+use std::net::SocketAddr;
+use std::path::PathBuf;
 
-use segments::Storage;
 use serde::{Deserialize, Serialize};
 use tracing_subscriber::{
     filter::EnvFilter,
@@ -10,11 +10,18 @@ use tracing_subscriber::{
         Layer,
     },
     layer::Layered,
-    reload::Handle,
     Registry,
+    reload::Handle,
 };
 
-use std::net::SocketAddr;
+pub use link::alerts;
+pub use link::local;
+pub use link::meters;
+pub use router::{Alert, IncomingMeter, Meter, Notification, OutgoingMeter};
+use segments::Storage;
+pub use server::Broker;
+
+use self::router::shared_subs::Strategy;
 
 mod link;
 pub mod protocol;
@@ -30,15 +37,6 @@ pub type Filter = String;
 pub type TopicId = usize;
 pub type Offset = (u64, u64);
 pub type Cursor = (u64, u64);
-
-pub use link::alerts;
-pub use link::local;
-pub use link::meters;
-
-pub use router::{Alert, IncomingMeter, Meter, Notification, OutgoingMeter};
-pub use server::Broker;
-
-use self::router::shared_subs::Strategy;
 
 #[derive(Debug, Default, Serialize, Deserialize, Clone)]
 pub struct Config {
@@ -73,6 +71,11 @@ pub enum TlsConfig {
         certpath: String,
         keypath: String,
     },
+    RustlsWithECC {
+        capath: String,
+        ecc_certpath: String,
+        ecc_keypath: String,
+    },
     NativeTls {
         pkcs12path: String,
         pkcs12pass: String,
@@ -89,6 +92,13 @@ impl TlsConfig {
                 certpath,
                 keypath,
             } => [capath, certpath, keypath]
+                .iter()
+                .all(|v| Path::new(v).exists()),
+            TlsConfig::RustlsWithECC {
+                capath,
+                ecc_certpath,
+                ecc_keypath,
+            } => [capath, ecc_certpath, ecc_keypath]
                 .iter()
                 .all(|v| Path::new(v).exists()),
             TlsConfig::NativeTls { pkcs12path, .. } => Path::new(pkcs12path).exists(),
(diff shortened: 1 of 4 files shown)


swanandx:
Hey, thanks for the PR 🚀

I believe introduction of new Enum variant and other changes can be avoided. If you see the [docs](https://docs.rs/rustls-pemfile/latest/rustls_pemfile/#example-code) of rustls_pemfile, we can use something similar to extract only the keys, and as it checks headers ( rsa, ecc etc. ) internally, it should work fine right?

Basically, instead of calling `rsa_private_keys()`, extract the keys using `read_one()` as shown in example ( we can use `read_all()` and then discard the other variants of `Item` which aren't key, but ignoring them as we go seems much elegant ).

Hope this helps! let me know what do you think 😁

AlexandreCassagne:
Thanks for your reply! Yes, that makes a lot of sense. I will play with read_one/read_all and update my PR. 

I will return the _first_ EC or RSA private key that appears by invoking read_one() repeatedly on the PEM file.

swanandx:
Hey @AlexandreCassagne 👋 any updates on the PR? Is it ready for review?

AlexandreCassagne:
Hey @swanandx,

 I do think it is finished! However, do you think we should add additional tests?

AlexandreCassagne:
I'm thinking in particular that these examples (rumqttc/examples/tls2.rs) should be converted to tests, and include one for each supported key type

rumqttd/src/lib.rs
```diff
@@ -92,7 +91,9 @@ impl TlsConfig {
                 .iter()
                 .all(|v| Path::new(v).exists()),
             TlsConfig::NativeTls { pkcs12path, .. } => Path::new(pkcs12path).exists(),
-        }
+        };
+        trace!("Validating TLS config paths with result: {:?}", result);
+        result
```
swanandx, line 96: I believe this log should be where the function is actually used, not inside the function! wdyt? also, is this helpful as if paths aren't validated, we return error anyway right??

rumqttd/src/server/tls.rs
```diff
@@ -187,18 +187,10 @@ impl TLSAcceptor {
...
-
-            // Get the first key
-            let key = match keys.first() {
-                Some(k) => k.clone(),
-                None => return Err(Error::InvalidServerKey(key_path.clone())),
-            };
-
-            (certs, PrivateKey(key))
+            let key = first_private_key_in_pemfile(key_path)
+                .map_err(|_| Error::InvalidServerKey(key_path.clone()))?;
```
swanandx, line 191: any type of `Error` returned from `first_private_key_in_pemfile` will get mapped to `InvalidServerKey`, that should not happen. 

maybe we can check `Error::ServerKeyNotFound` here and then pass in `key_file` to the fn.

rumqttd/src/server/tls.rs
```diff
@@ -227,3 +219,33 @@ impl TLSAcceptor {
...
+    loop {
+        match rustls_pemfile::read_one(rd) {
+            Ok(maybe_inner) => match maybe_inner {
+                None => {
+                    error!("No private key found in {:?}", key_path);
+                    return Err(Error::InvalidServerKey(key_path.clone()));
+                }
+                Some(Item::ECKey(key)) => return Ok(PrivateKey(key)),
+                Some(Item::RSAKey(key)) => return Ok(PrivateKey(key)),
+                Some(Item::PKCS8Key(key)) => return Ok(PrivateKey(key)),
```
swanandx, line 242: will it be better to use `|` like `Some(Item::ECKey | Item::RSAKey | ..)` ? 
ps: only do this if it doesn't compromise readability!

nit: can we move `None` below this? as it is less intended.

rumqttd/src/server/tls.rs
```diff
@@ -227,3 +219,33 @@ impl TLSAcceptor {
...
+                Some(_) => {}
+            },
+            Err(err) => {
+                error!("Error reading key file: {:?}", err);
```
swanandx, line 246: hope this `err` doesn't flood logs unnecessarily, plz check it once!

rumqttd/src/server/tls.rs
```diff
@@ -1,21 +1,21 @@
+#[cfg(feature = "use-rustls")]
+use std::{io::BufReader, sync::Arc};
 use std::fs::File;
-
 #[cfg(feature = "use-native-tls")]
 use std::io::Read;
+
+use rustls_pemfile::Item;
```
swanandx, line 7: this should be feature gated like other imports

AlexandreCassagne:
thanks for review - will apply your recommendations now

rumqttd/src/lib.rs
AlexandreCassagne: I've changed the traces as follows:


```rust
trace!("Validated certificate paths for server v4.{name}.");
trace!("Validated certificate paths for server v5.{name}.");
trace!("Validated certificate paths for server ws.{name}.");
```

at the caller site

rumqttd/src/server/tls.rs
AlexandreCassagne: Not sure what you mean here because it returns after reaching the log:

```rust
            Err(err) => {
                error!("Error reading key file: {:?}", err);
                **return Err(Error::InvalidServerKey(key_path.clone()));**
            }
```

rumqttd/src/server/tls.rs
AlexandreCassagne: For this one, you want me to 

- Check if `Error::ServerKeyNotFound` -> if so, return it
- Otherwise, unwrap the existing error (without mapping to InvalidSeverKey)

Did I understand that right ?

rumqttd/src/server/tls.rs
AlexandreCassagne: Done, now looks like this:


```rust
    loop {
        match rustls_pemfile::read_one(rd) {
            Ok(maybe_inner) => match maybe_inner {
                Some(item) => match item {
                    // If we find a private key, return it
                    Item::ECKey(key) | Item::RSAKey(key) | Item::PKCS8Key(key) => {
                        return Ok(PrivateKey(key))
                    }
                    // Otherwise, continue
                    _ => {}
                },
                None => {
                    error!("No private key found in {:?}", key_path);
                    return Err(Error::InvalidServerKey(key_path.clone()));
                }
            },
            Err(err) => {
                error!("Error reading key file: {:?}", err);
                return Err(Error::InvalidServerKey(key_path.clone()));
            }
        }
    }
```

rumqttd/src/lib.rs
swanandx: I wanted to say that, is the certs aren't valid, we are going to panic, so what's the use of this trace? I believe it's not providing any helpful info. Thoughts?

rumqttd/src/server/tls.rs
swanandx: - yes
- ? is different from unwrap, if we unwrap, it would panic, we don't want that. ? will just return the error, which is what we intend

rumqttd/src/server/tls.rs
swanandx: In this line, we print the `err`

```rust
error!("Error reading key file: {:?}", err);
```

I'm not sure how long / verbose that returned error will be, so just wanted to confirm if it's not that long! Something crazy would be like printing the whole wrong key file.

If it's too long, we can avoid printing the exact err

rumqttd/src/server/tls.rs
swanandx: we should also note that this is giving preference like `EC` > `RSA` > `PKCS8`! maybe add it as a comment?

rumqttd/src/server/tls.rs
swanandx: update: after reading docs / source, it seems it will be:

```
Error reading key file: Err(Error::InvalidData("MIIIPDCCBySgAwIBAgIQB2XGTnTlkdaAOcoqhHVj8DANBgkqhkiG9w0BAQsFADBw"))
```

Or something similar to this which has whole line. Don't think it's that long, so this should be fine. Plz correct me if i got something wrong, thanx 😁

rumqttd/src/server/tls.rs
swanandx: pushed commit to fix this :+1:

swanandx:
Hey, pushed some commits to fix clippy warnings & some fmt, plz have a look once!

> Do you think we should add additional tests?
>

ideally we should! but not sure how easy / hard it would get. so for now, let us manually test if everything is working as expected to unblock merging of this PR. Can you please just test it once with different keys? 

Thank you!

rumqttd/src/server/tls.rs
```diff
@@ -228,29 +227,24 @@ fn first_private_key_in_pemfile(key_path: &String) -> Result<PrivateKey, Error>
...
-                    error!("No private key found in {:?}", key_path);
-                    return Err(Error::InvalidServerKey(key_path.clone()));
-                }
-            },
-            Err(err) => {
-                error!("Error reading key file: {:?}", err);
+        let item = rustls_pemfile::read_one(rd).map_err(|err| {
+            error!("Error reading key file: {:?}", err);
+            Error::InvalidServerKey(key_path.clone())
+        })?;
```
AlexandreCassagne, line 237: much cleaner

AlexandreCassagne:
Sounds good. I have tested once with ECCs and will check RSAs

AlexandreCassagne:
@swanandx It was a nightmare but thankfully it worked. I only implemented RSA tests, I will add an EC variant of the test before we can merge.

Note that rumqttc suffers from the same logic problems that we noticed in rumqttd for tls.

I am not very proud of the last commit that patches it (it is ugly).

I think it requires a stronger refactor of that method -- I'd suggest reusing the `first_private_key_in_pemfile` we included earlier in this PR. That will probably require some rewrites in the function, possibly also suggesting that we should also remove the RSA/EC rumqttc `Key` enum (at `rumqttc/src/lib.rs:209`, delegating that responsibility to rustls' `read_one` method similarly to the rumqttd approach

However, all this is out of scope here --- would you mind reviewing the minimal changes I made to the client, along with the tests? And perhaps we move the above suggestions to a different PR to unblock?

AlexandreCassagne:
Also worth reviewing [this commit](https://github.com/bytebeamio/rumqtt/pull/743/commits/40df4e6b658031374863c7fcdeffacb1e82a01e4).

swanandx:
Hey @AlexandreCassagne , thanks for the efforts  and awesome work with getting it to actually work lol  :100: 

I do appreciate your work, but tbh this is a bit overkill, we should not modify rumqttc in this PR ( as this is strictly related to broker ). And for testing part, using rumqttc for testing rumqttd is cool, but it would be better to test broker externally ( maybe with tools like [mqttwrk](https://github.com/bytebeamio/mqttwrk) for conformace, which does the same, though it doesn't have tls yet ) for now ( getting out of scope of this PR as well haha )


> suggesting that we should also remove the RSA/EC rumqttc Key enum
>
This is tackled in #752 

will it be possible to revert these changes? ( this is the last working [commit](https://github.com/bytebeamio/rumqtt/pull/743/commits/f4d798073f1c76d0ec793522be07de40c5f64328) iirc )

It would be better to have a separate PR just dedicated to add tests ( like basic integrity test and all ) to rumqttd in future, let's avoid that in this PR.

Thank you once again :smile:

AlexandreCassagne:
Sounds good, I'll revert and move the additional commits into another branch of mine, as for our use case, we do need to be able to run tests before we push into production.

Nonetheless, I'd suggest keeping opening another PR with some of the code I wrote for these tests eventually. I learnt a lot about how rumqttc/d implementation of TLS worked by writing this, and also I ended up understanding why the Provision certificates worked, but not the ones I made with openssl. A documentation PR may come later :)

AlexandreCassagne:
Also - one thing I noticed is that the interface of Rumqttd's configuration doesn't really allow programmatic instantiation. As a result, I was effectively forced into saving files in the file system and using toml

Are you open to a PR for allowing other means of instantiating the TLS config programmatically with rustls data structures directly?

AlexandreCassagne:
@swanandx  Alright, reverted. Backup of the 7 deleted commits is in [AlexandreCassagne/tests-tls-rumqttc-can-connect-rumqttd](https://github.com/AlexandreCassagne/rumqtt/tree/tests-tls-rumqttc-can-connect-rumqttd)

swanandx (approved):

swanandx:
> I learnt a lot about how rumqttc/d implementation of TLS worked by writing this, and also I ended up understanding why the Provision certificates worked, but not the ones I made with openssl. A documentation PR may come later :)
>

That's great :100: 

> Are you open to a PR for allowing other means of instantiating the TLS config programmatically with rustls data structures directly?
> 
I didn't get what do you mean by rustls data structures :sweat_smile: We can init the config like `Config::default()` and then edit it further right? kinda like you did in PR.

Thank you so much for the contribution :rocket:

Issue #745, opened 2023-11-09
rumqttd: Configurable or "smart" RusTLS options.


nathansizemore:
# Expected Behavior

When setting up tls with rustls, the server private key should be able to be any valid private key.

# Current Behavior

When using rumqttd with rustls, the server private key only accepts RSA, but no call outs docs regarding this.

## Context

Let's Encrypt default `privkey1.pem` acme output isn't RSA. It's a lot of stepping though code because there isn't documentation stating this to figure out it was forced to be RSA here:

```rust
// Get private key
let key_file = File::open(key_path);
let key_file = key_file.map_err(|_| Error::ServerKeyNotFound(key_path.clone()))?;
let keys = rustls_pemfile::rsa_private_keys(&mut BufReader::new(key_file));
```

It should either be specified in the config on how to read the pem files, or try and be "smart" and iterate over more options than just RSA.

swanandx:
This should be solved in #743 ! Please check the discussion there, thank you 💯

nathansizemore:
No worries. It's the same thing basically, just for the broker end. :smiley_cat:

swanandx:
> more options than just RSA.
>

Like? [rustls_pemfile](https://docs.rs/rustls-pemfile/latest/rustls_pemfile/#example-code) can read RSA, EC and PKCS8. Is there any other type?

nathansizemore:
Ideally with rustls, it would cover all options they provide from their crates. So just those would be great. Native can do whatever native does.

nathansizemore:
I can do a draft PR does tries over each and returns an error if all fail. Would that be acceptable "smartness" for acceptance?

swanandx:
> Ideally with rustls, it would cover all options they provide from their crates. So just those would be great.
>

I didn't get wdym by using rustls to cover options, rustls [docs](https://docs.rs/rustls/latest/rustls/struct.Certificate.html) suggest using rustls_pemfile to parse the PEM files.

> I can do a draft PR that tries over each and returns an error if all fail.
>

Sure, it will be similar to PR mentioned above right? Just using `read_one()` to get the key.

PS: @henil do we want this in broker? I believe we should, wdyt?

nathansizemore:
I just meant trying `ec_private_keys`, `pkcs8_private_keys`, and `rsa_private_keys` by all options they support, since they provide those with `rustls_pemfile`.

swanandx:
[...]

PR #746, opened 2023-11-09
Accept any rustls_pemfile validated key.

nathansizemore:
Iterates over rustls_pemfile supported types for the keyfile supplied and allows any to be used instead of expecting rsa.

## Type of change

## Checklist:

- [ X] Formatted with `cargo fmt`
- [ X] Make an entry to `CHANGELOG.md` if it's relevant to the users of the library. If it's not relevant mention why.

Not in changelog because it should be default behavior instead of rumqttd being opinionated about rsa keys.

### DIFF
--- a/rumqttd/src/server/tls.rs
+++ b/rumqttd/src/server/tls.rs
@@ -175,6 +175,8 @@ impl TLSAcceptor {
         cert_path: &String,
         key_path: &String,
     ) -> Result<TLSAcceptor, Error> {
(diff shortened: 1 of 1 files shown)


nathansizemore:
refs: https://github.com/bytebeamio/rumqtt/issues/745#issuecomment-1803955618

nathansizemore:
Closing to bring in changes that also allow client certs to be optional.

Issue #747, opened 2023-11-09
rumqttd: client certs should be configurable


nathansizemore:
# Expected Behavior

To be able to use rustls for tls option and not require client certs.

# Current Behavior

All tls connections for rustls are required to client cert.

## Context

Client cert tls is the minority of tls sessions in the wild. It should not be the default, or mandated, to be able to use rustls.

swanandx:
Hey @nathansizemore , I just wanted to check if I understood the issue correctly, plz correct me if i got something wrong!
- by "not require client certs" you actually mean not requiring client cert verification!
- by "Client cert tls" you mean mutual TLS ( verifying client cert )

How we can disable client cert verification:
- by making `CA` certs optional
- or we can add an extra feature / configuration flag, so that we can explicitly disable client cert verification ( even if CA cert is provided ).

Can you please elaborate more on the issue, thank you so much!

nathansizemore:
Let me rephrase...

Right now, rumqttd provides two types of TLS:
* Standard TLS where the client verifies the server is trusted
* Mutual TLS/TLS with Client Authentication/whatever fancy name where both verify. 

rumqttd should allow the user to determine how TLS is setup first, and second the provider (rustls v native). Right now, it only allows you to choose whether you want to use TLS through rustls or native. Which an unknown side effect of that is if you pick rustls, you are forced to used mTLS. If you pick native, you are forced to use standard TLS. 

The patch I provided is just the minimum to allow use of rustls with standard tls flow. For complete parity:

* Config should have the CA field as optional, but required when compiling with `validate-tenant-prefix`.
* Native and rustls impls should be the same. Otherwise, why support both? Mutual TLS has been there forever, there is no reason only rustls should have that option.
* Right now, when using rustls, `tenant_id` is pulled from the client cert. This seems like a bug in rumqttd features, since it should only apply when `validate-tenant-prefix` feature is used. This should be fixed so that `validate-tenant-prefix` requires that mutual tls be used, which should also flip the CA field in the config to be forced.

swanandx:
gotcha! so we have two different things to do to tackle this two issues:
- making CA optional ( so mTLS is optional )
- feature gate `extract_tenant_id` logic under `validate-tenant-prefix`

right? @de-sh suggested we should tackle this in different PRs for each, would you please update the PR #748 accordingly or shall I do it? Thank you so much :100:

nathansizemore:
[...]

PR #748, opened 2023-11-09
Allow rustls configuration in rumqttd

nathansizemore:
* Allows private key to be any flavor supported by [`rustls_pemfile`](https://docs.rs/rustls-pemfile/latest/rustls_pemfile/).
* Allows TLS connections without client certs.

Removes setting the tenant id by client cert (keeps native and rustls impls the same).

closes #746 #747 

## Type of change

## Checklist:

### DIFF
--- a/rumqttd/src/server/tls.rs
+++ b/rumqttd/src/server/tls.rs
@@ -175,6 +175,8 @@ impl TLSAcceptor {
         cert_path: &String,
         key_path: &String,
     ) -> Result<TLSAcceptor, Error> {
+        use std::iter;
+
         let (certs, key) = {
             // Get certificates
             let cert_file = File::open(cert_path);
@@ -189,8 +191,16 @@ impl TLSAcceptor {
             // Get private key
             let key_file = File::open(key_path);
             let key_file = key_file.map_err(|_| Error::ServerKeyNotFound(key_path.clone()))?;
-            let keys = rustls_pemfile::rsa_private_keys(&mut BufReader::new(key_file));
-            let keys = keys.map_err(|_| Error::InvalidServerKey(key_path.clone()))?;
+            let mut key_file_buf = BufReader::new(key_file);
+            let mut keys = Vec::<Vec<u8>>::new();
+            for item in iter::from_fn(|| rustls_pemfile::read_one(&mut key_file_buf).transpose()) {
+                match item.unwrap() {
+                    rustls_pemfile::Item::RSAKey(key) => keys.push(key),
+                    rustls_pemfile::Item::PKCS8Key(key) => keys.push(key),
+                    rustls_pemfile::Item::ECKey(key) => keys.push(key),
+                    _ => return Err(Error::InvalidServerKey(key_path.clone())),
+                }
+            }
 
             // Get the first key
             let key = match keys.first() {
--- a/rumqttd/src/server/tls.rs
+++ b/rumqttd/src/server/tls.rs
@@ -120,13 +120,13 @@ impl TLSAcceptor {
             #[cfg(feature = "use-rustls")]
             TLSAcceptor::Rustls { acceptor } => {
                 let stream = acceptor.accept(stream).await?;
-                let (_, session) = stream.get_ref();
-                let peer_certificates = session
-                    .peer_certificates()
-                    .ok_or(Error::NoPeerCertificate)?;
-                let tenant_id = extract_tenant_id(&peer_certificates[0].0)?;
+                // let (_, session) = stream.get_ref();
+                // let peer_certificates = session
+                //     .peer_certificates()
+                //     .ok_or(Error::NoPeerCertificate)?;
+                // let tenant_id = extract_tenant_id(&peer_certificates[0].0)?;
                 let network = Box::new(stream);
-                Ok((tenant_id, network))
+                Ok((None, network))
             }
             #[cfg(feature = "use-native-tls")]
             TLSAcceptor::NativeTLS { acceptor } => {
@@ -171,7 +171,7 @@ impl TLSAcceptor {
 
     #[cfg(feature = "use-rustls")]
     fn rustls(
-        ca_path: &String,
+        ca_path: &Option<String>,
         cert_path: &String,
         key_path: &String,
     ) -> Result<TLSAcceptor, Error> {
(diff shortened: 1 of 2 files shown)


rumqttd/src/server/tls.rs
```diff
@@ -189,8 +191,16 @@ impl TLSAcceptor {
...
+            let mut key_file_buf = BufReader::new(key_file);
+            let mut keys = Vec::<Vec<u8>>::new();
+            for item in iter::from_fn(|| rustls_pemfile::read_one(&mut key_file_buf).transpose()) {
+                match item.unwrap() {
+                    rustls_pemfile::Item::RSAKey(key) => keys.push(key),
+                    rustls_pemfile::Item::PKCS8Key(key) => keys.push(key),
+                    rustls_pemfile::Item::ECKey(key) => keys.push(key),
+                    _ => return Err(Error::InvalidServerKey(key_path.clone())),
+                }
+            }
```
swanandx, line 203: - we only need one key, so we can avoid creating a `Vec<_>` and stop the loop once we get or first key!
- call to `item.unwrap()` should be avoided, we can return key not found or some other error.
- If read item is a cert, we still want to continue the loop, instead of returning error. This is how `rsa_private_key()` or other fns work, you can see their source in docs.
- That `transpose()` call can be avoided. we can use `loop{}` and `break` to read_one instead of creating iter.
- You can also refer to PR mentioned in issue related to this to get an clear idea!

rumqttd/src/server/tls.rs
```diff
@@ -171,10 +171,12 @@ impl TLSAcceptor {
 
     #[cfg(feature = "use-rustls")]
     fn rustls(
-        ca_path: &String,
+        ca_path: &Option<String>,
```
swanandx, line 174: shouldn't it be `Option<&String>`?

rumqttd/src/server/tls.rs
```diff
@@ -120,13 +120,13 @@ impl TLSAcceptor {
...
-                    .ok_or(Error::NoPeerCertificate)?;
-                let tenant_id = extract_tenant_id(&peer_certificates[0].0)?;
+                // let (_, session) = stream.get_ref();
+                // let peer_certificates = session
+                //     .peer_certificates()
+                //     .ok_or(Error::NoPeerCertificate)?;
+                // let tenant_id = extract_tenant_id(&peer_certificates[0].0)?;
                 let network = Box::new(stream);
-                Ok((tenant_id, network))
+                Ok((None, network))
```
swanandx, line 129: We do want to have ability to get `tenant_id` , and getting it from certs is the best way ig. Is there any other way around it? 

If there or no certs, is it `None` or `Some([])` ? I think we can just check:
```rust
if !peer_certificates.is_empty() {
  // call extract here
}
// otherwise tenand_id will be None
```

cc: any comments @henil ?

swanandx (commented):
Thanks for the PR! 

> Removes setting the tenant id by client cert 
>
We do need a way to specify tenant_id though, rather than removing this, we can make it optional? more on it in review comment.

can we tackle both mentioned issues in different PRs? easier to review / keep track of changes. Anyways, choice is up to you haha

nathansizemore:
> Thanks for the PR!
> 
> > Removes setting the tenant id by client cert
> 
> We do need a way to specify tenant_id though, rather than removing this, we can make it optional? more on it in review comment.
> 
> can we tackle both mentioned issues in different PRs? easier to review / keep track of changes. Anyways, choice is up to you haha

Yeah, the tenant id seems like a bigger change than what this is intended for and it just unifies the way tenant id is used across all connection types. Getting this merged, and a publish to crates.io would be wonderful so that we can track against crates.io instead of our fork.

Let me know if there is anything else!

rumqttd/src/server/tls.rs
henil: yeah `tenant_id` is only relevant when compiled with `validate-tenant-prefix` and we use this internally in our platform so i would like to keep it. We can also feature flag this logic so it doesn't run when broker is not compiled with `validate-tenant-prefix`.

> with native tls, it is commented out just like that as well
IIRC this was because at that time native-tls didn't support client side certificates

swanandx:
Hey, can we please this PR only related to allowing TLS connections without client certs ( with changes suggested in review comments )

rumqttd/src/server/tls.rs
nathansizemore: Tenant id coming from cert is now behind `validate-tenant-prefix`.

nathansizemore:
> Hey, can we please this PR only related to allowing TLS connections without client certs ( with changes suggested in review comments )
> 
> as issue #745 is already tackled by PR #743. My apologies for messing up 😬
> 
> Thank you for understanding!

So the impetus for this PR was we need both. This now contains both changes, ready, and the other is still in draft.

swanandx:
It's not just because there is other PR 😅 Separate PRs are helpful as mentioned in this [comment](https://github.com/bytebeamio/rumqtt/pull/748#pullrequestreview-1724192495). 

As these two changes are quite different, merging them in single commit won't be much helpful right. By having em separately will make it easier to undo/revert/reason about in future!

wdyt?

nathansizemore:
> Thanks for the PR!
> 
> > Removes setting the tenant id by client cert
> 
> We do need a way to specify tenant_id though, rather than removing this, we can make it optional? more on it in review comment.
> 
> can we tackle both mentioned issues in different PRs? easier to review / keep track of changes. Anyways, choice is up to you haha

Its not removed. It is feature gated as requested in the previous comment if you look I made a comment saying so. I mean wanting it in separate PRs is up to you. This was free work and free contribution. I'm OK with maintaining a fork. THe code is yours do with what you want, but I won't be contributing or splitting up any more of this work. Feel free to tackle that if you want!

swanandx:
> Its not removed. It is feature gated as requested in the previous comment if you look I made a comment saying so.
>

I wanted to quote the separate PR part only.

> I mean wanting it in separate PRs is up to you. This was free work and free contribution. I'm OK with maintaining a fork
>

I didn't mean to disrespect your/anyone's work/contribution 😕 my apologies if it came out that way!

nathansizemore:
> > Its not removed. It is feature gated as requested in the previous comment if you look I made a comment saying so.
> 
> I wanted to quote the separate PR part only.
> 
> > I mean wanting it in separate PRs is up to you. This was free work and free contribution. I'm OK with maintaining a fork
> 
> I didn't mean to disrespect your/anyone's work/contribution 😕 my apologies if it came out that way!

No disrespect taken, all good here :). I needed something fixed, so instead of just creating an issue, I submitted the changes back to the project. If you want them, feel free to merge them. They work for us, I have a fork, if they get merged cool, if not, cool. We needed rustls without client certs, we now have that. Its yours to do with what you want!

swanandx:
will be closing this PR in favor of #756 & #743 !

Thank you so much for your contribution and pointing out the issues :rocket:

Issue #751, opened 2023-11-12
Rumqttc: Support for SEC1 encoded ECC keys.


ijager:
# Expected Behavior

I would like to use an `"ECC"` key with an ECDSA certificate. I have generated the key with the following openssl command:

```bash
openssl ecparam -name prime256v1 -genkey -noout -out client.key
```

# Current Behavior

rumqtt cannot parse this key and this results in a `Tls(NoValidCertInChain)` error.

# Failure Information (for bugs)

Things go wrong in [rumqttc/src/tls.rs](rumqttc/src/tls.rs#L109L115)

```rust
let read_keys = match &client.1 {
    Key::RSA(k) => rustls_pemfile::rsa_private_keys(&mut BufReader::new(
        Cursor::new(k.clone()),
    )),
    Key::ECC(k) => rustls_pemfile::pkcs8_private_keys(&mut BufReader::new(
        Cursor::new(k.clone()),
    )),
};
```

The `ECC()` variant uses `pkcs8_private_keys()` to parse the key. However my key does not use PKCS#8 encoding, instead it uses `SEC1` encoding, which was specifically designed for Elliptic Curve Cryptography. (PKCS#8 can also encode ECC keys, but it involves converting the keys).

Rustls does support SEC1 encoding via `ec_private_keys()`. Rumqtt could add support by adding a `Key` variant.

Perhaps something like this:

```rust
 let read_keys = match &client.1 {
    Key::RSA(k) => rustls_pemfile::rsa_private_keys(&mut BufReader::new(
        Cursor::new(k.clone()),
    )),
    Key::PKCS(k) => rustls_pemfile::pkcs8_private_keys(&mut BufReader::new(
        Cursor::new(k.clone()),
    )),
    Key::EC(k) => rustls_pemfile:: ec_private_keys(&mut BufReader::new(
        Cursor::new(k.clone()),
    )),
};
```

I changed `ECC` to `PKCS` And added `EC`. This would be a breaking change, so not sure if that is desired. But we could also keep the `Key::ECC(k) => rustls_pemfile::pkcs8_private_keys(` variant for backwards compatibility and make it deprecated.

It would be in line with the three variants that Rustls has for keys: 

```rust
/// A DER-encoded plaintext RSA private key; as specified in PKCS#1/RFC3447
RSAKey(Vec<u8>),

/// A DER-encoded plaintext private key; as specified in PKCS#8/RFC5958
PKCS8Key(Vec<u8>),

[...]

PR #752, opened 2023-11-12
rumqttc: TLS Key encoding fix

ijager:
## Type of change

Breaking Change

## Checklist:

## Description

This PR lets users of rumqttc use any of the Rustls supported key encodings [RSA (PKCS#1), PKCS#8, SEC1].  

This issue addresses #751 

I left the `Key` enum as is to prevent a breaking change, however it does no longer matter if the user choses `RSA` or `ECC`. For a cleaner API I think the `Key` enum could be removed.

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Expose `EventLoop::clean` to allow triggering shutdown and subsequent storage of pending requests
-- Add support for `SEC1` encoded TLS keys (specifically focusing on elliptic curve cryptography (ECC)).
+- Support for all variants of TLS key formats currently supported by Rustls: `PKCS#1`, `PKCS#8`, `RFC5915`. In practice we should now support all RSA keys and ECC keys in `DER` and `SEC1` encoding. Previously only `PKCS#1` and `PKCS#8` where supported.
+- TLS Error variants: `NoValidCACert`, `NoValidKeyInChain`.
 
 ### Changed
 - Synchronous client methods take `&self` instead of `&mut self` (#646)
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Expose `EventLoop::clean` to allow triggering shutdown and subsequent storage of pending requests
 - Support for all variants of TLS key formats currently supported by Rustls: `PKCS#1`, `PKCS#8`, `RFC5915`. In practice we should now support all RSA keys and ECC keys in `DER` and `SEC1` encoding. Previously only `PKCS#1` and `PKCS#8` where supported.
-- TLS Error variants: `NoValidCACert`, `NoValidKeyInChain`.
+- TLS Error variants: `NoValidClientCertInChain`, `NoValidKeyInChain`.
 
 ### Changed
 - Synchronous client methods take `&self` instead of `&mut self` (#646)
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Expose `EventLoop::clean` to allow triggering shutdown and subsequent storage of pending requests
+- Add support for `SEC1` encoded TLS keys (specifically focusing on elliptic curve cryptography (ECC)). The `EC` and `PKCS` variants are added to the `Key` enum the `ECC` variant is deprecated because it was ambiguous.
 
 ### Changed
 - Synchronous client methods take `&self` instead of `&mut self` (#646)
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -9,10 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Expose `EventLoop::clean` to allow triggering shutdown and subsequent storage of pending requests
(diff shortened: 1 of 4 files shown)


ijager:
- Commit 1 implemented this by adding a new variant. 
- Commit 2 uses `rustls_pemfile::read_one` to avoid breaking changes.

ijager:
Tested with 

- [x] RSA (PKCS#1) key for RSA
- [x] PKCS#8 key for RSA
- [x] SEC1 key for ECDSA

ijager:
As discussed in #751 I have removed the `Key` enum. This is documented in the changelog. Examples are also updated.

rumqttc/CHANGELOG.md
```diff
@@ -9,9 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - Expose `EventLoop::clean` to allow triggering shutdown and subsequent storage of pending requests
+- Add support for `SEC1` encoded TLS keys (specifically focusing on elliptic curve cryptography (ECC)).
```
swanandx, line 12: we are adding support for other encoded keys ( supported by rustls pemfile ) as well right? can we make it something like that instead of just saying SEC1?

rumqttc/src/tls.rs
```diff
@@ -106,27 +107,19 @@ pub async fn rustls_connector(tls_config: &TlsConfiguration) -> Result<RustlsCon
...
-                    None => return Err(Error::NoValidCertInChain),
+                // Create buffer for key file
+                let mut key_buffer = BufReader::new(Cursor::new(client.1.clone()));
+                // We only read the first key in the key file
+                let key = match rustls_pemfile::read_one(&mut key_buffer) {
+                    Ok(Some(rustls_pemfile::Item::RSAKey(key)))
+                    | Ok(Some(rustls_pemfile::Item::PKCS8Key(key)))
+                    | Ok(Some(rustls_pemfile::Item::ECKey(key))) => key,
+                    Ok(None) | Ok(Some(_)) => return Err(Error::NoValidKey),
+                    Err(err) => return Err(Error::Io(err)),
```
swanandx, line 119: in case if it is `Some(_)` but not key, we want to continue looking for the key ( because that is how fns like `rsa_private_keys` work!, they ignore the other items ).

```rust
// keep reading Items one by one to find a Key, return error if none found.
let key = loop {
    // note: as read_one returns io::Error, and as our Error::IO(_) has from io::Error
    // i just put ? here. if this is not working, we might have to use .map_err()?
    let item = rustls_pemfile::read_one(key_buffer)?;
    match item {
        Some(Item::ECKey(key) | Item::RSAKey(key) | Item::PKCS8Key(key)) => {
            break key;
        }
        // added that InChain cuz we have it for Certs lol
        None => return Err(Error::NoValidKeyInChain),
        _ => {}
    }
}
```

swanandx (commented):
Thank you for the PR! Looks like `Key` is still there in `lib.rs` ( that is why CI is failing ), please fix it as well : )

PS: This is a breaking change, so can we please update the PR description to clarify it?

rumqttc/src/tls.rs
ijager: Ah good point!

rumqttc/CHANGELOG.md
ijager: I have changed it.

rumqttc/src/tls.rs
ijager: Done

rumqttc/src/tls.rs
```diff
@@ -102,31 +107,31 @@ pub async fn rustls_connector(tls_config: &TlsConfiguration) -> Result<RustlsCon
 
             // Add der encoded client cert and key
             let mut config = if let Some(client) = client_auth.as_ref() {
+
                 let certs =
                     rustls_pemfile::certs(&mut BufReader::new(Cursor::new(client.0.clone())))?;
-                // load appropriate Key as per the user request. The underlying signature algorithm
+                if certs.is_empty() {
+                    return Err(Error::NoValidCertInChain);
```
swanandx, line 114: `NoValidCertInChain` was used to invalid / missing CA certs previously, so instead of creating new variant for `NoValidCACert`, we should create `NoValidClientCert` with `#[error("No valid certificate for client authentication in chain")]` .

If we want, we can just update the desc of `NoValidCertInChain` to say `#[error("No valid CA certificate provided")]`

wdyt?

PS: update changelog accordingly!

rumqttc/src/tls.rs
```diff
@@ -102,31 +107,31 @@ pub async fn rustls_connector(tls_config: &TlsConfiguration) -> Result<RustlsCon
...
+                }
+                // Load appropriate Key as per the user request. The underlying signature algorithm
```
swanandx, line 116: outdated comment, we no longer are loading `Key` as per user request. Rather we are extracting first valid key from file provided by user.

rumqttc/src/tls.rs
ijager: Sounds good, then I will add `NoValidClientCertInChain`

rumqttc/src/tls.rs
ijager: Removed

rumqttc/src/tls.rs
```diff
@@ -51,23 +49,26 @@ pub enum Error {
...
-// // The cert handling functions return unit right now, this is a shortcut
-// impl From<()> for Error {
-//     fn from(_: ()) -> Self {
-//         Error::NoValidCertInChain
-//     }
-// }
-
 #[cfg(feature = "use-rustls")]
 pub async fn rustls_connector(tls_config: &TlsConfiguration) -> Result<RustlsConnector, Error> {
+    use rustls_pemfile::Item;
```
swanandx, line 70: can we please move this import to top ( where other imports are ) to be consistent with imports? Don't forget to feature gate it though haha

PS: I should have added this comment in last review but I forgot, my bad :sweat_smile:

ijager: Ah yes, auto import placed it there. good catch

ijager:
Not sure why that test failed, seems unrelated?

swanandx:
> Not sure why that test failed, seems unrelated?

Yup, it keeps failing randomly 😜

swanandx:
Thank you for the PR!

PR #756, opened 2023-11-21
fix: feature flag rustls client auth

de-sh:
Solves #747 

## Type of change

Bug fix: don't require client authentication for rustls

## Checklist:


rumqttd/Cargo.toml
```diff
@@ -43,7 +43,8 @@ default = ["use-rustls", "websocket"]
 use-rustls = ["dep:tokio-rustls", "dep:rustls-webpki", "dep:rustls-pemfile", "dep:x509-parser"]
 use-native-tls = ["dep:tokio-native-tls", "dep:x509-parser"]
 websocket = ["dep:async-tungstenite", "dep:tokio-util", "dep:futures-util", "dep:ws_stream_tungstenite"]
-validate-tenant-prefix = []
+verify-client-cert = []
```
swanandx, line 46:
```suggestion
verify-client-cert = ["use-rustls"]
```

shall we enable rustls as it is required? technically, it would work without enabling cuz we would never reach a point where verify-client-cert is used without enabling rustls. So not that required.

de-sh: I am keeping this open in the hope that one day nativetls will support mTLS, but yeah, upto you

swanandx: okie, will keep it open then.

PR #758, opened 2023-11-26
optional client certificates while using native-tls

swanandx:
client certificate is optional while using rumqttc with rustls. But with native-tls, it is required to specify certificates! 

Client authentication certificates should be optional for native-tls as well, this PR makes the required changes for it :rocket: 

## Type of change

## Checklist


de-sh (approved):
lgtm!

PR #760, opened 2023-11-29
drain pending Request from channel while cleanup

swanandx:
calling drain on receiver while doing cleanup would help us in preventing data loss to some extend!

## Type of change

## Checklist:


de-sh:
What about v5? https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/v5/eventloop.rs#L126

swanandx:
> What about v5? 
> 

Good catch! will do the changes there as well.

Issue #584, opened 2023-03-01
Advanced authentication methods


galah92:
I'd like to be able to approve/disapprove device connections based on advanced methods rather then simple password validation. For example, it would be great if we can have some type of a hook/middleware/notification saying a device is trying to connect with some credentials (client_id & password), and I'll be able to verify the password with my database. In the later phase, this will allow enforcing devices to send JWT-based password and have some kind of OAuth2-based authentication.

Is something like that currently supported?

Thank you

henil:
We would like to have that feature, but its not something currently in priority.

Nickztar:
This would be a great feature and is something that some other brokers support. Something like how HiveMQ does it: https://www.hivemq.com/extension/file-rbac-extension/

swanandx:
> While not a priority, are you amenable to a PR?

Yup! PRs are always welcome 💯

> Do you have any thoughts on the design?

This wasn't priority so didn't really give it much thought, but definitely would love to discuss and ideate on design!

andrewbaxter:
I don't have any specific ideas about how this would integrate since I'm not familiar with the code base, but here was my general idea:
* Instead of a hash map, config takes an object implementing an interface
* Create a static auth implementation, that gets a hash map
* Config file remains unchanged, but instantiates the static auth implementation
* When authentication is successful, the implementation returns a new object to be associated with the connection
* When receiving messages to a topic on the connection, use the associated object to optionally reject the message
* When about to notify on a subscribed connection, use the associated object to possibly not notify

My only idea on the specifics is that the config would take a `Box<dyn Authenticator>` for the authenticator.  I think the serialized config file struct and the config used by the broker objects themselves will need to be separated (so the broker doesn't store the deserialized config object).

andrewbaxter:
I think splitting the serial structures from the programmatic config structure make sense also by decoupling serialization considerations from server requirements.  I already see there's a couple places where the config layout was dictated by serializer restrictions (i.e. `_` in names).

That said, is there a reason to repeat the authorizer configuration in multiple places in the config?  For example currently v4 and v5 have their own username/password maps.  If I move it to the top level I can get away with just duplicating the top level config structure for now.

andrewbaxter:
Nevermind, I think I can leave auth in the nested structs without copying too much right now.

swanandx:
Hey @galah92 , this issue would be solved with PR #761 !

you can define your fn like:
```rust
[...]

PR #761, opened 2023-12-02
Adds external auth capabilities.

nathansizemore:
## Type of change

## Checklist:

## Notes

* Breaks config API by renaming `auth` to `static_auth`.

This solves two problems we currently have with rumqttd:

* The ability to get client id on new connections, when a new connection happens.
* Allow hooks into our external auth service.

### DIFF
--- a/rumqttd/src/link/remote.rs
+++ b/rumqttd/src/link/remote.rs
@@ -186,7 +186,7 @@ where
         packet => return Err(Error::NotConnectPacket(packet)),
     };
 
-    handle_auth(config.clone(), login.as_ref(), connect.client_id.clone())?;
+    handle_auth(config.clone(), login.as_ref(), &connect.client_id)?;
 
     // When keep_alive feature is disabled client can live forever, which is not good in
     // distributed broker context so currenlty we don't allow it.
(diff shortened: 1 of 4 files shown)


rumqttd/src/link/remote.rs
```diff
@@ -225,3 +210,157 @@ where
...
+fn handle_auth(
+    config: Arc<ConnectionSettings>,
+    login: Option<&Login>,
+    client_id: String,
+) -> Result<(), Error> {
+    if config.static_auth.is_none() && config.dynamic_auth.is_none() {
+        return Ok(());
+    }
+
+    if login.is_none() && (config.static_auth.is_some() || config.dynamic_auth.is_some()) {
```
swanandx, line 223: we already checked that either of auth or dymanic auth is present ( i.e. both aren't None ). so no need to check again!
```suggestion
    if login.is_none()  {
```
PS: please add the comment why login being absent isn't allowed here ( same comments which existed before will work as well :) )

rumqttd/src/link/remote.rs
```diff
@@ -225,3 +210,157 @@ where
...
+
+    if static_auth_verified {
+        return Ok(());
+    }
+
+    if let Some(ref auth) = config.dynamic_auth {
+        if auth(client_id, u.to_owned(), p.to_owned()) {
+            return Ok(());
+        }
+    }
```
swanandx, line 249: some points / questions:
- we should check for either static or dynamic, not both!
- dynamic auth should take preference, if it is present, we don't check for static auth.

so basically, if user adds dynamic auth, the auth from config will get ignored.

rumqttd/src/link/remote.rs
```diff
@@ -225,3 +210,157 @@ where
...
+    let p = login.password.as_str();
+
+    let mut static_auth_verified = false;
+    if let Some(ref pairs) = config.static_auth {
+        for (validu, validp) in pairs {
+            if u == validu && p == validp {
+                static_auth_verified = true;
+                break;
+            }
+        }
```
swanandx, line 238: using iter `any` would be more elegant, eg:
```rs
credentials.iter().any(|(user, pass)| (user, pass) == (&username, &password));
```

swanandx (commented):
Thank you so much for the PR. This should fix #584 !

added some comments, please have a look and feel free to comment :)

swanandx:
PR LGTM :)!

just last question, instead of making `external_auth` public, we can impl helper fn to set external auth ( like we do for reload filters & consolesettings ).:
```diff
@@ -132,11 +132,21 @@ pub struct ConnectionSettings {
     pub max_inflight_count: usize,
     pub auth: Option<HashMap<String, String>>,
     #[serde(skip)]
-    pub external_auth: Option<AuthHandler>,
+    external_auth: Option<AuthHandler>,
     #[serde(default)]
     pub dynamic_filters: bool,
 }

+impl ConnectionSettings {
+    /// Sepcify a function to be used for authentication.
+    ///
+    /// This will take precedence and auth specified in config will
+    /// be ignored.
+    pub fn set_external_auth(&mut self, auth: AuthHandler) {
+        self.external_auth.replace(auth);
+    }
+}
+
```

though, as you are the one of the users who will be using this, lmk which way seems better, Thank you!

nathansizemore:
> just last question, instead of making `external_auth` public, we can impl helper fn to set external auth ( like we do for reload filters & consolesettings ).:

If the method doesn't do anything else besides setting the prop, doesn't seem worth the extra code to me. But if it's some style preference or whatever, feel free to change it up.

swanandx (approved):

swanandx:
> If the method doesn't do anything else besides setting the prop, doesn't seem worth the extra code to me
>

you are right, we can tackle it when we add better way to construct configs without TOML file. :smile:

PR #762, opened 2023-12-04
fix(rumqttc): Lower the MSRV to Rust version 1.65.0

RoastVeg:
This PR lowers the MSRV of rumqttc to rust version 1.65.0, in line with the highest MSRV of its dependencies, which is `addr2line`. Everything still compiles and tests pass ok. rumqttd's MSRV is unchanged.

## Type of change

## Checklist:


swanandx:
I tried [cargo-msrv](https://github.com/foresterre/cargo-msrv) and it is giving:
```
The MSRV is: 1.64.0
```

can you just verify once :thinking:

RoastVeg:
> I tried [cargo-msrv](https://github.com/foresterre/cargo-msrv) and it is giving:
> 
> ```
> The MSRV is: 1.64.0
> ```
> 
> can you just verify once 🤔

I tried with `cargo msrv --min 1.64.0 --max 1.65.0` and it returns:
```
Fetching index
Determining the Minimum Supported Rust Version (MSRV) for toolchain x86_64-unknown-linux-gnu
Using check command cargo check
Check for toolchain '1.65.0-x86_64-unknown-linux-gnu' succeeded

Check for toolchain '1.64.0-x86_64-unknown-linux-gnu' failed with:
┌──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┐│ error: package `rumqttc v0.23.0 (/home/lmh/Documents/foss/rumqtt/rumqttc)` cannot be built because it requires   ││ rustc 1.65.0 or newer, while the currently active rustc version is 1.64.0                                        │└──────────────────────────────────────────────────────────────────────────────────────────────────────────────────┘   Finished The MSRV is: 1.65.0   █████████████████████████████████████████████████████████████████████████ 00:00:14
```
This aligns with the result I got from `cargo msrv list`, which is where I found that `addr2line` has an MSRV of `1.65.0`. Do you have an out of date `Cargo.lock`?

RoastVeg:
It looks like addr2line sets their MSRV to 1.65, but it can be compiled with 1.64. I'm not certain what the best practice is here, but surely rumqttc should follow its dependencies?

swanandx:
lets rely on cargo msrv for this case :)

swanandx (approved):

swanandx:
Thanks for contributing 😁

Issue #775, opened 2023-12-20
rumqttd: disconnect notification does not fire for non-graceful disconnects.


nathansizemore:
# Expected Behavior

When the OS reports a socket is disconnected, the link/broker would report the notification.

# Current Behavior

No event is published signaling the disconnect.

## Context

```toml
id = 0

[router]
id = 0
max_connections = 10010
max_outgoing_packet_count = 200
max_segment_size = 104857600
max_segment_count = 10

[v4.1]
name = "v4-1"
listen = "0.0.0.0:1337"
next_connection_delay_ms = 1
    [v4.1.connections]
    connection_timeout_ms = 60000
    max_payload_size = 20480
    max_inflight_count = 100
    dynamic_filters = true
```

```rust
let mut config = load_cfg();
if let Some(ref mut v4) = config.v4 {
    if let Some(settings) = v4.get_mut("1") {
        settings.connections.external_auth = Some(Arc::new(authorize_client));
    }
}

let mut broker = Broker::new(config);
let (mut link_tx, mut link_rx) = broker.link("singlenode").unwrap();
thread::spawn(move || {
    broker.start().unwrap();
});

link_tx.subscribe("#").unwrap();

loop {
    let notification = match link_rx.recv().unwrap() {
        Some(v) => v,
        None => continue,
    };

    match notification {
        Notification::Forward(forward) => {
            println!(
                "Topic = {:?}, Payload = {} bytes",
                forward.publish.topic,
                forward.publish.payload.len()
            );
        }
        Notification::Disconnect(d, props) => {
            println!("========== disconnect: \n{d:#?}\n{props:#?}");
        }
        v => {
            println!("{v:?}");
        }
    }
}
```

Connect client. Remove internet capabilities from client. Eventually, when the OS reports the socket as disconnected, tracing picks up, but no notification for the disconnect is sent. Only this in logs: 

> ERROR rumqttd::server::broker: Disconnected!!, error: Network(Io(Os { code: 110, kind: TimedOut, message: "Connection timed out" }))

To shorten the default OS timeout to 1sec with 3 attempts, I pulled in libc to test with this:

```rust
let (stream, addr) = match listener.accept().await {
    Ok((s, r)) => {
        let fd = s.as_raw_fd();
        let optval: i32 = 1;
        let optlen = mem::size_of::<i32>() as libc::socklen_t;
        let _ = unsafe {
            libc::setsockopt(
                fd,
                libc::SOL_SOCKET,
[...]

PR #776, opened 2023-12-20
Adds on_disconnect handler to rumqttd.

nathansizemore:
We're using rumqttd against devices that uses v4 protocol and do not always disconnect gracefully. It appears in this circumstance, the broker never reports a disconnect notification as would be expected here:

```rust
link_tx.subscribe("#").unwrap();

loop {
    let notification = match link_rx.recv().unwrap() {
        Some(v) => v,
        None => continue,
    };

    match notification {
        Notification::Forward(forward) => println!("forward"),
        Notification::Disconnect(d, props) => println!("disconnect: \n{d:#?}\n{props:#?}"),
        v => println!("{v:?}"),
    }
}
```

We also start processes based on a client id, so not having access to that information when the client is no longer connected is a pain point this also solves by passing in the id for the disconnected client to the handler.

This is related to #775 .

## Type of change

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1858,7 +1858,6 @@ dependencies = [
  "config",
  "flume",
  "futures-util",
- "libc",
  "metrics",
  "metrics-exporter-prometheus",
  "parking_lot",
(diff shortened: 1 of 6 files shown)


nathansizemore:
closing in favor of another PR will impl a little differently.

PR #777, opened 2023-12-20
Adds handler to set socket option in rumqttd.

nathansizemore:
This is useful for setting lower level settings on the socket, such as `SO_KEEPALIVE`, `TCP_KEEPCNT`, and various socket level flags. 

A use case we have is currently when a client is connect, the network drops on the client end, rumqttd will not show as disconnected until the TCP keep alive settings from the OS kick in, this defaults to 2 hours on Linux. This allows a spot so the connections that need specific TCP conditions can have that.

This adds extra configs, which reduce zombie connections that can exist. 

ref #775 

## Type of change

## Checklist:

### DIFF
--- a/rumqttd/CHANGELOG.md
+++ b/rumqttd/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* `set_socket_opts` in `ConnectionSettings`.
+
 ### Changed
 
 ### Deprecated
--- a/rumqttd/src/lib.rs
+++ b/rumqttd/src/lib.rs
@@ -5,6 +5,7 @@ use std::sync::Arc;
 use std::{collections::HashMap, path::Path};
 
 use serde::{Deserialize, Serialize};
+use tokio::net::TcpStream;
 use tracing_subscriber::{
     filter::EnvFilter,
     fmt::{
@@ -44,6 +45,7 @@ pub type ClientId = String;
 pub type AuthUser = String;
 pub type AuthPass = String;
 pub type AuthHandler = Arc<dyn Fn(ClientId, AuthUser, AuthPass) -> bool + Send + Sync + 'static>;
+pub type SocketOptHandler = Arc<dyn Fn(&mut TcpStream) + Send + Sync + 'static>;
 
 #[derive(Debug, Default, Serialize, Deserialize, Clone)]
 pub struct Config {
@@ -135,6 +137,8 @@ pub struct ConnectionSettings {
     pub external_auth: Option<AuthHandler>,
     #[serde(default)]
     pub dynamic_filters: bool,
+    #[serde(skip)]
+    pub set_socket_opts: Option<SocketOptHandler>,
 }
 
 impl fmt::Debug for ConnectionSettings {
@@ -146,6 +150,7 @@ impl fmt::Debug for ConnectionSettings {
             .field("auth", &self.auth)
             .field("external_auth", &self.external_auth.is_some())
             .field("dynamic_filters", &self.dynamic_filters)
+            .field("set_socket_opts", &self.set_socket_opts.is_some())
             .finish()
     }
 }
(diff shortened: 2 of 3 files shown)


swanandx:
Hey, thanks for the PR!

> This is useful for setting lower level settings on the socket, such as SO_KEEPALIVE, TCP_KEEPCNT, and various socket level flags.
>
`std` as well as `tokio`'s ( which we use ) `TcpStream` doesn't have fn to set `SO_KEEPALIVE` :( ( check discussions: https://github.com/rust-lang/rust/issues/69774 )

if we want to set those flags, we would need to pull in additional dependency like `socket2` or use unsafe libc as you did in #775  

I think configurations like this, which are for fine tuning, are better done / set at OS level, e.g. using `sysctl` on linux, instead of in broker itself. wdyt?

cc: @henil

nathansizemore:
np, I'm sure you'll be getting more PRs from us when various needs come up :)

>I think configurations like this, which are for fine tuning, are better done / set at OS level, e.g. using sysctl on linux, instead of in broker itself. wdyt?

This approach only works when the OS is _only_ a broker. It breaks once the OS is used for varying network operations. Say the broker is running on a machine, with socket settings tuned for the MQTT clients over TCP. Now say the that machine also makes some HTTP call for an auth service, or a db connection, now those sockets also inherit the settings from the OS, which you only wanted to apply to connections within MQTT.

Ideally, this is even more configurable in the future with passing in client id, but I hadn't found a way to do that without more changes, so for the time this is the first steps that gets us closer to where we need to be.

We have several client types that connect, but one particular model (derived by client id) is prone to network failure. This allows us to detect quicker when the connection has dropped.

swanandx:
> We have several client types that connect, but one particular model (derived by client id) is prone to network failure. This allows us to detect quicker when the connection has dropped.
>
Interesting :thinking: 

> A use case we have is currently when a client is connect, the network drops on the client end, rumqttd will not show as disconnected until the TCP keep alive settings from the OS kick in,
>
can you elaborate more on wdym by OS keep alive to kick in? for me, I get this log as soon as I close the connection, which should imply that we are knowing that client is disconnected right? 
```
 ERROR rumqttd::server::broker: Disconnected!!, error: Network(Io(Custom { kind: ConnectionAborted, error: "connection closed by peer" })) 
```

swanandx:
> I'm sure you'll be getting more PRs from us when various needs come up :)
>
That's awesome :rocket: we will try our best to keep up with those needs :)

swanandx:
oh understood! thanks for such detailed explanation :100: 

> You can think of them as MQTT pings, but at the TCP level
>
just curious, if we set MQTT keep-alive value to something low, then this should work as well right? it would know when client isn't sending pings. ( though it would increase number of pings overall :eyes: )

nathansizemore:
> just curious, if we set MQTT keep-alive value to something low, then this should work as well right? it would know when client isn't sending pings. ( though it would increase number of pings overall 👀 )

Right, this also works. I believe right now that is set from something in the client connect packet? (Not super versed in mqtt). Or a spot in the config where this could be forced instead of reading from whatever the client sends, if that is indeed how it currently works.

swanandx:
> I believe right now that is set from something in the client connect packet?
>
Yup, keep alive time is set while connecting in CONNECT packet.

If broker doesn't get any packet from client within 1.5 times the keep alive time, it assumes that client is disconnected.

>  a spot in the config where this could be forced instead of reading from whatever the client sends
>
If I'm not wrong, currently there isn't a way to force keepalive via config😅

nathansizemore:
> If I'm not wrong, currently there isn't a way to force keepalive via config😅

Hence, this PR :)

nathansizemore:
> > Hence, this PR :)
> 
> I was talking about MQTT Keepalive, this PR is for low level socket's SO_KEEPALIVE which are quite different 😅

Right. I was just trying to say this PR exists because we need a mechanism that is configurable, on a connection basis, that allows us to detect faster when a non-graceful shutdown has occurred. Both adjust the socket level flags and forcing the broker to override mqtt ping time would solve the issue. Having the ability to set lower level socket flags is less important atm than being able to quickly detect shutdowns, to us at least.

PR #779, opened 2023-12-29
Add `from_senders` for the Sync Client

BKDaugherty:
A lot of my team's code uses the synchronous client. I'd like to be able to easily write tests using a Fake Mqtt client. One easy way to do this for unit tests is by using something like the `from_senders` method that's available on the `AsyncClient`. This PR adds a `from_sender` method for the sync client which can be used to write tests like the one I wrote in this PR. 

It also updates a stale comment on the async client.

## Type of change

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - TLS Error variants: `NoValidClientCertInChain`, `NoValidKeyInChain`.
 - Drain `Request`s, which weren't received by eventloop, from channel and put them in pending while doing cleanup to prevent data loss.
 - websocket request modifier for v4 client
+- Surfaced `AsyncClient`'s `from_senders` method to the `Client` as `from_sender`
 
 ### Changed
 - Synchronous client methods take `&self` instead of `&mut self` (#646)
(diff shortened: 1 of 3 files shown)


swanandx (changes requested):
Thanks for the PR and your patience! 

Can you please add the same changes for `v5/client` as well? so we can stay consistent!

BKDaugherty:
Yeah absolutely!

swanandx:
any updates @BKDaugherty ?

BKDaugherty:
> any updates @BKDaugherty ?

Hey! Sorry @swanandx, I was AFK a bit for the holidays. Just updated to add this to the v5 client as well! Is there a proposal at all for unifying these? Would be curious to see that

swanandx (approved):

swanandx:
> I was AFK a bit for the holidays. Just updated to add this to the v5 client as well!
>

Oh no worries, Thank you so much!

> Is there a proposal at all for unifying these? Would be curious to see that
>
I was thinking about it, but then how would be distinguish between v4 & v5 clients? we might need to store a field like `protocol` in client and choose `publish` / `subscribe` based on it, otherwise introduce generics or something. I thought it might add much more complexity, if you think that isn't the case, feel free to open a RFC, would definitely love to see it :)

 > Hm. Any idea why the macos tests would fail? Doesn't seem related to this
 >
 i don't have that idea either haha, it keeps failing randomly. so we can ignore it

PR #780, opened 2023-12-31
fix: ensure pending packets are not lost on `EventLoop::clean`

de-sh:
Situation earlier:
1. `EventLoop` has pending packets yet to be processed.
2. `EventLoop::clean` is called.
3. `EventLoop::poll` has lost some packets which were earlier pending.

This can be solved with a queue in place of the pending iter.

## Checklist:

- [x] Make an entry to `CHANGELOG.md` if it's relevant to the users of the library. If it's not relevant mention why. (Do we need to do this when the release hasn't been made yet?)

### DIFF
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -269,11 +269,11 @@ impl EventLoop {
         rx: &Receiver<Request>,
         pending_throttle: Duration,
     ) -> Result<Request, ConnectionError> {
-        if pending.len() > 0 {
+        if let Some(req) = pending.pop() {
             time::sleep(pending_throttle).await;
             // We must call .next() AFTER sleep() otherwise .next() would
             // advance the iterator but the future might be canceled before return
-            Ok(pending.next().unwrap())
+            Ok(req)
         } else {
             match rx.recv_async().await {
                 Ok(r) => Ok(r),
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -13,7 +13,6 @@ use std::io;
 use std::net::SocketAddr;
 use std::pin::Pin;
 use std::time::Duration;
-use std::vec::IntoIter;
 
 #[cfg(unix)]
 use {std::path::Path, tokio::net::UnixStream};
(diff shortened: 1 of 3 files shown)


rumqttc/src/eventloop.rs
```diff
@@ -127,14 +126,13 @@ impl EventLoop {
     pub fn clean(&mut self) {
         self.network = None;
         self.keepalive_timeout = None;
-        let mut pending = self.state.clean();
+        self.pending.extend(self.state.clean());
```
swanandx, line 129: calling `.extend()` will imply that previous elements present in pending will persist across calls to `clean()` right? is this the expected behavior desired?

asking because previously we were just updating `self.pending` ( thus, discarding the elements stored in it previously )

de-sh: > calling .extend() will imply that previous elements present in pending will persist across calls to clean() right?
Yes, we are consuming the queue of pending in fifo order, so this is taken care of by `EventLoop::poll` on system restart, the trouble why this edit had to be made is, consider the code:
```
use tokio::{task, time};

use rumqttc::{self, AsyncClient, Event, MqttOptions, Outgoing, QoS};
use std::error::Error;
use std::time::Duration;

#[tokio::main(worker_threads = 1)]
async fn main() -> Result<(), Box<dyn Error>> {
    pretty_env_logger::init();
    // color_backtrace::install();

    let mut mqttoptions = MqttOptions::new("test-1", "localhost", 1883);
    mqttoptions.set_keep_alive(Duration::from_secs(5));

    task::spawn(async move {
        client
            .subscribe("hello/world", QoS::AtMostOnce)
            .await
            .unwrap();

        for i in 1..=10 {
            client
                .publish("hello/world", QoS::AtLeastOnce, false, vec![1; i])
                .await
                .unwrap();

            time::sleep(Duration::from_secs(1)).await;
        }
    });

    loop {
        let event = eventloop.poll().await;
        match &event {
            Ok(Event::Outgoing(Outgoing::Publish(i))) if i > &5 => {
                eventloop.clean();
                dbg!(&eventloop.pending);
            }
            Ok(v) => {
                println!("Event = {v:?}");
            }
            Err(e) => {
                println!("Error = {e:?}");
                return Ok(());
            }
        }
    }
}
```

Run the code on previous and current rumqttc to understand the effect of this change

PR #781, opened 2024-01-06
rumqttd: Changes peer initiated disconnect logs to info.

nathansizemore:
* Disconnect by peer shouldn't be an error. Peer initiated disconnects should be logged as good faith operations.

## Type of change

## Checklist:

### DIFF
--- a/rumqttd/CHANGELOG.md
+++ b/rumqttd/CHANGELOG.md
@@ -11,7 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Public re-export `Strategy` for shared subscriptions
-- Log level of link disconnect by peer.
+- Client id added to disconnect logs.
+- Peer initiated disconnects logged as info rather than error.
 
 ### Deprecated
 
--- a/rumqttd/CHANGELOG.md
+++ b/rumqttd/CHANGELOG.md
@@ -11,7 +11,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Public re-export `Strategy` for shared subscriptions
-- Client id added to disconnect logs.
 - Peer initiated disconnects logged as info rather than error.
 
 ### Deprecated
--- a/rumqttd/CHANGELOG.md
+++ b/rumqttd/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Public re-export `Strategy` for shared subscriptions
+_ Log level of link disconnect by peer.
 
 ### Deprecated
 
--- a/rumqttd/CHANGELOG.md
+++ b/rumqttd/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Public re-export `Strategy` for shared subscriptions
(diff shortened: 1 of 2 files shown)


rumqttd/src/server/broker.rs
```diff
@@ -572,9 +572,26 @@ async fn remote<P: Protocol>(
...
+                    if e.kind() == io::ErrorKind::ConnectionAborted {
+                        info!("Link disconnected");
+                    } else {
+                        error!(error=?e, "Disconnected!!");
+                    }
+                }
+                _ => error!(error=?e, "Disconnected!!"),
+            },
+            _ => error!(error=?e, "Disconnected!!"),
+        },
```
swanandx, line 594: we can avoid nested match, with help of match guards:

```rust
        // connection was closed by peer
        Err(remote::Error::Network(network::Error::Io(err)) | remote::Error::Io(err))
            if err.kind() == io::ErrorKind::ConnectionAborted =>
        {
            info!(error=?err, "Disconnected");
        }
        // Any other error
        Err(e) => {
            error!(error=?e, "Disconnected!!");
        }
```

rumqttd/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - Public re-export `Strategy` for shared subscriptions
+- Log level of link disconnect by peer.
```
swanandx, line 14: nit: we can be more specific about the change

swanandx (commented):
Thanks for the PR!

are there any other type of errors which should be fine if logged with `info!` or shall all disconnects be logged at `info!`?

nathansizemore:
> Thanks for the PR!
> 
> are there any other type of errors which should be fine if logged with `info!` or shall all disconnects be logged at `info!`?

Not that I am aware of atm. As they arise or I notice from our log stuff, I'll send new PRs.

rumqttd/src/server/broker.rs
```diff
@@ -568,12 +568,18 @@ async fn remote<P: Protocol>(
         // No need to send a disconnect message when disconnection
         // originated internally in the router.
         Err(remote::Error::Link(e)) => {
-            error!(error=?e, "router-drop");
```
swanandx, line 571: we don't need to add client_id like this, as it is recorded by the span.
previously this was there: 
```rust
Span::current().record("client_id", &client_id);
```
but it was removed in #566 . I'm unsure of the reason, but will check with @mnpw once and get back to you!

ps:  let's keep this PR only related to `ConnectionAbort` related change, so we can unblock, plus it is always better to tackle one particular thing in PR ( makes it easier to track down in future ). So can you please remove the `client_id` related changes ( here + in changelog ) so we can merge it!

nathansizemore: Sure.

rumqttd/src/server/broker.rs
```diff
@@ -571,9 +571,15 @@ async fn remote<P: Protocol>(
...
+        // Connection was closed by peer
+        Err(remote::Error::Network(network::Error::Io(err)) | remote::Error::Io(err))
+            if err.kind() == io::ErrorKind::ConnectionAborted =>
+        {
+            info!(error=?err, "disconnected");
+        }
         // Any other error
         Err(e) => {
-            error!(error=?e, "Disconnected!!");
+            error!(error=?e, "disconencted");
```
swanandx, line 582:
```suggestion
            error!(error=?e, "disconnected");
```

swanandx (approved):

PR #784, opened 2024-01-08
fix(rumqttd): use keepalive duration when reading from network

swanandx:
MQTT keepalive duration set by the client wasn't being used. This PR aims to fix this issue :)

## Type of change

## Checklist:


wvffle (approved):
Works like a charm! Thanks!

henil (approved):

PR #788, opened 2024-01-24
feat(rumqttc): optimize topic and filter validation

swanandx:
This PR aims to:
- optimize the filter validation by avoiding allocation during validation
- optimize the topic validation by using binary operators instead of multiple conditional statements

## Type of change

## Checklist:

This is just optimization and doesn't affect user API, so CHANGELOG entry isn't required.

### DIFF
--- a/rumqttc/src/mqttbytes/topic.rs
+++ b/rumqttc/src/mqttbytes/topic.rs
@@ -6,7 +6,11 @@ pub fn has_wildcards(s: &str) -> bool {
 /// Checks if a topic is valid
 pub fn valid_topic(topic: &str) -> bool {
     // topic can't contain wildcards
-    !topic.contains('+') && !topic.contains('#')
+    if topic.contains('+') || topic.contains('#') {
+        return false;
+    }
+
+    true
 }
 
 /// Checks if the filter is valid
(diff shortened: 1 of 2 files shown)


de-sh (approved):
Mostly LGTM, would love to get an error variant that better explains the problem encountered if possible, but that should be a separate good first issue

rumqttc/src/mqttbytes/topic.rs
```diff
@@ -5,15 +5,8 @@ pub fn has_wildcards(s: &str) -> bool {
...
-        return false;
-    }
-
-    if topic.contains('#') {
-        return false;
-    }
-
-    true
+    // topic can't contain wildcards
+    !topic.contains('+') && !topic.contains('#')
```
henil, line 9:
```suggestion
    if topic.contains('+') || topic.contains('#') {
        return false;
    }
   
    true
```
i feel this is more readable

swanandx: i wanted to avoid use of conditional statements as shown about and fn calls like:
```rust
!has_wildcard(topic)
```

but guess i'm just over optimizing haha, do you think fn call to `has_wildcard` be better? ( note: it might add additional fn symbols and overhead ). otherwise, I will do the suggested changes!

swanandx: update: lets avoid that fn call for now, pushed the suggested changed :+1:

henil (approved):

PR #790, opened 2024-01-29
chore(rumqttc): update to latest versions of rustls and friends

qwandor:
## Type of change

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -25,6 +25,10 @@ To update your code simply remove `Key::ECC()` or `Key::RSA()` from the initiali
 - Make v5 `RetainForwardRule` public, in order to allow setting it when constructing `Filter` values.
 - Use `VecDeque` instead of `IntoIter` to fix unintentional drop of pending requests on `EventLoop::clean` (#780)
 - `StateError::IncommingPacketTooLarge` is now `StateError::IncomingPacketTooLarge`.
+- Update `tokio-rustls` to `0.25.0`, `rustls-native-certs` to `0.7.0`, `rustls-webpki` to `0.102.1`,
+  `rusttls-pemfile` to `2.0.0`, `async-tungstenite` to `0.24.0`, `ws_stream_tungstenite` to `0.12.0`
+  and `http` to `1.0.0`. This is a breaking change as types from some of these crates are part of
+  the public API.
 
 ### Deprecated
 
--- a/rumqttc/examples/tls.rs
+++ b/rumqttc/examples/tls.rs
@@ -15,12 +15,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
     // Use rustls-native-certs to load root certificates from the operating system.
     let mut root_cert_store = tokio_rustls::rustls::RootCertStore::empty();
-    for cert in rustls_native_certs::load_native_certs().expect("could not load platform certs") {
-        root_cert_store.add(&tokio_rustls::rustls::Certificate(cert.0))?;
-    }
+    root_cert_store.add_parsable_certificates(
+        rustls_native_certs::load_native_certs().expect("could not load platform certs"),
+    );
 
     let client_config = ClientConfig::builder()
-        .with_safe_defaults()
         .with_root_certificates(root_cert_store)
         .with_no_client_auth();
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -182,10 +182,25 @@ dependencies = [
  "futures-util",
  "log",
  "pin-project-lite",
- "rustls-native-certs",
  "tokio",
- "tokio-rustls 0.24.1",
- "tungstenite",
+ "tungstenite 0.20.1",
+]
+
+[[package]]
+name = "async-tungstenite"
+version = "0.24.0"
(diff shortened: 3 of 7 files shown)


qwandor:
This is blocked on `ws_stream_tungstenite` updating to the latest version of `async-tungstenite`.

qwandor:
Sent https://github.com/najamelan/ws_stream_tungstenite/pull/15 to update `ws_stream_tungstenite`. Once that's released this PR should be ready to go.

swanandx:
Hey @qwandor , the PR to update `ws_stream_tungstenite` is merged! so we should be unblocked right?

can we please break down this PR into two separate PRs ( for rumqttc & rumqttd respectively )? would be better for review & merge. 

also, regarding some changes like removing trust_anchors, it would be helpful if you mention why this change was made in PR description!

Thanks :)

qwandor:
> Hey @qwandor , the PR to update `ws_stream_tungstenite` is merged! so we should be unblocked right?
Yep, updated it now.

> can we please break down this PR into two separate PRs ( for rumqttc & rumqttd respectively )? would be better for review & merge.
Done, though you'll probably run into merge conflicts with `Cargo.lock`.

> also, regarding some changes like removing trust_anchors, it would be helpful if you mention why this change was made in PR description!
The method was renamed and does more of the work that we had to do manually before.

swanandx:
> Done, though you'll probably run into merge conflicts with Cargo.lock.
>
oh didn't thought about it tbh, as I never encountered em so for :sweat_smile: . Reason for asking to separate em out is for maintainability.

swanandx:
will review this PR in depth once we are done with #795 :)

rumqttc/examples/tls.rs
```diff
@@ -15,12 +15,11 @@ async fn main() -> Result<(), Box<dyn Error>> {
 
     // Use rustls-native-certs to load root certificates from the operating system.
     let mut root_cert_store = tokio_rustls::rustls::RootCertStore::empty();
-    for cert in rustls_native_certs::load_native_certs().expect("could not load platform certs") {
-        root_cert_store.add(&tokio_rustls::rustls::Certificate(cert.0))?;
-    }
+    root_cert_store.add_parsable_certificates(
+        rustls_native_certs::load_native_certs().expect("could not load platform certs"),
+    );
```
swanandx, line 20: using `add(_)` to add individual cert give us better control, as `add_parsable_certificates` might ignore some certs. 

as this is just an _example_, it doesn't matter tbh, so feel free to mark this as resolved if you wish :)

rumqttc/CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ To update your code simply remove `Key::ECC()` or `Key::RSA()` from the initiali
 - Make v5 `RetainForwardRule` public, in order to allow setting it when constructing `Filter` values.
 - Use `VecDeque` instead of `IntoIter` to fix unintentional drop of pending requests on `EventLoop::clean` (#780)
 - `StateError::IncommingPacketTooLarge` is now `StateError::IncomingPacketTooLarge`.
+- Update `tokio-rustls` to `0.25.0`, `rustls-native-certs` to `0.7.0`, `rustls-webpki` to `0.102.1`,
+  `rusttls-pemfile` to `2.0.0`, `async-tungstenite` to `0.24.0`, `ws_stream_tungstenite` to `0.12.0`
+  and `http` to `1.0.0`. This is a breaking change as types from some of these crates are part of
```
swanandx, line 28: this isn't a breaking change right? as none of the functionality / API exposed to users is changing? :eyes:

swanandx (approved):
Thanks for the PR! LGTM :+1:

swanandx:
you were right about the merge conflicts in `Cargo.lock`! what would be the best way to fix them? I guess we can just sync up this with `main` and rebuilding `Cargo.lock` should fix it.

rumqttc/CHANGELOG.md
qwandor: The `rustls` types are part of `rumqttc`'s public API so this is a breaking change. Clients which depend on `rustls` 0.21 and `rustls-native-certs` 0.6 and call `rumqttc` methods using their types will break if they update `rumqttc` without updating the other dependencies to match.

rumqttc/examples/tls.rs
qwandor: I think ignoring invalid certificates is the right thing to do, rather than crashing just because someone happens to have some root certificates in their system certificate store which we can't parse. They probably don't have control over the certificates provided by the OS anyway, and skipping some doesn't make things any less secure.

qwandor:
I've rebased and fixed the conflicts in Cargo.lock.

rumqttc/examples/tls.rs
swanandx: we were simply calling `.unwrap()` [here](https://github.com/bytebeamio/rumqtt/pull/790/files#diff-a7c14b22199b25375255ee5a3d137a7f77ad9fdde94ff211cd27c6441f3cbd1dR369) in `lib.rs`, hence I asked.

qwandor: Good point, that should probably change too.

PR #791, opened 2024-02-02
feat(rumqttc): allow empty client id

### LINKED ISSUE #667: Client ID should be optional
# Expected Behavior

When creating a client, using an empty client ID should work.

Per the 3.1.1 spec http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html

```
A Server MAY allow a Client to supply a ClientId that has a length of zero bytes, however if it does so the Server MUST treat this as a special case and assign a unique ClientId to that Client. It MUST then process the CONNECT packet as if the Client had provided that unique ClientId [MQTT-3.1.3-6].
```

# Current Behavior

Using an empty client id for the rumqttc client options results in a panic.

```
        let id = id.into();
        if id.starts_with(' ') || id.is_empty() {
            panic!("Invalid client id");
        }
```

arunanshub:
Fixes #667 

## Type of change

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - `MqttOptions::new` now accepts empty client id.
+- `MqttOptions::set_clean_session` now panics if client ID is empty and `clean_session` flag is set to false.
 - Synchronous client methods take `&self` instead of `&mut self` (#646)
 - Removed the `Key` enum: users do not need to specify the TLS key variant in the `TlsConfiguration` anymore, this is inferred automatically.
 To update your code simply remove `Key::ECC()` or `Key::RSA()` from the initialization.
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -8,7 +8,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
-- `MqttOptions::try_set_clean_session` to check if `clean_session` is `true` when `client_id` is an empty string.
 - Expose `EventLoop::clean` to allow triggering shutdown and subsequent storage of pending requests
 - Support for all variants of TLS key formats currently supported by Rustls: `PKCS#1`, `PKCS#8`, `RFC5915`. In practice we should now support all RSA keys and ECC keys in `DER` and `SEC1` encoding. Previously only `PKCS#1` and `PKCS#8` where supported.
 - TLS Error variants: `NoValidClientCertInChain`, `NoValidKeyInChain`.
@@ -17,7 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Surfaced `AsyncClient`'s `from_senders` method to the `Client` as `from_sender`
 
 ### Changed
-- `MqttOptions::new` does not panic if the `client_id` is an empty string or starts with a whitespace.
+- `MqttOptions::new` now accepts empty client id.
 - Synchronous client methods take `&self` instead of `&mut self` (#646)
 - Removed the `Key` enum: users do not need to specify the TLS key variant in the `TlsConfiguration` anymore, this is inferred automatically.
 To update your code simply remove `Key::ECC()` or `Key::RSA()` from the initialization.
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `MqttOptions::try_set_clean_session` to check if `clean_session` is `true` when `client_id` is an empty string.
 - Expose `EventLoop::clean` to allow triggering shutdown and subsequent storage of pending requests
 - Support for all variants of TLS key formats currently supported by Rustls: `PKCS#1`, `PKCS#8`, `RFC5915`. In practice we should now support all RSA keys and ECC keys in `DER` and `SEC1` encoding. Previously only `PKCS#1` and `PKCS#8` where supported.
 - TLS Error variants: `NoValidClientCertInChain`, `NoValidKeyInChain`.
@@ -16,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Surfaced `AsyncClient`'s `from_senders` method to the `Client` as `from_sender`
 
 ### Changed
+- `MqttOptions::new` does not panic if the `client_id` is an empty string or starts with a whitespace.
 - Synchronous client methods take `&self` instead of `&mut self` (#646)
 - Removed the `Key` enum: users do not need to specify the TLS key variant in the `TlsConfiguration` anymore, this is inferred automatically.
 To update your code simply remove `Key::ECC()` or `Key::RSA()` from the initialization.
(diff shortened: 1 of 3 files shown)


rumqttc/src/lib.rs
```diff
@@ -624,11 +613,44 @@ impl MqttOptions {
...
     /// happens. Local queue state is also held to retransmit packets after reconnection.
+    ///
+    /// NOTE: It panicks if the `client_id` is a zero byte string and `clean_session` is
+    /// not `true`.
+    ///
+    /// ```should_panic
+    /// # use rumqttc::MqttOptions;
+    /// let mut options = MqttOptions::new("", "localhost", 1883);
+    /// options.set_clean_session(false);
+    /// ```
```
swanandx, line 624: panic related info is added in `Panic` sections generally.

```suggestion
    /// # Panic
    /// 
    /// Panics if `clean_session` is false when `client_id` is empty.
    /// 
    /// ```should_panic
    /// # use rumqttc::MqttOptions;
    /// let mut options = MqttOptions::new("", "localhost", 1883);
    /// options.set_clean_session(false);
    /// ```
```

rumqttc/src/lib.rs
```diff
@@ -624,11 +613,44 @@ impl MqttOptions {
...
     pub fn set_clean_session(&mut self, clean_session: bool) -> &mut Self {
+        if self.client_id.is_empty() {
+            assert!(
+                clean_session,
+                "Cannot unset clean session when client id is empty"
+            );
+        }
```
swanandx, line 631: assert message can be improved like: "Clean session must be true when client id is provided"

btw, we can use single assert statement instead of two checks:
```rust
assert!(!client_id.is_empty() || clean_session, "Clean session must be true when client id is empty")
```

rumqttc/src/lib.rs
```diff
@@ -624,11 +613,44 @@ impl MqttOptions {
...
+    /// assert_eq!(error, OptionError::CleanSession);
+    /// ```
+    #[cfg(feature = "url")]
+    pub fn try_set_clean_session(&mut self, clean_session: bool) -> Result<&mut Self, OptionError> {
+        if self.client_id.is_empty() && !clean_session {
+            return Err(OptionError::CleanSession);
+        }
+        self.clean_session = clean_session;
+        Ok(self)
+    }
```
swanandx, line 652: `url` feature is only used for `from_url` fn, this fn isn't required here! 

we should add this check in url's try_from impl block!

note: there is PR to refactor MqttOptions which would remove this url feature #789

rumqttc/src/lib.rs
```diff
@@ -1006,10 +1023,4 @@ mod test {
             OptionError::Inflight
         );
     }
-
-    #[test]
-    #[should_panic]
```
swanandx, line 1011: instead of removing whole test, we should have just removed this `#[should_panic]` attribute.

rumqttc/src/v5/mod.rs
```diff
@@ -113,13 +113,13 @@ impl MqttOptions {
     /// - port: The port number on which broker must be listening for incoming connections
     ///
     /// ```
-    /// # use rumqttc::MqttOptions;
+    /// # use rumqttc::v5::MqttOptions;
     /// let options = MqttOptions::new("123", "localhost", 1883);
     /// ```
     /// NOTE: you are not allowed to use an id that starts with a whitespace or is empty.
     /// for example, the following code would panic:
     /// ```should_panic
```
swanandx, line 121: this shouldn't panic, that's the aim of PR right? :laughing:

rumqttc/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `MqttOptions::try_set_clean_session` to check if `clean_session` is `true` when `client_id` is an empty string.
```
swanandx, line 11: as suggested in other comment, we should remove this.

swanandx (changes requested):
Thanks for the PR! I have added some review comments, plz have a look once :)

Other than that:
- please do same changes in v5's MqttOptions!
- can we add test to verify we will only panic if client id is empty && clean session is false

Rest LGTM! :rocket: Nice work!

rumqttc/src/lib.rs
```diff
@@ -624,7 +613,21 @@ impl MqttOptions {
...
+    /// ```should_panic
+    /// # use rumqttc::MqttOptions;
+    /// let mut options = MqttOptions::new("", "localhost", 1883);
+    /// options.set_clean_session(false);
+    /// ```
     pub fn set_clean_session(&mut self, clean_session: bool) -> &mut Self {
+        assert!(
+            self.client_id.is_empty() && clean_session,
+            "Cannot unset clean session when client id is empty"
+        );
```
swanandx, line 630: this condition is incorrect! as this will panic even when client id isn't empty!

please consider using assert condition provided in review comment: https://github.com/bytebeamio/rumqtt/pull/791#discussion_r1475840369

you can verify it while adding tests :)

rumqttc/src/v5/mod.rs
swanandx, line 1: we aren't checking empty client id in `set_clean_session` function in v5! please update it.

rumqttc/src/v5/mod.rs
```diff
@@ -821,8 +804,7 @@ mod test {
     }
 
     #[test]
-    #[should_panic]
-    fn no_client_id() {
+    fn allow_no_client_id() {
```
swanandx, line 807:
```suggestion
    fn allow_empty_client_id() {
```

rumqttc/CHANGELOG.md
swanandx, line 1: please consider adding the changed behavior of `set_clean_session` as well.

swanandx (changes requested):
can we please add some tests to verify `set_clean_session` is working correctly?

thanks :)

rumqttc/src/v5/mod.rs
swanandx: in v5, under session expiry interval:

> Setting Clean Start to 1 and a Session Expiry Interval of 0, is equivalent to setting CleanSession to 1 in the MQTT Specification Version 3.1.1. Setting Clean Start to 0 and no Session Expiry Interval, is equivalent to setting CleanSession to 0 in the MQTT Specification Version 3.1.1. 

so basically, when the clean start flag is false along with no session expiry interval, then only it is equivalent to _clean session_ as MQTTv3.1.1.

i tried with `mosquitto_pub` client, with clean start false and no ( infinite ) session expiry in v5, and it says "Error: You must provide a client id if you are using an infinite session expiry interval."

( tried paho python client as well, but ig they generate random client id anyways ).

checking both session expiry and clean start won't be possible here, as we can't determine the order they will be set in by users. So i think it should be fine for now. wdyt?

swanandx:
Hey @arunanshub , any updates!?

arunanshub:
Working on it. Upon referencing Mosquitto lib, it has been difficult to
find where they are validating the "clean_start", "client_id" and save
session options. It seems that a single codebase provides support for v3
and v5. v5 is enabled by flag `CLIENT_RR` (MQTT v5 request response), and
they are using a lot of conditional compilation. Will keep you updated.

swanandx:
> it has been difficult to find where they are validating the "clean_start", "client_id" and save session options
>

as discussed already, they ( or other clients ) don't validate it!

This PR is blocked by adding tests to check if are using assert correctly and other changes suggested in review.

Can you please prioritise em so we can close the PR? Then you can continue your research and open follow up PR ( if required ) 😁

swanandx (approved):

PR #792, opened 2024-02-03
fix(rumqttd): fix session present flag in connack

swanandx:
we were setting `session_present` flag as `true` even when there was no previous session stored ( due to rumqttd storing matrices ). 

This PR aims to fix it by making session state as Option, so we can distinguish between just storing matrices and storing the session state. 

## Type of change

## Checklist:

### DIFF
--- a/rumqttd/src/router/routing.rs
+++ b/rumqttd/src/router/routing.rs
@@ -307,26 +307,26 @@ impl Router {
         // for qos2 pending pubrels
         let mut pending_acks = VecDeque::new();
 
-        let tracker = 'tracker: {
-            if !clean_session {
-                let Some(saved) = saved else {
-                    break 'tracker Tracker::new(client_id.clone());
-                };
+        let tracker = if !clean_session {
+            let saved_state = saved.and_then(|saved| {
                 connection.events = saved.metrics;
+                saved.session_state
+            });
 
-                let Some(session_state) = saved.session_state else {
-                    break 'tracker Tracker::new(client_id.clone());
-                };
-                connection.subscriptions = session_state.subscriptions;
-                // for using in acklog
-                pending_acks = session_state.unacked_pubrels.clone();
-                outgoing.unacked_pubrels = session_state.unacked_pubrels;
-                session_state.tracker
-            } else {
-                // Only retrieve metrics in clean session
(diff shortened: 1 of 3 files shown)


rumqttd/src/router/routing.rs
```diff
@@ -304,23 +303,32 @@ impl Router {
...
+        let tracker = 'tracker: {
+            if !clean_session {
+                let Some(saved) = saved else {
+                    break 'tracker Tracker::new(client_id.clone());
+                };
+                connection.events = saved.metrics;
+
+                let Some(session_state) = saved.session_state else {
+                    break 'tracker Tracker::new(client_id.clone());
+                };
```
henil, line 319: can we directly return `Tracker::new()` without using break with `'tracker` here?

swanandx: can't use `return` as it will directly return from the function! We want tracker in the `tracker` variable as result of the expression.

Otherwise, we would have to rely in way too much nested if-let-else blocks.

swanandx: tbh, i really loved the `'labels` haha, but updated code to not use em. as result, we have to use `map_or_else` to avoid crazy nesting of blocks :)

rumqttd/src/router/routing.rs
```diff
@@ -304,23 +303,32 @@ impl Router {
...
+            saved_state.map_or_else(
+                || Tracker::new(client_id.clone()),
+                |session_state| {
+                    connection.subscriptions = session_state.subscriptions;
+                    // for using in acklog
+                    pending_acks = session_state.unacked_pubrels.clone();
+                    outgoing.unacked_pubrels = session_state.unacked_pubrels;
+                    session_state.tracker
+                },
+            )
```
henil, line 325: looks like its going to be hard to read this way as well, add comment on top of this explaining what it does. else LGTM.

henil (approved):

rumqttd/src/router/routing.rs
swanandx: > looks like its going to be hard to read this way as well
>
couldn't think of any other better way :sweat_smile: 

> add comment on top of this explaining what it does
> 
Done :smile:

Thanks for the review :)

PR #795, opened 2024-02-09
chore(rumqttd): Update rustls and friends.

qwandor:
## Type of change

## Checklist:

### DIFF
--- a/rumqttd/Cargo.toml
+++ b/rumqttd/Cargo.toml
@@ -20,10 +20,10 @@ flume = { version = "0.11.0", default-features = false, features = ["async"]}
 slab = "0.4.9"
 thiserror = "1.0.49"
 tokio-util = { version = "0.7", features = ["codec"], optional = true }
-tokio-rustls = { version = "0.24", optional = true }
-rustls-webpki = { version = "0.101.6", optional = true }
-tokio-native-tls = { version = "0.3", optional = true }
-rustls-pemfile = { version = "1", optional = true }
+tokio-rustls = { version = "0.25.0", optional = true }
+rustls-webpki = { version = "0.102.1", optional = true }
+tokio-native-tls = { version = "0.3.1", optional = true }
+rustls-pemfile = { version = "2.0.0", optional = true }
 async-tungstenite = { version = "0.23", default-features = false, features = ["tokio-runtime"], optional = true }
 ws_stream_tungstenite = { version= "0.11", default-features = false, features = ["tokio_io"], optional = true }
 x509-parser = {version= "0.15.1", optional = true}
--- a/rumqttd/src/link/bridge.rs
+++ b/rumqttd/src/link/bridge.rs
@@ -18,8 +18,8 @@ use tokio::{
 #[cfg(feature = "use-rustls")]
 use tokio_rustls::{
     rustls::{
-        client::InvalidDnsNameError, Certificate, ClientConfig, Error as TLSError,
-        OwnedTrustAnchor, PrivateKey, RootCertStore, ServerName,
+        pki_types::{InvalidDnsNameError, ServerName},
+        ClientConfig, Error as TLSError, RootCertStore,
     },
     TlsConnector,
 };
@@ -198,28 +198,15 @@ pub async fn tls_connect<P: AsRef<Path>>(
 ) -> Result<Box<dyn N>, BridgeError> {
     let mut root_cert_store = RootCertStore::empty();
 
-    let ca_certs = rustls_pemfile::certs(&mut BufReader::new(Cursor::new(fs::read(ca_file)?)))?;
-    let trust_anchors = ca_certs.iter().map_while(|cert| {
-        if let Ok(ta) = webpki::TrustAnchor::try_from_cert_der(&cert[..]) {
-            Some(OwnedTrustAnchor::from_subject_spki_name_constraints(
-                ta.subject,
-                ta.spki,
-                ta.name_constraints,
-            ))
-        } else {
-            None
-        }
-    });
-
-    root_cert_store.add_trust_anchors(trust_anchors);
+    let ca_certs = rustls_pemfile::certs(&mut BufReader::new(Cursor::new(fs::read(ca_file)?)))
+        .collect::<Result<Vec<_>, _>>()?;
+    root_cert_store.add_parsable_certificates(ca_certs);
 
     if root_cert_store.is_empty() {
         return Err(BridgeError::NoValidCertInChain);
     }
 
-    let config = ClientConfig::builder()
-        .with_safe_defaults()
-        .with_root_certificates(root_cert_store);
+    let config = ClientConfig::builder().with_root_certificates(root_cert_store);
 
     let config = if let Some(ClientAuth {
         certs: certs_path,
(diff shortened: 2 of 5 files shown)


rumqttd/src/link/bridge.rs
```diff
@@ -198,62 +198,50 @@ pub async fn tls_connect<P: AsRef<Path>>(
...
+            .next()
+            .ok_or(BridgeError::NoValidCertInChain)??
+            .into(),
+            Some(rustls_pemfile::Item::Pkcs8Key(_)) => rustls_pemfile::pkcs8_private_keys(
                 &mut BufReader::new(Cursor::new(fs::read(key_path)?)),
-            )?,
+            )
+            .next()
+            .ok_or(BridgeError::NoValidCertInChain)??
+            .into(),
```
swanandx, line 234: why do we need to re-read the same key? this would be same as:

```rust
let read_key = match rustls_pemfile::read_one(&mut BufReader::new(Cursor::new(fs::read(
    key_path,
)?)))? {
    Some(rustls_pemfile::Item::Pkcs1Key(key)) => key.into(),
    Some(rustls_pemfile::Item::Pkcs8Key(key)) => key.into(),
    Some(rustls_pemfile::Item::Sec1Key(key)) => key.into(),
    None | Some(_) => return Err(BridgeError::NoValidCertInChain),
};

```
also, i think `read_one` shall be called in loop ( as we should ignore _other_ `Item` which aren't _keys_, just like `xx_private_keys()` fn do ), so:

```rust
let read_key = loop {
    match rustls_pemfile::read_one(&mut BufReader::new(Cursor::new(fs::read(key_path)?)))? {
        Some(rustls_pemfile::Item::Pkcs1Key(key)) => break key.into(),
        Some(rustls_pemfile::Item::Pkcs8Key(key)) => break key.into(),
        Some(rustls_pemfile::Item::Sec1Key(key)) => break key.into(),
        Some(_) => {}
        None => return Err(BridgeError::NoValidCertInChain),
    }
};

```

rumqttd/src/link/bridge.rs
```diff
@@ -198,62 +198,50 @@ pub async fn tls_connect<P: AsRef<Path>>(
...
-            ))
-        } else {
-            None
-        }
-    });
-
-    root_cert_store.add_trust_anchors(trust_anchors);
+    let ca_certs = rustls_pemfile::certs(&mut BufReader::new(Cursor::new(fs::read(ca_file)?)))
+        .collect::<Result<Vec<_>, _>>()?;
+    root_cert_store.add_parsable_certificates(ca_certs);
```
swanandx, line 203: also using `add_parsable_certificates` might leave out some certs right?

wdyt of rewriting this section ( from line 201-103 ) as:
```rust
for cert in rustls_pemfile::certs(&mut BufReader::new(Cursor::new(fs::read(ca_file)?))) {
    root_cert_store.add(cert?)?
}
```

rumqttd/src/server/tls.rs
```diff
@@ -200,39 +200,42 @@ impl TLSAcceptor {
...
             let ca_file = File::open(ca_path);
             let ca_file = ca_file.map_err(|_| Error::CaFileNotFound(ca_path.clone()))?;
             let ca_file = &mut BufReader::new(ca_file);
-            let ca_certs = rustls_pemfile::certs(ca_file)?;
+            let ca_certs = rustls_pemfile::certs(ca_file).collect::<Result<Vec<_>, _>>()?;
             let ca_cert = ca_certs
-                .first()
-                .map(|c| Certificate(c.to_owned()))
+                .into_iter()
+                .next()
```
swanandx, line 224: if we only need the _first_ one, shall we really check if all other certs are valid are not? ( this is what `collect` does right? )

rumqttd/src/server/tls.rs
```diff
@@ -200,39 +200,42 @@ impl TLSAcceptor {
...
             let mut store = RootCertStore::empty();
             store
-                .add(&ca_cert)
+                .add(ca_cert)
                 .map_err(|_| Error::InvalidCACert(ca_path.to_string()))?;
 
-            builder.with_client_cert_verifier(Arc::new(AllowAnyAuthenticatedClient::new(store)))
+            // This will only return an error if no trust anchors are provided or invalid CRLs are
+            // provided. We always provide a trust anchor, and don't provide any CRLs, so it should
+            // never return an error and it is safe to unwrap.
```
swanandx, line 234:
```suggestion
            // This will only return an error if no trust anchors or invalid CRLs are provided.
            // We always provide a trust anchor, and don't provide any CRLs,
            // so it is safe to unwrap.
```

rumqttd/Cargo.toml
```diff
@@ -20,10 +20,10 @@ flume = { version = "0.11.0", default-features = false, features = ["async"]}
 slab = "0.4.9"
 thiserror = "1.0.49"
 tokio-util = { version = "0.7", features = ["codec"], optional = true }
-tokio-rustls = { version = "0.24", optional = true }
-rustls-webpki = { version = "0.101.6", optional = true }
-tokio-native-tls = { version = "0.3", optional = true }
-rustls-pemfile = { version = "1", optional = true }
+tokio-rustls = { version = "0.25.0", optional = true }
+rustls-webpki = { version = "0.102.1", optional = true }
+tokio-native-tls = { version = "0.3.1", optional = true }
```
swanandx, line 25: would it be better to not include the PATCH version. basically only like `0.25` and `0.3` or `2`?

swanandx (changes requested):
Thank you so much for the PR! :rocket: 

please feel free to comment on requested changes :)

rumqttd/src/link/bridge.rs
```diff
@@ -198,62 +198,50 @@ pub async fn tls_connect<P: AsRef<Path>>(
...
-        let certs = read_certs.into_iter().map(Certificate).collect();
-        config.with_client_auth_cert(certs, PrivateKey(read_key))?
+        config.with_client_auth_cert(read_certs, read_key)?
     } else {
         config.with_no_client_auth()
     };
 
     let connector = TlsConnector::from(Arc::new(config));
-    let domain = ServerName::try_from(host).unwrap();
+    let domain = ServerName::try_from(host).unwrap().to_owned();
```
swanandx, line 244:
```suggestion
    let domain = ServerName::try_from(host)?.to_owned();
```

rumqttd/src/link/bridge.rs
```diff
@@ -198,62 +198,50 @@ pub async fn tls_connect<P: AsRef<Path>>(
...
         };
 
-        let read_key = match read_keys.first() {
-            Some(v) => v.clone(),
-            None => return Err(BridgeError::NoValidCertInChain),
-        };
-
-        let certs = read_certs.into_iter().map(Certificate).collect();
-        config.with_client_auth_cert(certs, PrivateKey(read_key))?
+        config.with_client_auth_cert(read_certs, read_key)?
```
swanandx, line 238: nit: rename `read_certs` to `certs` ( similarly `key` )

rumqttd/Cargo.toml
qwandor: Why? I only tested with this version so there is no guarantee that an earlier version would work, and specifying the full versions still allows newer semver-compatible versions. E.g. `0.3.1` allows 0.3.1 <= version < 0.4.0, whereas `0.3` allows 0.3.0 <= version < 0.4.0.

swanandx: gotcha, thanks for clarifying 😃

rumqttd/src/link/bridge.rs
qwandor: Good question, the existing code was doing it this way so I assumed there was some reason for it.

The existing code doesn't support sec1 or ignoring non-key items so maybe let's change that in a separate PR, rather than mixing behaviour changes with dependency updates.

rumqttd/src/server/tls.rs
qwandor: That's what the existing code is doing, but okay.

swanandx: > That's what the existing code is doing, but okay.
>

right, but I just wanted to know from users perspective, what would one expect it to work like, will only one ( in this care, the first one ) being correct and other can be whatever should work, or whole file must be correct for it to work?

that is why I asked it :sweat_smile: wdyt?

rumqttd/src/link/bridge.rs
swanandx: > The existing code doesn't support sec1 or ignoring non-key items so maybe let's change that in a separate PR, rather than mixing behaviour changes with dependency updates.
>

fair point, lets open PR after this PR is merged :)

rumqttd/src/server/tls.rs
swanandx: also, as per this [comment](https://github.com/bytebeamio/rumqtt/pull/795#discussion_r1489471752), would it be better to club this change with that follow-up PR?

swanandx (approved):
LGTM :rocket: ! ( except unresolved questions )

Please lmk if you have tested it, I will also test it once to verify integrity and then we are good to go :) 

Thank you so much for your contribution!

rumqttd/src/server/tls.rs
swanandx: ah okie, then current state is fine.

rumqttd/src/link/bridge.rs
swanandx: hey @qwandor , would you like to open PR for this?

swanandx: I have opened the PR: https://github.com/bytebeamio/rumqtt/pull/802 :)

PR #800, opened 2024-02-17
ci(actions): implement code coverage

arunanshub:
## Type of change

## Checklist:

Not relevant to users since it is a CI update.


.github/workflows/build.yml
```diff
@@ -7,15 +7,23 @@ on:
       - '**.rs'
       - 'Cargo.*'
       - '*/Cargo.*'
+      - '.github/workflows/*'
```
swanandx, line 10: we should not run _this_ workflow incase the workflow files are edited, or new workflow is added.

I'm assuming it was added here so that we can trigger the workflow in this PR for testing, so it can be removed.

.github/workflows/build.yml
```diff
@@ -7,15 +7,23 @@ on:
...
+  push:
+    branches:
+      - main
```
swanandx, line 13: putting it here ( instead of separate file ), will trigger this bulky workflow after merging PRs as well ( and as we are running it before merge as well, is it worth it? )

directly pushing to main branch is bad and must not be performed.

wdyt?

.github/workflows/build.yml
```diff
@@ -57,4 +65,26 @@ jobs:
...
+      - name: Combine coverage
+        run: cargo llvm-cov report --lcov --output-path coverage.lcov
+
+      - name: Upload Coverage Report
+        uses: coverallsapp/github-action@v2.2.3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          file: ./coverage.lcov
+          parallel: true
+          flag-name: run-${{ matrix.os }}-cargo
```
swanandx, line 80: why do we need to upload it to coverallapp?

how would the flow look like? I mean, let's say I open a PR, then if I have to check the coverage, I would need to visit / lookup where on coverallapp is it available? so what is the benefit?

why can't we just comment it right below the PR? and why coverall instead of Codecov?

.github/workflows/build.yml
```diff
@@ -57,4 +65,26 @@ jobs:
         run: cargo hack doc --verbose --no-deps --each-feature --no-dev-deps --optional-deps url -p rumqttc -p rumqttd
 
       - name: Test rumqttc and rumqttd
-        run: cargo hack test --verbose --each-feature --optional-deps url -p rumqttc -p rumqttd
+        run: |
+            cargo hack --each-feature --optional-deps url llvm-cov -p rumqttc -p rumqttd --no-report nextest
```
swanandx, line 69: can you elaborate this part? is it running `cargo hack llvm-cov` and `llvm-con` runs `nextest`? 

if is kinda confusing to figure out what's exactly going on due to arrangement of arguments and flags, can we refactor if possible?

.github/workflows/build.yml
```diff
@@ -35,9 +43,9 @@ jobs:
         with:
           shared-key: ${{ runner.os }}-cargo
 
-      - uses: taiki-e/install-action@v1
+      - uses: taiki-e/install-action@v2
         with:
-          tool: cargo-hack
+          tool: cargo-hack,cargo-llvm-cov,nextest
```
swanandx, line 48: have we verified if `nextest` is really beneficial? I can see the CI runtime is almost same, can you just confirm once if `nextest` is adding anything to value?

swanandx (commented):
Hey, just wanted to clear some doubts, I have asked em in review comments, thank you :)

.github/workflows/build.yml
arunanshub: > directly pushing to main branch is bad and must not be performed.

`push: branches: [main]` serves two purposes.

1. It executes when someone pushes directly to `main`, which we don't want anyway.
2. Also, it will be executed when the pull request gets merged into main, and becomes a commit on the `main` branch.

We need that to satisfy the second condition only. If you want to be extra safe, disable `push to main` from repo settings.

.github/workflows/build.yml
arunanshub: If we install coveralls app, we will get a comment after every run of the test.

Something like this: https://github.com/arunanshub/pyflocker/pull/444#issuecomment-1951575395

Coveralls instead of codecov because it has less failure rates (due to network errors, etc) that codecov. And coveralls especially because our use case is simple: we just need the coverage report and a badge.

.github/workflows/build.yml
swanandx: > it will be executed when the pull request gets merged into main
>

workflow must pass before merging to main, it will basically be just two runs of same thing. that i why i asked

.github/workflows/build.yml
arunanshub: > we should not run this workflow incase the workflow files are edited, or new workflow is added.

Care to explain the logic behind this?

.github/workflows/build.yml
swanandx: nextest isn't running [doctest](https://github.com/nextest-rs/nextest/issues/16), which means we are just ignoring them.

what features do we want exactly though? just curious. are we seeing improvement in test suite runtime? Other features require some configuration as well right?

.github/workflows/build.yml
swanandx: for example, addition of dependabot workflow, when ( if ) we would add it / change it, this workflow will get triggered, which won't add any value.

that was logic behind it.

.github/workflows/build.yml
swanandx: so basically like:
```sh
cargo hack llvm-cov nextest --no-report --each-feature --optional-deps url -p rumqttc -p rumqttd
```

.github/workflows/build.yml
arunanshub: > which won't add any value

Certainly, GitHub does not impose fees for workflow runs on public repositories. Could you kindly elaborate on the specific value proposition you are referencing?

.github/workflows/build.yml
arunanshub: >  just two runs of same thing

We need a separate run on the main branch to get the coverage badge from coveralls.

.github/workflows/build.yml
arunanshub: > which means we are just ignoring them

Neither llvm-cov supports coverage over doctests. We can test the docs separately for now, and when nextest starts supporting them, we'll include them too.

.github/workflows/build.yml
swanandx: what is a badge?

arunanshub: in https://coveralls.io/github/bytebeamio/rumqtt:

[Coverage Status](https://coveralls.io/github/bytebeamio/rumqtt?branch=main)

.github/workflows/build.yml
arunanshub: > `--each-feature`

This is a flag of `cargo hack`, not `llvm-cov` or `nextest`

> `--optional-deps url`

This needs to be supplied to `cargo hack` to tell it that we need it include it as the optional dep when running test for each feature.

.github/workflows/build.yml
swanandx: > If we install coveralls app, we will get a comment after every run of the test.
>
ah okie, I didn't see any comment on this PR, so thought coverall doesn't support it.

> Coveralls instead of codecov because it has less failure rates (due to network errors, etc) that codecov.
>
nice :)

arunanshub: > ah okie, I didn't see any comment on this PR, so thought coverall doesn't support it.

We need admin access to install coveralls app.

.github/workflows/build.yml
swanandx: what is the use of badge? to showcase in readme? I think that can be ignored.

but yeah, if it will be used as _base_ for other PRs coverage comparison, then we gotta have it :+1:

.github/workflows/build.yml
swanandx: btw, instead of _uploading_ it to somewhere, we can just use some action to add comment in PR?

using actions like https://github.com/thollander/actions-comment-pull-request or https://github.com/marketplace/actions/add-pr-comment#proxy-for-fork-based-prs should do the job.

this way we won't need any other external setup.

downside? we might not be able to track the history, but code coverage history isn't required anyways right?

.github/workflows/build.yml
swanandx: value as in no point / benefit in running it.

even when Github doesn't impose fees, it feels off just to spam it. ( plus soon enough we won't be so sure why this line is required and someone gonna open PR to remove unnecessary condition haha :rofl: )

If we setup condition that workflow _must_ pass before merging, this can conflict with while adding workflows if it fails.

anyways, considering the frequency `.github/workflows/*` will be edited, I agree with you. let's keep it

.github/workflows/build.yml
arunanshub: > downside? we might not be able to track the history, but code coverage history isn't required anyways right?

It is beneficial to track how coverage has changed over commits, and comparing it over pull request.

.github/workflows/build.yml
swanandx: can you please try it out? it works for me :)

arunanshub: Since `--each-feature` is a flag of `cargo hack` not `llvm-cov` or `nextest`, do you see a problem with current approach?

.github/workflows/build.yml
swanandx: > It is beneficial to track how coverage has changed over commits
>
care to explain how is it beneficial?

> comparing pull request coverage to base branch's counterpart
>
okie :+1:

arunanshub: > care to explain how is it beneficial?

- Identify regressions: Catch errors introduced unintentionally by seeing drops in covered code.
- Measure test effectiveness: Monitor if new tests cover new code and if existing tests stay relevant.
- Target testing efforts: Prioritize tests for areas with low or decreasing coverage.

.github/workflows/build.yml
arunanshub: > not including doc tests still bugs me a lot and make me doubt the accuracy of the report haha ( especially in our case )

Will add it as a separate step.

.github/workflows/build.yml
swanandx: > For you.
> 
right :smile: readability is always dependent on person reading it, if i wrote it, I will find it readable as I know what I was doing, but that might not be the same for contributors / other maintainers of repo.

so, I don't wanna argue about which one is more readable. 

As a reviewer it was my duty to review it and present my thoughts. Rest is up-to the contributor :)

swanandx: as said earlier, choose whatever you feel readable :)

README.md
```diff
@@ -5,6 +5,9 @@
   <a href="https://github.com/bytebeamio/rumqtt/actions/workflows/main.yml">
     <img alt="build status" src="https://github.com/bytebeamio/rumqtt/actions/workflows/main.yml/badge.svg">
   </a>
+  <a href='https://coveralls.io/github/bytebeamio/rumqtt?branch=main'>
+    <img src='https://coveralls.io/repos/github/bytebeamio/rumqtt/badge.svg?branch=main' alt='Coverage Status' />
+  </a>
```
swanandx, line 10:
```suggestion
  <a href="https://coveralls.io/github/bytebeamio/rumqtt?branch=main">
    <img alt="coverage status" src="https://coveralls.io/repos/github/bytebeamio/rumqtt/badge.svg?branch=main">
  </a>
```

also is explicit branch `main` required? doesn't it default to it anyways.

README.md
```diff
@@ -5,6 +5,9 @@
   <a href="https://github.com/bytebeamio/rumqtt/actions/workflows/main.yml">
     <img alt="build status" src="https://github.com/bytebeamio/rumqtt/actions/workflows/main.yml/badge.svg">
```
swanandx, line 6: while we are at badges, and as you deleted `main.yml` workflow, how is this badge gonna work?

arunanshub: This workflow runs for every commit to main. That is, when a pull request gets merged, this workflow will run.

README.md
arunanshub: This is what coveralls suggests, to include the name of the branch.

README.md
swanandx: yup, but if i'm not wrong, the _badge_ checks the latest run status of workflow file with name `main.yml`.

and as there is no such workflow, the badge isn't going to work.

arunanshub: Oh right. I'll fix that.

swanandx: As the workflow is run with every PR, if we do it like that, it indicates the status of PR.

so if I just open PR which fails to build, it is gonna show fail. this was the reason I did change it to `main.yml` previously, you can see the commit history :)

arunanshub: I beg to differ.

```
[![CI](https://github.com/bytebeamio/rumqtt/actions/workflows/ci.yml/badge.svg?branch=main)](https://github.com/bytebeamio/rumqtt/actions/workflows/ci.yml)
```

By default, the badge is going to show the status of the default branch, not for every pull request. If you want to be pedantic, you can include the name of the branch too, as I did in the above code block. I recommend you go through the [docs for github actions](https://docs.github.com/en/actions).

swanandx (approved):

PR #803, opened 2024-02-20
ci(actions): disable incremental compilation to save space

arunanshub:
## Type of change

## Checklist:

CI update that should save 2GB of space and prevent rustc from stopping tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ name: CI
 
 env:
   RUSTFLAGS: "-D warnings"
+  CARGO_INCREMENTAL: 0
 
 defaults:
   run:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -43,6 +43,11 @@ jobs:
         with:
           shared-key: ${{ runner.os }}-cargo
 
+      - name: Disable incremental compilation
+        # Disable incremental compilation to save diskspace (the CI doesn't recompile modified files)
+        # https://github.com/apache/arrow-datafusion/issues/6676
+        run: echo "CARGO_INCREMENTAL=0" >> $GITHUB_ENV
```
swanandx, line 49: we can setup this flag at top in `env:` section.

PR #804, opened 2024-02-21
ci(actions): prevent "out of space" error on github actions ubuntu image

arunanshub:
## Type of change

## Checklist:

This is a CI change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,8 +13,10 @@ repository = "https://github.com/bytebeamio/rumqtt"
 license = "Apache-2.0"
 authors = ["tekjar <raviteja@bytebeam.io>"]
 
-[profile.dev.build-override]
-debug = true
+[profile.dev]
+# producing debug info in builds takes up a lot of space on runner, and we
+# are currently not using it.
+debug = false
 
 [profile.release]
 codegen-units = 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,10 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      # use mold linker instead of the default linker only on linux
+      - uses: rui314/setup-mold@v1
+        if: ${{ runner.os == 'Linux' }}
+
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,6 +36,15 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Run sccache-cache
+        uses: mozilla-actions/sccache-action@v0.0.4
+
+      - name: Set Rust caching env vars only on non-release runs
+        if: github.event_name != 'release' && github.event_name != 'workflow_dispatch'
+        run: |
+          echo "SCCACHE_GHA_ENABLED=true" >> $GITHUB_ENV
+          echo "RUSTC_WRAPPER=sccache" >> $GITHUB_ENV
+
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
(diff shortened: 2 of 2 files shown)


swanandx:
we do cache the build artifacts using Swatinem/rust-cache@v2 , and it seems working fine.

arunanshub:
sscache is different from rust-cache.

> sccache is a [ccache](https://ccache.dev/)-like compiler caching tool. It is used as a compiler wrapper and avoids compilation when possible, storing cached results either on [local disk](https://github.com/mozilla/sccache/blob/main/docs/Local.md) or in one of [several cloud storage backends](https://github.com/mozilla/sccache#storage-options).

swanandx:
> sscache is different from rust-cache.
>
how will it benefit us _differently_ from rust-cache here?

arunanshub:
## Observations and ideas

1. > Unfortunately, almost nobody does this. A [typical example](https://github.com/actions/cache/blob/main/examples.md#rust---cargo) would just cache the whole of ./target directory. That’s wrong — the ./target is huge, and most of it is useless on CI.
   
   Caching entire target directory provides no benefit.

2. How does sscache compare to rust-cache?
   > I personally don’t use sccache, as I didn’t find it to actually meaningfully improve compile times for me over what Cargo & CI caching natively do for my use-cases. Though, that was more than a couple of years ago. [matklad on Reddit](https://www.reddit.com/r/rust/comments/rvqxkf/comment/hr73ruo), [Faster rust builds](https://matklad.github.io/2021/09/04/fast-rust-builds.html)

3. Since this only happens on Ubuntu, is it tied to a linker issue?

swanandx:
> Though, that was more than a couple of years ago.
>
quoted em as you told me to check there

> [Faster rust builds](https://matklad.github.io/2021/09/04/fast-rust-builds.html)
>
this just confirms that rust-cache is good enough, and we don't need to change it.

> Since this only happens on Ubuntu, is it tied to a linker issue?
>
no idea

> Excluding debuginfo should save space.
>
yes

> Can we split up the build steps into their own groups? Eg, check/lint and tests in their own group?
>
why? no significant benefits there. We will know if CI failed due to compilation error of tests even without separation right?

arunanshub:
> We will know if CI failed due to compilation error of tests even without separation right?

My suspicion is that the steps before the actual test contribute to filling up the disk space. However if we build them as a separate job, it would use a separate image. After that job is finished, that image will be deallocated, along with the build artifacts. This is just a hypothesis, I have yet to test it out.

swanandx:
> My suspicion is that the steps before the actual test contribute to filling up the disk space
>

don't think so, actual issue here should be that code coverage. It may the culprit to fill up the disk space like https://github.com/taiki-e/cargo-llvm-cov/issues/284 . llvm-cov significantly reduced cache sizes as well.

> then why is it working on Mac and Windows?
>
might be due to difference in size of produced artifacts / caches? 

also sccache spammed the cache store https://github.com/bytebeamio/rumqtt/actions/caches . So it is hard to clear up all the caches and start from scratch.

arunanshub:
> also sscache spammed the cache store https://github.com/bytebeamio/rumqtt/actions/caches. So it is hard to clear up all the caches and start from scratch.

It isn't a big issue since GitHub will evict the caches anyway.

> might be due to difference in size of produced artifacts / caches?

I find it hard to believe that there is a significant difference in sizes of artifacts/caches produced by Ubuntu and mac/windows.

---

What's interesting is, the person who wrote `cargo-hack` deletes several unused dirs before testing on Ubuntu.

https://github.com/taiki-e/cargo-llvm-cov/issues/335#issuecomment-1890331417

Cargo.toml
```diff
@@ -13,8 +13,10 @@ repository = "https://github.com/bytebeamio/rumqtt"
 license = "Apache-2.0"
 authors = ["tekjar <raviteja@bytebeam.io>"]
 
-[profile.dev.build-override]
-debug = true
+[profile.dev]
+# producing debug info in builds takes up a lot of space on runner, and we
+# are currently not using it.
```
swanandx, line 18: this doesn't only affect the ci / runners. it will affect the local builds as well.

ideally we should have separated this change from only CI specific change.

Issue #805, opened 2024-02-22
RFC: rumqttc publish methods should return Pkid promise that resolves when packet is handled


de-sh:
This is the RFC for API changes in `publish` / `subscribe` / `unsubscribe` in rumqttc. Feel free to comment and provide feedback, thanks!

## Example in the case of Publish
```rust
let promise = async_client.publish(..).await?;
let pkid = promise.await;
```

Within rumqttc this will be implemented similar to how [oneshot channels](https://docs.rs/tokio/latest/tokio/sync/oneshot/fn.channel.html) work, i.e. [here](https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/state.rs#L337) we will respond onto the oneshot channel with the pkid value. Also we need not go so far when we know that the publish is on QoS 0 or when pkid is non-zero by default maybe?

swanandx:
so something like, `publish` will return `Pkid` where:
```rust
struct Pkid {
  id_rx: Receiver<u32>,
}

impl Future for Pkid {
  type Output = u32;
  fn poll(..) -> .. {
    // if we recv something on -d_rx, return id
    // we will send on that channel for state.
  }
}
```
right?

so I am curious how are we going to send the `Sender` to state so we can use it for sending the pkid?

> when pkid is non-zero by default maybe?
>
when is it non-zero by default?

> as users can just ignore/drop the promise token.
>
that would mean the _receiver_ is dropped as well. this would cause the `Sender` to fail!

de-sh:
> ```rust
>impl Future for Pkid {
>  type Output = u32;
>  fn poll(..) -> .. {
>    // if we recv something on -d_rx, return id
>    // we will send on that channel for state.
>  }
>}
>```
We can just use oneshot as is, no need of reimplementing

swanandx:
> Include it in the request?
>
gotcha. so unlike https://github.com/bytebeamio/rumqtt/compare/main...pkid-poc , by using the channels we are avoiding direct use of `Arc<Mutex<_>>` !

> [Not in the case of oneshot, it might error out,](https://docs.rs/tokio/latest/tokio/sync/oneshot/struct.Sender.html#method.send) 
>
that is what i meant by _fail_ :sweat_smile: 

> but we need not care about this either ways from within Eventloop
>
okie.

dlips:
Hi, I am one of the possible users of this functionality from #349 :) Thanks for bringing this up again

[...]

PR #806, opened 2024-02-22
feat: oneshot channel returns `Pkid`

### LINKED ISSUE #349: Get packet id for publish
There doesn't seem to be a way to get the packet id for a published QoS1/2 message, to track its state
Is this correct? Is there a workaround?

de-sh:
Fixes #349, implements #805 

## Type of change

Allows user access to pkid information once it is assigned by rumqttc internals during run.

## Checklist:

### Tests performed
Ran this against provided example async code, output looks as follows
```
Event = Incoming(ConnAck(ConnAck { session_present: false, code: Success }))
Event = Outgoing(Subscribe(1))
Pkid: 1
Event = Outgoing(Publish(2))
Pkid: 2
Event = Incoming(SubAck(SubAck { pkid: 1, return_codes: [Success(AtMostOnce)] }))
Event = Outgoing(PubRel(2))
Event = Incoming(PubRec(PubRec { pkid: 2 }))
Event = Incoming(Publish(Topic = hello/world, Qos = AtMostOnce, Retain = false, Pkid = 0, Payload Size = 1))
Event = Incoming(PubComp(PubComp { pkid: 2 }))
Event = Outgoing(Publish(3))
Pkid: 3
Event = Outgoing(PubRel(3))
Event = Incoming(PubRec(PubRec { pkid: 3 }))
Event = Incoming(Publish(Topic = hello/world, Qos = AtMostOnce, Retain = false, Pkid = 0, Payload Size = 2))
Event = Incoming(PubComp(PubComp { pkid: 3 }))
Event = Outgoing(Publish(4))
Pkid: 4
Event = Outgoing(PubRel(4))
Event = Incoming(PubRec(PubRec { pkid: 4 }))
Event = Incoming(Publish(Topic = hello/world, Qos = AtMostOnce, Retain = false, Pkid = 0, Payload Size = 3))
Event = Incoming(PubComp(PubComp { pkid: 4 }))
Event = Outgoing(Publish(5))
Pkid: 5
Event = Outgoing(PubRel(5))
Event = Incoming(PubRec(PubRec { pkid: 5 }))
Event = Incoming(Publish(Topic = hello/world, Qos = AtMostOnce, Retain = false, Pkid = 0, Payload Size = 4))
Event = Incoming(PubComp(PubComp { pkid: 5 }))
Event = Outgoing(Publish(6))
Pkid: 6
Event = Outgoing(PubRel(6))
Event = Incoming(PubRec(PubRec { pkid: 6 }))
Event = Incoming(Publish(Topic = hello/world, Qos = AtMostOnce, Retain = false, Pkid = 0, Payload Size = 5))
Event = Incoming(PubComp(PubComp { pkid: 6 }))
Event = Outgoing(PingReq)
Event = Outgoing(Publish(7))
Pkid: 7
Event = Incoming(PingResp)
Event = Outgoing(PubRel(7))
Event = Incoming(PubRec(PubRec { pkid: 7 }))
Event = Incoming(Publish(Topic = hello/world, Qos = AtMostOnce, Retain = false, Pkid = 0, Payload Size = 6))
Event = Incoming(PubComp(PubComp { pkid: 7 }))
Event = Outgoing(Publish(8))
Pkid: 8
Event = Outgoing(PubRel(8))
Event = Incoming(PubRec(PubRec { pkid: 8 }))
Event = Incoming(Publish(Topic = hello/world, Qos = AtMostOnce, Retain = false, Pkid = 0, Payload Size = 7))
Event = Incoming(PubComp(PubComp { pkid: 8 }))
Event = Outgoing(Publish(9))
Pkid: 9
Event = Outgoing(PubRel(9))
Event = Incoming(PubRec(PubRec { pkid: 9 }))
Event = Incoming(Publish(Topic = hello/world, Qos = AtMostOnce, Retain = false, Pkid = 0, Payload Size = 8))
Event = Incoming(PubComp(PubComp { pkid: 9 }))
Event = Outgoing(Publish(10))
Pkid: 10
Event = Outgoing(PubRel(10))
Event = Incoming(PubRec(PubRec { pkid: 10 }))
Event = Incoming(Publish(Topic = hello/world, Qos = AtMostOnce, Retain = false, Pkid = 0, Payload Size = 9))
Event = Incoming(PubComp(PubComp { pkid: 10 }))
Event = Outgoing(Publish(11))
Pkid: 11
Event = Outgoing(PubRel(11))
Event = Incoming(PubRec(PubRec { pkid: 11 }))
Event = Incoming(Publish(Topic = hello/world, Qos = AtMostOnce, Retain = false, Pkid = 0, Payload Size = 10))
Event = Incoming(PubComp(PubComp { pkid: 11 }))
Event = Outgoing(PingReq)
Event = Incoming(PingResp)
```
Thus the implementation is complete.

### DIFF
--- a/rumqttc/src/v5/state.rs
+++ b/rumqttc/src/v5/state.rs
@@ -485,12 +485,19 @@ impl MqttState {
     /// Adds next packet identifier to QoS 1 and 2 publish packets and returns
     /// it buy wrapping publish in packet
     fn outgoing_publish(&mut self, mut publish: Publish) -> Result<(), StateError> {
+        // NOTE: pkid promise need not be fulfilled for QoS 0,
+        // user should know this but still handled in Client.
         if publish.qos != QoS::AtMostOnce {
             if publish.pkid == 0 {
                 publish.pkid = self.next_pkid();
             }
 
             let pkid = publish.pkid;
+            // Fulfill the pkid promise
+            if let Some(pkid_tx) = publish.pkid_tx.take() {
+                _ = pkid_tx.send(pkid);
+            }
+
             if self
                 .outgoing_pub
                 .get(publish.pkid as usize)
@@ -604,6 +611,10 @@ impl MqttState {
 
         let pkid = self.next_pkid();
         subscription.pkid = pkid;
+        // Fulfill the pkid promise
+        if let Some(pkid_tx) = subscription.pkid_tx.take() {
+            _ = pkid_tx.send(pkid);
+        }
 
         debug!(
             "Subscribe. Topics = {:?}, Pkid = {:?}",
@@ -620,6 +631,10 @@ impl MqttState {
     fn outgoing_unsubscribe(&mut self, mut unsub: Unsubscribe) -> Result<(), StateError> {
         let pkid = self.next_pkid();
         unsub.pkid = pkid;
+        // Fulfill the pkid promise
+        if let Some(pkid_tx) = unsub.pkid_tx.take() {
+            _ = pkid_tx.send(pkid);
+        }
 
         debug!(
             "Unsubscribe. Topics = {:?}, Pkid = {:?}",
--- a/rumqttc/src/v5/state.rs
+++ b/rumqttc/src/v5/state.rs
@@ -1,3 +1,5 @@
+use crate::Pkid;
+
 use super::mqttbytes::v5::{
     ConnAck, ConnectReturnCode, Disconnect, DisconnectReasonCode, Packet, PingReq, PubAck,
     PubAckReason, PubComp, PubCompReason, PubRec, PubRecReason, PubRel, PubRelReason, Publish,
@@ -11,6 +13,7 @@ use bytes::{Bytes, BytesMut};
 use std::collections::{HashMap, VecDeque};
 use std::convert::TryInto;
 use std::{io, time::Instant};
+use tokio::sync::oneshot;
 
 /// Errors during state handling
 #[derive(Debug, thiserror::Error)]
@@ -151,7 +154,7 @@ impl MqttState {
         // remove and collect pending publishes
         for publish in self.outgoing_pub.iter_mut() {
             if let Some(publish) = publish.take() {
-                let request = Request::Publish(publish);
+                let request = Request::Publish(None, publish);
                 pending.push(request);
             }
         }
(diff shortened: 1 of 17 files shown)


swanandx (commented):
I like the idea and also tried it out, with both sync & async, and it works nicely!

Here is my feedback:
- we should avoid placing the `pkid_tx` directly in packets, mqttbytes should strictly be related to protocol parsing. We can maybe put it in request like `Request::Publish(tx, publish)` ?
- just `Pkid` seems intuitive than `PkidPromise` 
- users don't need to know about the oneshot channel, thus instead of just type alias, we should impl / rename some methods. e.g. async client is fine, but when in sync, using `pkid.blocking_recv()` isn't. maybe we can rename that recv to something else?
- if we fail to _send_ the pkid from state, maybe we can log some debug lvl info? not sure how often we will fail though.
- examples can be simplified, we just want to show _how user can get pkid_. So we can simply `await` right after getting `Pkid`. Using joinsets / delay queues in unnecessarily adding complexity. Also, users want to associate the publish with the pkid right? awaiting em later via joinset or something will be same as checking for `Outgoing` event. 
- as you suggested in RFC, having many open oneshot channels might addup perf impact which we are unsure of as of now. Whether to feature gate this or not is your call ( as you would know better haha )

de-sh:
> as you suggested in RFC, having many open oneshot channels might addup perf impact which we are unsure of as of now. Whether to feature gate this or not is your call ( as you would know better haha )

rumqttc/src/v5/state.rs
```diff
@@ -183,21 +186,21 @@ impl MqttState {
     /// be put on to the network by the eventloop
     pub fn handle_outgoing_packet(&mut self, request: Request) -> Result<(), StateError> {
         match request {
-            Request::Publish(publish) => {
+            Request::Publish(tx, publish) => {
                 self.check_size(publish.size())?;
```
de-sh, line 190: `EventLoop` drops the `pkid_tx` here if size is not valid

de-sh:
I have a few comments about the `PkidError` type, it needs to be more descriptive of why it failed to resolve into a valid number. This could be one of many:
1. `EventLoop` was shutdown, hence the tx was dropped.
2. `EventLoop` rejected the packet and hence did not assign it a pkid

We should probably handle this in a separate PR

swanandx:
> it's just to give users an idea of what is possible, they can choose to not go the route we illustrate, but we can try and educate them about the possibilities.
>
my thinking was, the examples should _show_ them _how_ they can get the pkid in general. for simple examples sake, getting pkid one by one should be fine imo ( in async & in sync )

we can make separate examples tailored toward why they might want to use joinset ( or something else ). There we should also explain why it is used and when they should and shouldn't use it.

> From a user perspective, you wouldn't want them to confuse between the awaiting and the occupied state.
>
I didn't get this completely, but i would like to differ. Adding `Promise` to name is too verbose(? idk correct word). e.g. the receiver is `Receiver` not `ReceiverPromise`.

de-sh:
> e.g. the receiver is Receiver not ReceiverPromise.

PR #812, opened 2024-03-04
fix(rumqttd/protocol/v5): Parse ConnAck and UnsubAck packets instead of panicking

izik1:
If re-adding the `_ => unreachable!()` is desired, let me know. I can't think of any benefit to it so I removed it.

## Type of change

## Checklist:

  - I have zero clue what to write in the changelog entry for this, so I currently haven't written anything

### DIFF
--- a/rumqttd/src/protocol/v5/mod.rs
+++ b/rumqttd/src/protocol/v5/mod.rs
@@ -426,7 +426,14 @@ impl Protocol for V5 {
                 let (pubcomp, properties) = pubcomp::read(fixed_header, packet)?;
                 Packet::PubComp(pubcomp, properties)
             }
-            _ => unreachable!(),
+            PacketType::ConnAck => {
+                let (puback, properties) = connack::read(fixed_header, packet)?;
+                Packet::ConnAck(puback, properties)
+            }
+            PacketType::UnsubAck => {
+                let (unsuback, properties) = unsuback::read(fixed_header, packet)?;
+                Packet::UnsubAck(unsuback, properties)
+            }
         };
 
         Ok(packet)


swanandx:
broker can never send a _connect_ packet, and hence, it can never _read_ an `ConnAck` ( same with unsuback ). Hence it isn't required and is _unreachable_.

izik1:
It's absolutely reachable, a non compliant (malicious) client can send a ConnAck/UnsubAck specifically because it knows a rumqttd broker is running (this happens before auth is checked, since checking auth requires reading a packet) and would panic the connection. Perhaps the correct course of action would instead be an error?

I got the body of the fix from "do what v4 does" and it does the same thing (hence writing a PR instead of an issue):
https://github.com/bytebeamio/rumqtt/blob/8af9d88c7862f2cbc2c69dcd98b8751ddf8c85dc/rumqttd/src/protocol/v4/mod.rs#L318
https://github.com/bytebeamio/rumqtt/blob/8af9d88c7862f2cbc2c69dcd98b8751ddf8c85dc/rumqttd/src/protocol/v4/mod.rs#L328
 if this is undesirable a change to v4 parsing is required as well.

swanandx:
oh yeah, you are right, we should not panic. Instead of error, lets parse it for now as you mentioned as do same in v4.

can we move handling connack right after connect ( like in v4, and same for unsuback ).

also, we can remove that `_ => unreachable` from v4 as well right?

for changelog, you can add under fix that "handle Connack and unsuback packets properly instead of panic" or something similar ( i'm not that good at this either haha )

thanks for pointing it out and the fix!

izik1:
Yeah, I can do all that.

swanandx:
hey @izik1 , this PR somehow slipped through, can you please fix the merge conflicts so that we can merge it? Thanks for understanding!

Issue #595, opened 2023-03-26
Empty strings are considered as valid topic


ThejasKiranPS:
# Expected Behavior

Publishing or Subscribing to an empty-string-topic should throw an error.

# Current Behavior

Allows the usage of empty string as topic.

# References 
[Topic semantic and usage](http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718109)

There is also a size limit and a null-character rule which I think is also not checked for.

ThejasKiranPS:
We could create a `Topic` struct which runs all these checks when a new topic is being created (`Topic::new(&str)`)

or manually check with [`fn valid_topic()`](https://github.com/bytebeamio/rumqtt/blob/ed3f78679d552dc1ae6043bdbcc731e58393ea97/rumqttd/src/protocol/mod.rs#L607) everywhere. Is there a better way to do this?

AlexandreCassagne:
> We could create a `Topic` struct which runs all these checks when a new topic is being created (`Topic::new(&str)`)
> 
> or manually check with [`fn valid_topic()`](https://github.com/bytebeamio/rumqtt/blob/ed3f78679d552dc1ae6043bdbcc731e58393ea97/rumqttd/src/protocol/mod.rs#L607) everywhere. Is there a better way to do this?

The use of a `Topic` type which handles validation at construction only is the way to go, IMO!

swanandx:
> The use of a Topic type which handles validation at construction only is the way to go, IMO!
>
agree, but topic/filter needs to be validated only once! 

in rumqttc, when we call publish / subscribe.
in rumqttd, when publish/subscribe packet is received and being handled.

So we have to think on pros/cons of using `Topic`. also, we just need to add some more checks like length and null byte in `validate_topic` fn and this should resolve the issue as well right?

PR #813, opened 2024-03-06
fix(rumqtt): validate topic and topic filter

arunanshub:
- fix(mqttbytes): add function to validate topic name and filter according to MQTTv3 spec
- fix(v5/client): validate topic name and alias

## Type of change

## Checklist:

> [!IMPORTANT]
> This is a proof of concept. We validate topic name/filter everytime client publishes/subscribes, respectively.

### DIFF
--- a/rumqttc/src/mqttbytes/topic.rs
+++ b/rumqttc/src/mqttbytes/topic.rs
@@ -1,12 +1,32 @@
+/// Maximum length of a topic or topic filter according to
+/// [MQTT-4.7.3-3](https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718109)
+pub const MAX_TOPIC_LEN: usize = 65535;
+
 /// Checks if a topic or topic filter has wildcards
 pub fn has_wildcards(s: &str) -> bool {
-    s.contains('+') || s.contains('#')
+    s.contains(['+', '#'])
 }
 
-/// Checks if a topic is valid
+/// Check if a topic is valid for PUBLISH packet.
 pub fn valid_topic(topic: &str) -> bool {
-    // topic can't contain wildcards
-    if topic.contains('+') || topic.contains('#') {
+    is_valid_topic_or_filter(topic) && !has_wildcards(topic)
+}
+
+/// Check if a topic is valid to qualify as a topic name or topic filter.
+///
+/// According to MQTT v3 Spec, it has to follow the following rules:
+/// 1. All Topic Names and Topic Filters MUST be at least one character long [MQTT-4.7.3-1]
+/// 2. Topic Names and Topic Filters are case sensitive
+/// 3. Topic Names and Topic Filters can include the space character
+/// 4. A leading or trailing `/` creates a distinct Topic Name or Topic Filter
+/// 5. A Topic Name or Topic Filter consisting only of the `/` character is valid
+/// 6. Topic Names and Topic Filters MUST NOT include the null character (Unicode U+0000) [MQTT-4.7.3-2]
+/// 7. Topic Names and Topic Filters are UTF-8 encoded strings, they MUST NOT encode to more than 65535 bytes.
+fn is_valid_topic_or_filter(topic_or_filter: &str) -> bool {
+    if topic_or_filter.is_empty()
+        || topic_or_filter.len() > MAX_TOPIC_LEN
+        || topic_or_filter.contains('\0')
+    {
         return false;
     }
 
@@ -17,7 +37,7 @@ pub fn valid_topic(topic: &str) -> bool {
 ///
 /// <https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106>
 pub fn valid_filter(filter: &str) -> bool {
-    if filter.is_empty() {
+    if !is_valid_topic_or_filter(filter) {
         return false;
     }
 
--- a/rumqttc/src/v5/client.rs
+++ b/rumqttc/src/v5/client.rs
@@ -6,7 +6,7 @@ use super::mqttbytes::v5::{
     Filter, PubAck, PubRec, Publish, PublishProperties, Subscribe, SubscribeProperties,
     Unsubscribe, UnsubscribeProperties,
 };
-use super::mqttbytes::QoS;
+use super::mqttbytes::{validate_topic_name_and_alias, QoS};
 use super::{ConnectionError, Event, EventLoop, MqttOptions, Request};
 use crate::valid_topic;
 
@@ -84,10 +84,11 @@ impl AsyncClient {
         P: Into<Bytes>,
     {
         let topic = topic.into();
+        let is_ok = validate_topic_name_and_alias(&topic, &properties);
         let mut publish = Publish::new(&topic, qos, payload, properties);
         publish.retain = retain;
         let publish = Request::Publish(publish);
-        if !valid_topic(&topic) {
+        if !is_ok {
             return Err(ClientError::Request(publish));
         }
         self.request_tx.send_async(publish).await?;
--- a/rumqttc/src/v5/mqttbytes/mod.rs
+++ b/rumqttc/src/v5/mqttbytes/mod.rs
@@ -1,4 +1,8 @@
+// reexport valid_filter and valid_topic since they are identical in nature for
+// both v3 and v5
+pub use crate::mqttbytes::{valid_filter, valid_topic};
 use std::{str::Utf8Error, vec};
+use super::PublishProperties;
 
 /// This module is the place where all the protocol specifics gets abstracted
 /// out and creates a structures which are common across protocols. Since,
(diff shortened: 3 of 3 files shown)


swanandx:
note that it won't _fix_ #595 as broker still doesn't perform the checks. I would recommend to remove that from PR description so it won't automatically close the issue if this PR is merged.

arunanshub:
It is still an ongoing work. Since broker will perform the same task, I will implement it in this PR itself.

rumqttc/src/client.rs
```diff
@@ -141,63 +141,91 @@ impl AsyncClient {
...
         Ok(())
     }
 
     /// Attempts to send a MQTT Subscribe for multiple topics to the `EventLoop`
     pub fn try_subscribe_many<T>(&self, topics: T) -> Result<(), ClientError>
     where
-        T: IntoIterator<Item = SubscribeFilter>,
+        T: IntoIterator<Item = SubscribeFilter> + Clone,
     {
+        let is_err = topics.clone().into_iter().any(|t| !valid_filter(t.path));
```
henil, line 199: i think we don't need `clone` here since `valid_filter` only needs a reference.

rumqttc/src/client.rs
```diff
@@ -338,25 +369,33 @@ impl Client {
     /// Sends a MQTT Subscribe for multiple topics to the `EventLoop`
     pub fn subscribe_many<T>(&self, topics: T) -> Result<(), ClientError>
     where
-        T: IntoIterator<Item = SubscribeFilter>,
+        T: IntoIterator<Item = SubscribeFilter> + Clone,
     {
+        let is_err = topics.clone().into_iter().any(|t| !valid_filter(t.path));
```
henil, line 374: same here

rumqttc/src/client.rs
```diff
@@ -338,25 +369,33 @@ impl Client {
...
+            return Err(ClientError::Request(request));
+        }
         self.client.request_tx.send(request)?;
         Ok(())
     }
 
     pub fn try_subscribe_many<T>(&self, topics: T) -> Result<(), ClientError>
     where
-        T: IntoIterator<Item = SubscribeFilter>,
+        T: IntoIterator<Item = SubscribeFilter> + Clone,
```
henil, line 386: same here

rumqttc/src/mqttbytes/topic.rs
```diff
@@ -1,12 +1,33 @@
...
+/// 6. Topic Names and Topic Filters MUST NOT include the null character (Unicode U+0000) [MQTT-4.7.3-2]
+/// 7. Topic Names and Topic Filters are UTF-8 encoded strings, they MUST NOT encode to more than 65535 bytes.
+fn is_valid_topic_or_filter(topic_or_filter: impl AsRef<str>) -> bool {
+    let topic_or_filter = topic_or_filter.as_ref();
+    if topic_or_filter.is_empty()
+        || topic_or_filter.len() > MAX_TOPIC_LEN
+        || topic_or_filter.contains('\0')
+    {
         return false;
     }
```
henil, line 32: i think we should rename this to be more descriptive since its an helper function, some topic/filter returning `true` here doesn't necessarily mean its a valid `topic/filter`. 

for example: `wrong/#/filter` will return true but its neither a valid topic name or topic filter.

rumqttc/src/v5/mqttbytes/mod.rs
```diff
@@ -32,61 +36,26 @@ pub fn qos(num: u8) -> Option<QoS> {
...
+    let is_alias_given = properties
+        .as_ref()
+        .map_or(false, |p| p.topic_alias.is_some());
+    let is_alias_valid = properties
+        .as_ref()
+        .map_or(false, |p| p.topic_alias.map_or(false, |a| a > 0));
+
+    (is_topic_valid || is_topic_empty)
+        && (!is_topic_empty || is_alias_given)
+        && (!is_alias_given || is_alias_valid)
```
henil, line 58: imo this is harder to read in comparison to if...else branch, wdyt @swanandx?

rumqttc/src/mqttbytes/topic.rs
arunanshub: I was thinking of a name that sounds something like "is a valid candidate for topic name or topic filter". Any ideas?

rumqttc/src/client.rs
arunanshub: If `Clone` is not used then `into_iter()` would consume `topics`. Which means we won't be able to use it any further (eg, in argument to `Subscribe::new_many`).

rumqttc/src/mqttbytes/topic.rs
henil: `can_be_topic_or_filter`?

rumqttc/src/client.rs
henil: we can collect it early and iterate over its reference and pass it to below functions, i.e. functions below this will need to be changed from taking an iterator to taking `&[T]` or `T`

flxo (commented):
Thanks for the contribution.

Is there a benchmark somewhere? The topic validation is in the hot path and must not have any crucial performance impact.

arunanshub:
Hello @flxo. The hot paths that you mention use generic functions (eg. `impl AsRef<str>`). According to [this discussion on Reddit](https://www.reddit.com/r/rust/comments/16ziv0i/comment/k3ex0pb), in release builds the generic functions are inlined. Furthermore, according to [matklad on his blog](https://matklad.github.io/2021/07/09/inline-in-rust.html#Inlining-in-Practice), private functions are always inlined. I believe that all of these factors should help with performance.

Also feel free to review the changes and suggest any scope of improvements, should you happen to find any.

flxo:
> Hello @flxo. The hot paths that you mention use generic functions (eg. `impl AsRef<str>`). According to [this discussion on Reddit](https://www.reddit.com/r/rust/comments/16ziv0i/comment/k3ex0pb), in release builds the generic functions are inlined. Furthermore, according to [matklad on his blog](https://matklad.github.io/2021/07/09/inline-in-rust.html#Inlining-in-Practice), private functions are always inlined. I believe that all of these factors should help with performance.


Sure - this all helps. I'm just curious if someone used e.g [criterion](https://crates.io/crates/criterion) to verify that best effort is done. 

> Also feel free to review the changes and suggest any scope of improvements, should you happen to find any.

Sure.

rumqttc/src/mqttbytes/topic.rs
```diff
@@ -1,12 +1,33 @@
...
+/// 3. Topic Names and Topic Filters can include the space character
+/// 4. A leading or trailing `/` creates a distinct Topic Name or Topic Filter
+/// 5. A Topic Name or Topic Filter consisting only of the `/` character is valid
+/// 6. Topic Names and Topic Filters MUST NOT include the null character (Unicode U+0000) [MQTT-4.7.3-2]
+/// 7. Topic Names and Topic Filters are UTF-8 encoded strings, they MUST NOT encode to more than 65535 bytes.
+fn can_be_topic_or_filter(topic_or_filter: impl AsRef<str>) -> bool {
+    let topic_or_filter = topic_or_filter.as_ref();
+    if topic_or_filter.is_empty()
+        || topic_or_filter.len() > MAX_TOPIC_LEN
+        || topic_or_filter.contains('\0')
```
flxo, line 29: How does `str::contains` compare to [memchr](https://github.com/BurntSushi/memchr)? Probably `str::contains` is well optimised but haven't checked.

rumqttc/src/mqttbytes/topic.rs
```diff
@@ -16,8 +37,9 @@ pub fn valid_topic(topic: &str) -> bool {
 /// Checks if the filter is valid
 ///
 /// <https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718106>
-pub fn valid_filter(filter: &str) -> bool {
-    if filter.is_empty() {
+pub fn valid_filter(filter: impl AsRef<str>) -> bool {
```
flxo, line 40: I'd name this `is_filter_valid`

rumqttc/src/v5/client.rs
```diff
@@ -84,10 +84,11 @@ impl AsyncClient {
         P: Into<Bytes>,
     {
         let topic = topic.into();
-        let mut publish = Publish::new(&topic, qos, payload, properties);
+        let is_ok = validate_topic_name_and_alias(&topic, &properties);
```
flxo, line 87: better name: `is_valid`

rumqttd/src/protocol/mod.rs
```diff
@@ -596,17 +596,33 @@ pub fn qos(num: u8) -> Option<QoS> {
 }
 
 /// Checks if a topic or topic filter has wildcards
-pub fn has_wildcards(s: &str) -> bool {
-    s.contains('+') || s.contains('#')
+pub fn has_wildcards(s: impl AsRef<str>) -> bool {
+    s.as_ref().contains(['+', '#'])
```
flxo, line 600: Same as above: Could `memchr` help here?

rumqttd/src/protocol/mod.rs
```diff
@@ -596,17 +596,33 @@ pub fn qos(num: u8) -> Option<QoS> {
...
-    if topic.contains('+') {
-        return false;
-    }
+/// Check if a topic is valid for PUBLISH packet.
+pub fn valid_topic(topic: impl AsRef<str>) -> bool {
+    is_valid_topic_or_filter(&topic) && !has_wildcards(topic)
+}
 
-    if topic.contains('#') {
+pub const MAX_TOPIC_LEN: usize = 65535;
```
flxo, line 608: There's a link to the MQTT standard obove. Can you please add this here as well?

rumqttd/src/protocol/mod.rs
```diff
@@ -616,32 +632,39 @@ pub fn valid_topic(topic: &str) -> bool {
...
-            if entry.len() > 1 && entry.contains('+') {
-                return false;
-            }
+    // rev() is used so we can easily get the last entry
+    let mut hirerarchy = filter.split('/').rev();
+
+    // split will never return an empty iterator
+    // even if the pattern isn't matched, the original string will be there
+    // so it is safe to just unwrap here!
+    let last = hirerarchy.next().unwrap();
```
flxo, line 647: The `unwrap` here can easily be avoided with a `let else` that returns `false`. yes - the unwrap is safe but maybe you want to forbid unwrap in the whole crate at some point in time and avoiding the use of `unwrap` make the pile of work less.

rumqttd/src/protocol/v5/publish.rs
```diff
@@ -34,6 +35,10 @@ pub fn read(
     bytes.advance(variable_header_index);
     let topic = read_mqtt_bytes(&mut bytes)?;
 
+    if !valid_topic(from_utf8(&topic).map_err(|_| Error::TopicNotUtf8)?) {
```
flxo, line 38: Here the validation is done (again) when obtaining `len`. This can be a problem when `len` is used during encoding since the utf8 check is not cheap at all. 

Have you thought about a `TryFrom` impl for all the packet types in here? Validate the parameters once at construction time (and maybe precalculate the length once.

rumqttc/src/mqttbytes/topic.rs
arunanshub: In [memchr's repo](https://github.com/BurntSushi/memchr?tab=readme-ov-file#why-is-the-standard-librarys-substring-search-so-much-slower) they've described `str.contains` to be slower than `memchr`. It would be a good idea to use an optimized routine in hotspots.

rumqttc/src/mqttbytes/topic.rs
arunanshub: Since it is a public function, we may run into the risk of backwards incompatibility due to name change. What do you think?

rumqttc/src/v5/client.rs
arunanshub: Done

rumqttc/src/v5/mqttbytes/mod.rs
arunanshub: I did some investigation. It seems that apart from the length of the function, the assembly for if-else version of `validate_topic_name_and_alias` remains the same. How should we proceed?

https://godbolt.org/z/6q4qd75nY

Issue #816, opened 2024-03-11
rumqttc: value returned by `v4::Connect.write` doesn't match reality


de-sh:
```rs
let connect = v4::Connect {
  protocol: Protocol::V4,
  keep_alive: u16::MAX,
  client_id: "".to_owned(),
  clean_session: true,
  last_will: Some(LastWill::new(
    "".to_owned(),
    Bytes::new(),
    QoS::AtLeastOnce,
    true,
  )),
  login: Some(Login::new("", "")),
};
let mut buf = BytesMut::new();
let written = connect.write(&mut buf);
        
assert_eq!(written, Ok(buf.len()));
```

# Expected Behavior
written should be equal to 18

# Current Behavior

 It is equal to 16

# Failure Information (for bugs)

This part of the [code](https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/mqttbytes/v4/connect.rs#L133) evaluates written as len, which doesn't account for the length of header fields of the packet, i.e. 1 + count.

swanandx:
I think the `len()` method on any packet doesn't account for `FixedHeader` ( which is 2 bytes as per [standards](https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718029) ).

But in `write()`, as we write along with `FixedHeader`, we are seeing that deviation.

this brings the question of whether `len()` should return length of `Connect` struct, or the actual _Connect_ packet that will get sent.

>  i.e. 1 + count
>
can you please explain this a bit?

de-sh:
I believe `len()` accounts for the size of payload, while size of serialized packets we account for with `size()`

swanandx:
we have fixed it in rumqttc with PR #818 , but this issue is also observed in rumqttd as well, will close this once it is fixed there.

swanandx:
done with PR #819 :rocket: !

PR #818, opened 2024-03-11
Make sure `write` returns the number of bytes it writes everywhere

amokfa:
Solves #816 

## Type of change

`Connect::write` was returning wrong value in couple places.

## Checklist:


swanandx:
i think we can now remove this comment as it is outdate: https://github.com/bytebeamio/rumqtt/blob/8af9d88c7862f2cbc2c69dcd98b8751ddf8c85dc/rumqttd/src/protocol/v4/publish.rs#L66

swanandx (approved):
Thanks for the PR!

PR #819, opened 2024-03-12
Make Type::write return correct value for rumqttd

amokfa:
## Type of change

write method wasn't returning the number of bytes read correctly for some types in rumqttd.

## Checklist:


swanandx (approved):
thanks :) :rocket:

PR #825, opened 2024-03-19
refactor: simplify `Network`with `Framed<.., Codec>`

de-sh:
As discussed in #814 this simplifies `Network` using `Framed` and `Codec`

## Type of change

* `mqttbytes::Error` is no longer `Clone, Copy, PartialEq, Eq` and 2 new variants, for io and packet size limit
* Refactors different parts of the codebase to regularize and simplify.
* Deprecates bulk read and moves write buffer into `Network`, which might lead to issues with packets in the buffer being dropped when an error is faced. But this seems like premature-optimization to be frank

## Checklist:

- [x] Formatted with `cargo fmt`
- [x] Make an entry to `CHANGELOG.md` if it's relevant to the users of the library. If it's not relevant mention why.

PR #823, opened 2024-03-19
feat: rumqttc async v5 client request batching 

flxo:
Process multiple client requests before flushing the network in the async v5 client.

## Type of change

New feature (non-breaking change which adds functionality)

## Checklist:

- [X] Formatted with `cargo fmt`
- [X] Make an entry to `CHANGELOG.md` if it's relevant to the users of the library. If it's not relevant mention why.

Fixes:

* #810
* #814 
* #815

PR #826, opened 2024-03-20
feat: buffer `Framed<.., Codec>`  packets

### LINKED ISSUE #810: rumqttc: Outgoing publications are sent one by one
# Expected Behavior

Outgoing publications are sent to the network io in batches and not in inefficient writes per publish.

# Current Behavior

The (v5) client code "flushes" the outgoing queue after each request from the application [here](https://github.com/bytebeamio/rumqtt/blob/b8766e109edf5caaa579a89aa71199638690fa11/rumqttc/src/eventloop.rs#L233C58-L233C71). This leads to one `write` or `sendto` systemcall per publication. This is inefficient. 

## Proposal

Queue up to n requests from the request channel and flush similar to the processing of the incoming packages where up to 10 packets are processed in a batch. 

I implemented a small [poc](https://github.com/flxo/rumqtt/tree/hack-io-framed) that loops over requests and processes up to n incoming or outgoing packages before flushing the io. We're currently evaluating the performance impact in our scenario which is has publications from the client to the broker in high frequencies (a less common use case). 

Let me know if this is a possible way to tackle the problem.

de-sh:
resolves #810 on top of #825 

Defaults to buffering 10 packets before each flush, behavior can be changed with `MqttOptions.set_max_request_batch()`. Instantly flushes outgoing `PingReq` as well(required to not close the connection).

## Checklist:

### DIFF
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -2,7 +2,7 @@ use crate::{framed::Network, Transport};
 use crate::{Incoming, MqttState, NetworkOptions, Packet, Request, StateError};
 use crate::{MqttOptions, Outgoing};
 
-use crate::framed::N;
+use crate::framed::AsyncReadWrite;
 use crate::mqttbytes::v4::*;
 use flume::{bounded, Receiver, Sender};
 use tokio::net::{lookup_host, TcpSocket, TcpStream};
@@ -369,7 +369,7 @@ async fn network_connect(
         _ => options.broker_address(),
     };
 
-    let tcp_stream: Box<dyn N> = {
+    let tcp_stream: Box<dyn AsyncReadWrite> = {
         #[cfg(feature = "proxy")]
         match options.proxy() {
             Some(proxy) => proxy.connect(&domain, port, network_options).await?,
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -185,7 +185,7 @@ impl EventLoop {
             o = network.read() => {
                 let incoming = o?;
                 if let Some(packet) = self.state.handle_incoming_packet(incoming)? {
-                    network.send(packet).await?;
+                    network.write(packet).await?;
                 }
                 Ok(self.state.events.pop_front().unwrap())
             },
@@ -224,7 +224,7 @@ impl EventLoop {
             ), if !self.pending.is_empty() || (!inflight_full && !collision) => match o {
                 Ok(request) => {
                     if let Some(outgoing) = self.state.handle_outgoing_packet(request)? {
-                        network.send(outgoing).await?;
+                        network.write(outgoing).await?;
                     }
 
                     Ok(self.state.events.pop_front().unwrap())
@@ -239,8 +239,11 @@ impl EventLoop {
                 timeout.as_mut().reset(Instant::now() + self.mqtt_options.keep_alive);
 
                 if let Some(outgoing) = self.state.handle_outgoing_packet(Request::PingReq(PingReq))? {
-                    network.send(outgoing).await?;
+                    network.write(outgoing).await?;
                 }
+                // NOTE: Pings should be sent instantly
+                network.flush().await?;
+
                 Ok(self.state.events.pop_front().unwrap())
             }
         }
@@ -353,6 +356,7 @@ async fn network_connect(
             options.max_incoming_packet_size,
             options.max_outgoing_packet_size,
             network_timeout,
+            options.max_request_batch,
         );
         return Ok(network);
     }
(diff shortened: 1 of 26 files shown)


flxo (commented):
Thanks for the update!

I don't understand why there's homebrewn counting of the outgoing packets. The counter also includes packets generated in response to incoming packets, so the name of the configuration is misleading.

The manual count + flush also works against the capacity of `Framed` (which defaults here). Wouldn't it be better to leave the need to a implicit flush to `Framed`? Flushing must still happen after each batch of incoming packets *and* requests.

I totally agree with wrapping of `SinkExt::flush` with the timeout.

My suggestion is:

* rename `max_request_batch` to `io_buffer_capacity`. Apply that when `Framed` is constructed.
* Remove `Network::write` and use `SinkExt::feed` instead (or write a wrapper if you wan't to hide the`Sink`)
* Keep `Network::flush` with the timeout feature.

cheers

@flxo

rumqttc/src/eventloop.rs
```diff
@@ -185,7 +185,7 @@ impl EventLoop {
             o = network.read() => {
                 let incoming = o?;
                 if let Some(packet) = self.state.handle_incoming_packet(incoming)? {
-                    network.send(packet).await?;
+                    network.write(packet).await?;
```
flxo, line 188: This packet is never flushed unless there's a ping or `options.max_request_batch - 1` more packets "sent".

de-sh: Had the same doubt, was using tests as a crutch, but I see how even the tests aren't complete.

What type of acks should we instantly respond to, I guess only `PingResp`?

de-sh:
> Remove Network::write and use SinkExt::feed instead (or write a wrapper if you wan't to hide theSink)
I would like to keep write and wrap the internal workings for the time being, mainly because this isn't directly exposed to the user and readability of code is important for maintenance.

How does `Framed` do implicit flushing again? just calling. I have validated that using `framed.feed()` doesn't flush on breaching buffer, what could I be doing wrong?

rumqttc/src/eventloop.rs
flxo: I'm not so deep in the MQTT standard but think it's just `PingResp`.

rumqttc/src/framed.rs
```diff
@@ -23,28 +25,55 @@ impl Network {
...
+        let should_flush = match packet {
+            Packet::Connect(_) | Packet::PingReq | Packet::PingResp => true,
+            _ => false,
+        };
+        self.framed
+            .feed(packet)
+            .await
+            .map_err(StateError::Deserialization)?;
+
+        if should_flush || self.framed.write_buffer().len() + packet_size >= self.buffer_capacity {
```
flxo, line 66: `FramedWrite` does that job for you [here](https://docs.rs/tokio-util/latest/src/tokio_util/codec/framed_impl.rs.html#264). I think removing the manual buffer size check is good and avoid not necessary complexity here. 

I'd also do the "when to flush" logic in `EventLoop` and not in here because it must reflect MQTT and the IO part should be agnostic.

Add a loop (up to e.g 10) around the `select!` that breaks if `pending` or `request_rx` is empty and the rx buffer of `Framed` is empty after a `fill_buf`. There is no need for extra logic in `Network`.

like this:

```
for _ in 0..10 {
  select! {
    r = next_request() => {
       packet = state.process(r)
       network.feed(packet)
    }
    p = network.read() => {
       packet = state.handle_incoming(p)
       network.feed(packet)
       if matches(packet, PingResp) {
           // Ensure the flush by breaking.
           break;
       }
    }
    _ = ping_interval.tick() => {
       packet = PingReq
       network.feed(packet)
    }
  }

  let data_pending = network.fill_rx_buf(); // get inner of framed; get read_buffer_mut, fill it.

  // Break if there is nothing to do. 
  if pending.is_empty() && requests_rx.is_empty() && !data_pending() {
    break;
  }
}
network.flush()
```

rumqttc/src/v5/framed.rs
```diff
@@ -16,36 +16,66 @@ use std::time::Duration;
...
+            None => Err(StateError::ConnectionClosed),
+        }
+    }
+
+    /// Write packets into buffer, flushes `Connect`/`PingReq`/`PingResp` packets instantly,
+    /// or on breaching buffer capacity
+    pub async fn write(&mut self, packet: Packet) -> Result<(), StateError> {
+        let packet_size = packet.size();
+        let should_flush = match packet {
+            Packet::Connect(..) | Packet::PingReq(_) | Packet::PingResp(_) => true,
```
flxo, line 61: Same as above.

flxo (commented):
Left some comments. My suggestion is to move the "when to flush" logic to `EventLoop`. See comments.

de-sh:
closed in favor of #823

PR #828, opened 2024-03-25
fix(rumqttc): validate filters for subscribe.

### LINKED ISSUE #787: rumqttc: filter's are not validated before subscribing
# Expected Behavior

Filter's should be validated before subscribing and invalid filters should be rejected

# Current Behavior

No validation is performed

## Context

For publish, we check [valid_topic](https://github.com/bytebeamio/rumqtt/blob/742d8bacc20dbab70374ebc6df711276feecc887/rumqttc/src/client.rs#L84) before sending it to broker, we should do same for subscribe.

[valid_filter](https://github.com/bytebeamio/rumqtt/blob/742d8bacc20dbab70374ebc6df711276feecc887/rumqttc/src/mqttbytes/topic.rs#L22) is defined, but never used. ( same for v5! )

it should be used before [sending the request](https://github.com/bytebeamio/rumqtt/blob/742d8bacc20dbab70374ebc6df711276feecc887/rumqttc/src/client.rs#L155), similar to how we do in publish.

pranavgoel29:
Fixes #787

use `valid_filter()` to check if filter is valid or not while creating subscribe request.

## Type of change

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+* Validate filters while creating subscription requests.
 * Make v4::Connect::write return correct value
 
 ### Security
--- a/rumqttc/src/client.rs
+++ b/rumqttc/src/client.rs
@@ -3,7 +3,7 @@
 use std::time::Duration;
 
 use crate::mqttbytes::{v4::*, QoS};
-use crate::{valid_topic, ConnectionError, Event, EventLoop, MqttOptions, Request};
+use crate::{valid_filter, valid_topic, ConnectionError, Event, EventLoop, MqttOptions, Request};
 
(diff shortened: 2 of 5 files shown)


swanandx (approved):
noice :rocket: thanks for the PR!

arunanshub:
@swanandx this was already being tracked in #813, wasn't it?

swanandx:
> @swanandx this was already being tracked in #813, wasn't it?

that PR linked to issue which was regarding just _fixing_ the `validate_topic` function so we check for it's length right? This one is for using that `validate_filter` function, it doesn't fix the implementation of validation function.

anyways, having small and atomic PRs is always nice!

arunanshub:
@swanandx a notification from your side before merging would have been appreciated.

swanandx:
My bad 😅 this issue was assigned to @pranavgoel29 long ago, even before the mentioned PR was there, so we decided to merge this as it only tackled the issue in scope

Issue #830, opened 2024-03-26
rumqttc panics if network connection is closed by broker


swanandx:
# Expected Behavior

rumqttc should report an error that network connection was closed.

# Current Behavior

it reaches `unreachable!()` and hence causing panic

![image](https://github.com/bytebeamio/rumqtt/assets/73115739/97ff06d1-7e92-4375-aaf9-99b6470ec0bf)

PR #833, opened 2024-03-26
fix: don't panic on connection closure

de-sh:
Solves #830

## Checklist:


rumqttc/src/v5/state.rs
```diff
@@ -64,6 +64,8 @@ pub enum StateError {
     PubCompFail { reason: PubCompReason },
     #[error("Connection failed with reason '{reason:?}' ")]
     ConnFail { reason: ConnectReturnCode },
+    #[error("Connection closed by peer abruptly")]
+    ConnectionAborted
```
flxo, line 68: Would love to have this named `ConnectionClosed`. Think this is the term use widely when a `read` returns `0`.

PR #835, opened 2024-03-26
feat: add `bind_addr` for `NetworkOptions` to enable `TCPSocket.bind()`

caoruijuan13:
## Checklist:

### DIFF
--- a/rumqttc/src/lib.rs
+++ b/rumqttc/src/lib.rs
@@ -98,7 +98,10 @@
 #[macro_use]
 extern crate log;
 
-use std::fmt::{self, Debug, Formatter};
+use std::{
+    fmt::{self, Debug, Formatter},
+    net::SocketAddr,
+};
 
 #[cfg(any(feature = "use-rustls", feature = "websocket"))]
 use std::sync::Arc;
@@ -370,6 +373,7 @@ pub struct NetworkOptions {
     tcp_send_buffer_size: Option<u32>,
     tcp_recv_buffer_size: Option<u32>,
     conn_timeout: u64,
+    bind_addr: Option<SocketAddr>,
     #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
     bind_device: Option<String>,
 }
@@ -380,6 +384,7 @@ impl NetworkOptions {
             tcp_send_buffer_size: None,
             tcp_recv_buffer_size: None,
             conn_timeout: 5,
+            bind_addr: None,
             #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
             bind_device: None,
         }
@@ -402,6 +407,11 @@ impl NetworkOptions {
     /// get timeout in secs
     pub fn connection_timeout(&self) -> u64 {
         self.conn_timeout
+    }
+
+    pub fn set_bind_addr(&mut self, bind_addr: SocketAddr) -> &mut Self {
+        self.bind_addr = Some(bind_addr);
+        self
     }
 
     /// bind connection to a specific network device by name
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `size()` method on `Packet` calculates size once serialized.
 * `read()` and `write()` methods on `Packet`.
 * `ConnectionAborted` variant on `StateError` type to denote abrupt end to a connection
+* `bind_addr` for `NetworkOptions` to enable `TCPSocket.bind()`
 
 ### Changed
 
--- a/rumqttc/src/eventloop.rs
+++ b/rumqttc/src/eventloop.rs
@@ -324,6 +324,10 @@ pub(crate) async fn socket_connect(
             socket.set_recv_buffer_size(recv_buffer_size).unwrap();
         }
 
+        if let Some(bind_addr) = network_options.bind_addr {
+            socket.bind(bind_addr)?;
+        }
+
         #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
         {
             if let Some(bind_device) = &network_options.bind_device {


swanandx:
hey @caoruijuan13 , thanks for the contribution, can you please elaborate more on why this is required?

rumqttc/src/lib.rs
```diff
@@ -404,6 +409,11 @@ impl NetworkOptions {
         self.conn_timeout
     }
 
+    pub fn set_bind_addr(&mut self, bind_addr: SocketAddr) -> &mut Self {
```
flxo, line 412: Comment is missing here on a `pub fn`

flxo, line 412: What about using `impl` [std::net::ToSocketAddrs](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html#)?

caoruijuan13:
> hey @caoruijuan13 , thanks for the contribution, can you please elaborate more on why this is required?

In my actual use, I need to bind specific address for the client to avoid routing loop, maybe others need it too.

cavivie:
Fortunately I took another look at the PR list, otherwise I would have done this again.

Our company now has a multi-platform and multi-interface scenario test. This will cause the bind device to be unable to bind specific address routing in scenarios such as macOS/iOS/Windows. Since the routing characteristics of different platforms are different, it is difficult to use simple routing. Control is achieved by simply binding addresses for rumqtt to avoid re-routing decisions to achieve the same effect, which is really awesome and we need it too.

rumqttc/src/lib.rs
cavivie: > What about using `impl` [std::net::ToSocketAddrs](https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html#)?

Like [this](https://github.com/cavivie/rumqtt/commit/5d7ec634edbfccbe5929ae7b72adc7165c7f98fd#diff-a7c14b22199b25375255ee5a3d137a7f77ad9fdde94ff211cd27c6441f3cbd1dR412-R419)?

caoruijuan13: Cool, to_socket_addrs() is very useful.

rumqttc/src/lib.rs
```diff
@@ -404,6 +409,14 @@ impl NetworkOptions {
         self.conn_timeout
     }
 
+    /// bind connection to a specific socket address
+    pub fn set_bind_addr(&mut self, bind_addr: impl ToSocketAddrs) -> &mut Self {
+        self.bind_addr = bind_addr
+            .to_socket_addrs()
+            .map_or(None, |mut iter| iter.next());
```
swanandx, line 416: is it fine to just ignore if there was any error in converting to socket address? 

Someone can mistakenly supply invalid addr and think it's working just fine.

for now, as there is no mechanism for returning error here, maybe we can just panic or simply ignore just as we do rn.

which one would you prefer?

swanandx (commented):
Thank you so much for contributing!

Please do the needful for v5 client as well. 😁

rumqttc/src/lib.rs
cavivie: If it may cause an error, then I think we should not hide the error, either panic or throw. In this case, I think we can simply throw an error since there is no need to support too much chain calls.

```
/// bind connection to a specific socket address
pub fn set_bind_address(&mut self, bind_address: ToSocketAddrs) -> std::io::Result<()> {
    self.bind_address = bind_address.to_socket_addrs()?.next();
    Ok(())
}
```

caoruijuan13: I have also considered this issue and feel that return `Result` is better.

But `set_bind_device` is just like this.

caoruijuan13: And currently v5 is supported, they all use `rumqttc::NetworkOptions` option, but v5 client set it by `MqttOptions`, and client set it by `Eventloop`  just like:
```
    let mut mqtt_options_v5 = rumqttc::v5::MqttOptions::parse_url(url.clone()).unwrap();
    let mut network_options = rumqttc::NetworkOptions::new();
    network_options.set_bind_addr(*crate::app::addrs::LOCAL_ADDR);
    mqtt_options_v5.set_network_options(network_options);
```

cavivie: Since I'm not sure whether to return an error or mask an error, I think we shouldn't wrap the error behavior too much, so I think we should not use ToSocketAddrs, and this conversion should be decided by the caller (because we are not a low-level crate)

caoruijuan13: Yes, it wouldn't be a problem if we used SocketAddr instead of ToSocketAddrs.

cavivie:
@flxo @swanandx I wonder what's stopping us from going further now?

swanandx:
> @flxo @swanandx I wonder what's stopping us from going further now?

same changes needs to be done for v5 client as well.
And i think the discussion about what to do incase it fails to convert to sockeraddr isn't resolved right?

cavivie:
> same changes needs to be done for v5 client as well.

The v5 client is already supported because v5 [MqttOptions NetworkOptions](https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/v5/mod.rs#L98) references the already implemented [lib NetworkOptions](https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/lib.rs#L369), doesn't it?

> And i think the discussion about what to do incase it fails to convert to sockeraddr isn't resolved right?

As I said before, we use SocketAddr directly instead of ToSocksAddr, isn't this solution enough to solve this problem?
I think the conversion is handled by the caller rather than in this non-low-level crate, what you think so?

swanandx:
> The v5 client is already supported because v5 MqttOptions NetworkOptions references the already implemented lib NetworkOptions, doesn't it?
>

that is `bind_device` which is different from `bind_addr` right? if not, bind_device already exists for v4 client, so we don't even need anything.


> I think the conversion is handled by the caller rather than in this non-low-level crate, what you think so?
>
didn't get it, wdym?

cavivie:
> that is bind_device which is different from bind_addr right? if not, bind_device already exists for v4 client, so we don't even need anything.

bind_addr is different from bind_device. 
bind_addr: Bind a socket to a socket address like "192.168.128.1:8080".
bind_device: Bind a socket to a particular device like "eth0", as specified in the passed interface name. 

Currently v5 and non-v5 share the same `NetworkOptions` struct, it only needs to be implemented in one place, and the other place will be automatically implemented. For non-v5, just set `NetworkOptions` directly on `MqttOptions`, but for v5, need to create `NetworkOptions` and set it into the `EventLoop` struct.

If it’s still unclear, I have a [PR](https://github.com/bytebeamio/rumqtt/pull/861) here to make v5 and non-v5 network options consistent in the way they are used externally.

> didn't get it, wdym?

Use the (recommend) way:
```
/// callers can call `set_bind_address(addr: ToSocketAddrs)`.
/// example: set_bind_address("192.168.227.1:8086").
///
/// bind connection to a specific socket address
pub fn set_bind_address(&mut self, bind_address: SocketAddr) -> &mut Self {
    self.bind_address = Some(bind_address);
    self
}
```
Instead of the way(unrecommended):
```
/// callers can call `set_bind_address(addr: SocketAddr)`;
///
/// bind connection to a specific socket address
pub fn set_bind_address(&mut self, bind_address: ToSocketAddrs) -> std::io::Result<()> {
    self.bind_address = bind_address.to_socket_addrs()?.next();
    Ok(())
}
```

Usage for callers:
```
/// address convert errors should be handled by callers themself, 
/// not by the rumqttc library, this example just unwrap (panic) it.
use std::net::ToSocketAddrs;
let addr = "192.168.227.1:8086"
let addr: SocketAddr = addr.to_socket_addrs().unwrap();

/// use SocketAddr variable to call `set_bind_address`,
/// won't use `set_bind_address("192.168.227.1:8086")`.
network_options.set_bind_address(addr);
```

PR #837, opened 2024-03-28
rumqttc: fix subscribe many

flxo:
## Type of change

## Checklist:

Fix a flaw in the construction of the `Subscribe` packets where a iterator shouldn't be used twice.


arunanshub:
Hello @flxo, thank you for your contribution. Currently we are tracking and fixing the same issue in PR #813. It not only fixes issue with `subscribe_many`, but also fixes and issue with topic name and topic filter validation.

Once the changes are approved, that shall be merged into the main branch.

flxo:
Closing in favour of #813

PR #851, opened 2024-04-26
Acked POC improvements

xiaocq2001:
## Type of change

Added reason code for PubAck/Rec/Comp.
Fix notification flow for NoticeFuture and Eventloop.

## Checklist:


rumqttc/src/state.rs
```diff
@@ -309,8 +322,13 @@ impl MqttState {
             .remove(&pubrec.pkid)
             .ok_or(StateError::Unsolicited(pubrec.pkid))?;
 
+        // Notify user about the publish, pubrel and pubcomp will be handled in background
+        if let Some(tx) = tx  {
+            tx.success();
+        }
```
FSMaxB, line 328: What is "handled in the background" supposed to mean in this context.

From what I understand, this makes the `NoticeFuture` complete on `PubRec` instead of `PubComp` for `ExactlyOnce` QoS, which is the wrong behavior from what I understand.

FSMaxB (commented):
Just a question

rumqttc/src/state.rs
xiaocq2001: IMO the notice is to confirm that the message actually received by broker.
So in QoS0 it's noticed on message sent (since there is no ACK), in QoS1 it's noticed on ACK reception, in QoS2 it's noticed on REC which indicates broker accept the message. The PubRel and PubComp do not affect the broker reception of message, but some way to let broker release QoS2 message.

FSMaxB: > QoS2 it's noticed on REC which indicates broker accept the message. The PubRel and PubComp do not affect the broker reception of message, but some way to let broker release QoS2 message.

The original code by @de-sh notifies on `PubComp` not on `PubRec` for QoS2 which I believe to be the correct behavior. Because when using this feature, I want to be notified once I'm certain the message will fulfil it's QoS. For QoS2, the message is only sent after sending `PubRel` to the broker, which the broker then acknowledges by responding with `PubComp`. If the client dies after receiving `PubRec` but before the broker received `PubRel`, the message might get lost and the `NoticeFuture` shouldn't resolve for a message that might not fulfil it's QoS, which for 1 and 2 means it not getting lost.

At least that's my take based on my understanding of how MQTT QoS works and my take on the `NoticeFuture` feature.

xiaocq2001: What I mean is we can get notice on message ownership transfer, which is mentioned in the section 4.4, "When its original sender receives the PUBREC packet, ownership of the Application Message is transferred to the receiver". But I am totally fine to be noticed by PUBCOMP, where message ownership definitely transferred.

xiaocq2001: BTW I see you working on migrating this ack wait feature to main in https://github.com/bytebeamio/rumqtt/pull/883, hoping the PR and discuss here can help.

xiaocq2001:
Some updates for reference:
* HashMap -> VecDeque
  * To save memory
  * VecDeque is fast to push_back and pop_first
* Publish NoticeTx
* Notify on QoS2 PUBCOMP

KillingJacky:
Hi @xiaocq2001 , thanks for your work, this feature is exactly what I need for my current project. I can't wait for the upstream, and your PR seems quite promising, so how mature is your PR? If I can spend some effort to do the tests, can I use your PR right now?

xiaocq2001:
> Hi @xiaocq2001 , thanks for your work, this feature is exactly what I need for my current project. I can't wait for the upstream, and your PR seems quite promising, so how mature is your PR? If I can spend some effort to do the tests, can I use your PR right now?

Feel free to test and use that.

rumqttc/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `size()` method on `Packet` calculates size once serialized.
 * `read()` and `write()` methods on `Packet`.
 * `ConnectionAborted` variant on `StateError` type to denote abrupt end to a connection
+* `set_session_expiry_interval` and `session_expiry_interval` methods on `MqttOptions`.
```
de-sh, line 15: Sorry for the extremely late response, if you are still interested in contributing this change, please do open a separate PR. Currently this PR will be hard to accept given it has deviated from the issue focus.

de-sh (commented):
Please provide your suggestions against #916 if you have any

rumqttc/CHANGELOG.md
xiaocq2001: For session expiry interval support, it's already done in https://github.com/bytebeamio/rumqtt/pull/854.

KillingJacky:
> > Hi @xiaocq2001 , thanks for your work, this feature is exactly what I need for my current project. I can't wait for the upstream, and your PR seems quite promising, so how mature is your PR? If I can spend some effort to do the tests, can I use your PR right now?
> 
> Feel free to test and use that.

I'v done integrating with your `acked_test` branch and tested it with several test cases, so far so good, thank you very much for your work.

ryanwinter:
Thanks @KillingJacky!

rumqttc/src/state.rs
```diff
@@ -193,11 +210,14 @@ impl MqttState {
...
-        let tx = self
+        // No expecting ordered acks for suback
+        // Search outgoing_sub to find the right suback.pkid
+        let pos = self
             .outgoing_sub
-            .remove(&suback.pkid)
+            .iter()
+            .position(|(pkid, _)| *pkid == suback.pkid)
             .ok_or(StateError::Unsolicited(suback.pkid))?;
+        let (_, tx) = self.outgoing_sub.remove(pos).unwrap();
```
joshuachp, line 220:
```suggestion
        let (_, tx) = self.outgoing_sub.swap_remove_back(pos).unwrap();
```

Since the ordering is not important, could we maybe just swap_remove the element?

PR #852, opened 2024-04-28
feat: Added AUTH packet support for enhanced authentication.

ting-ms:
## Type of change

## Checklist:


rumqttc/src/v5/eventloop.rs
```diff
@@ -414,22 +421,34 @@ async fn mqtt_connect(
...
 
     // validate connack
-    match network.read().await? {
-        Incoming::ConnAck(connack) if connack.code == ConnectReturnCode::Success => {
-            if let Some(props) = &connack.properties {
-                if let Some(keep_alive) = props.server_keep_alive {
-                    options.keep_alive = Duration::from_secs(keep_alive as u64);
-                }
-                network.set_max_outgoing_size(props.max_packet_size);
+    loop {
```
de-sh, line 424: There are two ways to do this, one as has been already implemented in this PR, the other would involve improving the visibility into the whole process, updating the event log and returning events as they occur.
Example log of events
1. Outgoing Connect packet
2. Incoming/Outgoing Auth packets for as long as the challenge/response cycle goes on
4. Incoming Connack packet

This would require a refactor of how the code is currently structured, let me delve deeper into this.

de-sh (commented):
Would be wonderful if you could provide us with guidance on setting up a test with a broker that supports auth packets, thank you!

ting-ms:
> Would be wonderful if you could provide us with guidance on setting up a test with a broker that supports auth packets, thank you!


Please reference the following docs for the testing guidance:
[RumqttcAuthTestWithEMQX.pdf](https://github.com/bytebeamio/rumqtt/files/15492921/RumqttcAuthTestWithEMQX.pdf)
[RumqttcAuthTestWithEventGrid.pdf](https://github.com/bytebeamio/rumqtt/files/15492920/RumqttcAuthTestWithEventGrid.pdf)

ting-ms:
Attached updated design doc.
[RumqttcEnhancedAuhtenticationInterfaceDesign.pdf](https://github.com/bytebeamio/rumqtt/files/15495494/RumqttcEnhancedAuhtenticationInterfaceDesign.pdf)

rumqttc/src/v5/eventloop.rs
xiaocq2001: Is there any update on involve improving the visibility into the whole process, updating the event log and returning events as they occur?

Issue #853, opened 2024-04-30
Connect with session_expiry_interval for MQTT v5


xiaocq2001:
# Expected Behavior

When connecting to broker, `session_expiry_interval` should be supported.

# Current Behavior

There is no API to set `session_expiry_interval` for CONNECT request.

# Failure Information (for bugs)

## Context

* Operating System: Ubuntu 22.04
* Toolchain version: default

## Failure Logs

Please include any relevant log snippets or files here.

While running example `async_manual_acks_v5`, the broker does not restore the session and re-send publishes because `session_expiry_interval` value `0xFFFFFFFF` is not offered on connection:

```
Event = Incoming(ConnAck(ConnAck { session_present: true, code: Success, properties: Some(ConnAckProperties { session_expiry_interval: None, receive_max: Some(20), max_qos: None, retain_available: None, max_packet_size: None, assigned_client_identifier: None, topic_alias_max: Some(10), reason_string: None, user_properties: [], wildcard_subscription_available: None, subscription_identifiers_available: None, shared_subscription_available: None, server_keep_alive: None, response_information: None, server_reference: None, authentication_method: None, authentication_data: None }) }))
 DEBUG rumqttc::v5::state > Subscribe. Topics = [Filter { path: "hello/world", qos: AtLeastOnce, nolocal: false, preserve_retain: false, retain_forward_rule: OnEverySubscribe }], Pkid = 1
Event = Outgoing(Subscribe(1))
 DEBUG rumqttc::v5::state > Publish. Topic = hello/world, Pkid = 2, Payload Size = 1
Event = Outgoing(Publish(2))
 DEBUG rumqttc::v5::state > SubAck Pkid = 1, QoS = AtLeastOnce
Event = Incoming(SubAck(SubAck { pkid: 1, return_codes: [Success(AtLeastOnce)], properties: None }))
Event = Incoming(Publish(Publish { dup: false, qos: AtLeastOnce, retain: false, topic: b"hello/world", pkid: 21, payload: b"\x01", properties: None }))
Event = Incoming(PubAck(PubAck { pkid: 2, reason: Success, properties: None }))
 DEBUG rumqttc::v5::state > Publish. Topic = hello/world, Pkid = 3, Payload Size = 2
Event = Outgoing(Publish(3))
Event = Incoming(Publish(Publish { dup: false, qos: AtLeastOnce, retain: false, topic: b"hello/world", pkid: 22, payload: b"\x01\x01", properties: None }))
[...]

PR #854, opened 2024-04-30
feat: add MQTT v5 connection `session_expiry_interval` support

xiaocq2001:
Fixing issue https://github.com/bytebeamio/rumqtt/issues/853

## Type of change

* Added `session_expiry_interval` support for MQTT V5 connection, 

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `size()` method on `Packet` calculates size once serialized.
 * `read()` and `write()` methods on `Packet`.
 * `ConnectionAborted` variant on `StateError` type to denote abrupt end to a connection
-* `set_session_expiry_interval` and `get_session_expiry_interval` methods on `MqttOptions`.
+* `set_session_expiry_interval` and `session_expiry_interval` methods on `MqttOptions`.
 
 ### Changed
 
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `size()` method on `Packet` calculates size once serialized.
 * `read()` and `write()` methods on `Packet`.
 * `ConnectionAborted` variant on `StateError` type to denote abrupt end to a connection
+* `set_session_expiry_interval` and `get_session_expiry_interval` methods on `MqttOptions`.
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


rumqttc/examples/async_manual_acks_v5.rs
```diff
@@ -10,6 +10,7 @@ fn create_conn() -> (AsyncClient, EventLoop) {
     let mut mqttoptions = MqttOptions::new("test-1", "localhost", 1884);
     mqttoptions
         .set_keep_alive(Duration::from_secs(5))
+        .set_session_expiry_interval(u32::MAX.into())
```
swanandx, line 13: we only want to show manual acks in these example, why we need session expiry interval to be u32 max?

swanandx (commented):
Thanks for the PR! can you please also mention the changes in CHANGELOG as well?

rumqttc/examples/async_manual_acks_v5.rs
xiaocq2001: According to section 3.1.11.2 in https://docs.oasis-open.org/mqtt/mqtt/v5.0/mqtt-v5.0.pdf
```
720 If the Session Expiry Interval is 0xFFFFFFFF (UINT_MAX), the Session does not expire.
721
722 The Client and Server MUST store the Session State after the Network Connection is closed if the 
723 Session Expiry Interval is greater than 0 [MQTT-3.1.2-23].
```

Server stores state only if session expiry interval is greater than 0, so on connect it must be provided to broker. We want the connection session does not expire, so we set the expiry interval to UINT_MAX (u32 max) here.

xiaocq2001:
> Thanks for the PR! can you please also mention the changes in CHANGELOG as well?

Updated.

rumqttc/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `size()` method on `Packet` calculates size once serialized.
 * `read()` and `write()` methods on `Packet`.
 * `ConnectionAborted` variant on `StateError` type to denote abrupt end to a connection
+* `set_session_expiry_interval` and `get_session_expiry_interval` methods on `MqttOptions`.
```
de-sh, line 15:
```suggestion
* `set_session_expiry_interval` and `session_expiry_interval` methods on `MqttOptions`.
```

rumqttc/examples/async_manual_acks_v5.rs
de-sh: We probably should not make the change within this particular example as `u32::MAX` should be the default value?

xiaocq2001: The default is absent, the value 0 is used, in this case broker does not store the session state.
```
717 If the Session Expiry Interval is absent the value 0 is used. If it is set to 0, or is absent, the Session ends 
718 when the Network Connection is closed.
```

de-sh: Understood, this makes sense, LGTM @swanandx

xiaocq2001:
Moved the code to apply broker returned session_expiry_interval here, please check.

swanandx (approved):
LGTM! thanks for your contribution!

Issue #863, opened 2024-05-11
In mqttc v5, restore session only if session is resumed


xiaocq2001:
# Current Behavior

Currently rumqttc for MQTTv5 always republish packets which aren't acked, no matter it's clean session or resume session.

# Expected Behavior

For simple we can use the `session_present` in `CONNACK` returned by broker, if it's 1 the session is resumed, if it's 0 the session is new and the pending requests should be discarded.

de-sh:
@swanandx does broker by default set this correctly?

PR #864, opened 2024-05-14
rumqttc: resume session only if CONNACK with session present 1

xiaocq2001:
## Type of change

Implement for https://github.com/bytebeamio/rumqtt/issues/863.

Bug fix (non-breaking change which fixes an issue):

In MQTT, session is not always resumed, but in current mqttc the pending publishes are restored anyway.
With this fix the CONNACK session present returned from broker is checked, to confirm if a session is actually resumed, if not the pending publishes are cleaned and new session is started.

## Checklist:


rumqttc/src/v5/eventloop.rs
```diff
@@ -406,12 +407,22 @@ async fn mqtt_connect(
...
             if let Some(props) = &connack.properties {
+                // Override local keep_alive value if set by server.
                 if let Some(keep_alive) = props.server_keep_alive {
                     options.keep_alive = Duration::from_secs(keep_alive as u64);
                 }
                 network.set_max_outgoing_size(props.max_packet_size);
+                // Override local session_expiry_interval value if set by server.
+                if (props.session_expiry_interval).is_some() {
+                    options.set_session_expiry_interval(props.session_expiry_interval);
+                }
```
de-sh, line 425: These changes should rather go into #854

xiaocq2001: Fine, I'll move that.

de-sh (changes requested):
https://github.com/xiaocq2001/rumqtt/pull/1

swanandx:
can you please fix the code / tests? 

the `reconnection_resends_unacked_packets_from_the_previous_connection_first` test is failing, i think we need to connect with clean session false so that broker will save the state right?

xiaocq2001:
To test, I think we need https://github.com/bytebeamio/rumqtt/pull/854. With that, we need to connect with clean session false and session expiry interval non-zero to let broker save state.

xiaocq2001:
> To test, I think we need #854. With that, we need to connect with clean session false and session expiry interval non-zero to let broker save state.

Facing an issue on connection. Due to MQTT protocol difference between versions, in v5 session_expiry_interval is supported but in v3 it is not, what is the suggestion on handling such difference in tests?

de-sh:
> > To test, I think we need #854. With that, we need to connect with clean session false and session expiry interval non-zero to let broker save state.
> 
> Facing an issue on connection.
> 
> ```
> ---- reconnection_resumes_from_the_previous_state stdout ----
> Polled = Ok(Incoming(ConnAck(ConnAck { session_present: false, code: Success })))
> Polled = Ok(Outgoing(Publish(1)))
> Polled = Ok(Incoming(PubAck(PubAck { pkid: 1 })))
> Polled = Ok(Outgoing(Publish(2)))
> Polled = Ok(Incoming(PubAck(PubAck { pkid: 2 })))
> Polled = Ok(Outgoing(Publish(3)))
> Polled = Ok(Outgoing(Publish(4)))
> Polled = Ok(Outgoing(Publish(5)))
> Polled = Ok(Outgoing(PingReq))
> Polled = Ok(Outgoing(Publish(6)))
> Polled = Ok(Outgoing(Publish(7)))
> Polled = Ok(Outgoing(Publish(8)))
> Polled = Ok(Outgoing(Publish(9)))
> Polled = Ok(Outgoing(Publish(10)))
> Polled = Err(MqttState(AwaitPingResp))
> Polled = Ok(Incoming(ConnAck(ConnAck { session_present: false, code: Success })))
> ```
> 
> It seems on host side session is not stored?
> 
> Another question: due to MQTT protocol difference between versions, in v5 session_expiry_interval is supported but in v3 it is not, what is the suggestion on handling such difference in tests?

https://github.com/xiaocq2001/rumqtt/pulls solves this inconsistency in the tests

xiaocq2001:
I realize that in failing tests, the broker is started by `Broker::new`, which starts a broker without any stored state, that's why broker send ConnAck with session present false, we need some new method to introduce disconnection, maybe like what is in async_manual_ack example.

de-sh (approved):

PR #867, opened 2024-05-21
fix(disconnect): make `DisconnectProperties` struct public

arunanshub:
## Type of change

## Checklist:


de-sh (approved):
LGTM

de-sh:
Have cutout a release that does the same but as a patch to v0.23.0, if you are using stable, please use that as main is not prepared for next release.

Issue #868, opened 2024-05-21
rumqttc: Reduce memory usage of MqttState


hippalus:
# Expected Behavior

The MQTT state management should be optimized to use minimal memory while maintaining efficient handling of QoS 2 publish packets. We expect the system to handle incoming and outgoing QoS 2 publish packet identifiers efficiently without consuming excessive memory.

# Current Behavior
The current implementation utilizes Vec<Option<u16>> for both incoming_pub and outgoing_rel, resulting in a total memory usage of approximately 280KiB for these vectors alone. This usage is significant and could be optimized.

## Context
![image](https://github.com/bytebeamio/rumqtt/assets/17534654/54f0b562-2573-4722-ba4d-754b221205c7)

* Operating System: Ubuntu 22.04.1 aarch64
* Toolchain version: rustc 1.78.0

# Proposed Solution
Reduces memory consumption by using bitsets instead of vectors.

PR #869, opened 2024-05-21
refactor(rumqttc): Replace Vec with FixedBitSet for QoS 2 packet trac…

hippalus:
…king.(#868)

## Type of change

- Refactor (code improvements and optimizations)
- Replaced `Vec<Option<u16>>` with `FixedBitSet` for `outgoing_rel` and `incoming_pub`.
- Updated methods to accommodate the new data structure, including `clean`, `handle_incoming_pubrec`, `handle_incoming_pubrel`, and others.
- Updated unit tests to ensure correct functionality with the new data structure.
refactor(mqtt_state): Replace Vec with FixedBitSet for QoS 2 packet tracking (#868)

Refactored the `MqttState` struct to use `FixedBitSet` instead of `Vec<Option<u16>>` for the `outgoing_rel` and `incoming_pub` fields. This change optimizes memory usage and improves performance for tracking packet identifiers in QoS 2 operations.
 

Before
Screenshot 2024-05-06 at 15 14 22

After
Screenshot 2024-05-21 at 17 07 01

## Checklist:

### DIFF
--- a/rumqttc/src/state.rs
+++ b/rumqttc/src/state.rs
@@ -1,7 +1,7 @@
 use crate::{Event, Incoming, Outgoing, Request};
 
 use crate::mqttbytes::v4::*;
-use crate::mqttbytes::*;
+use crate::mqttbytes::{self, *};
 use fixedbitset::FixedBitSet;
 use std::collections::VecDeque;
 use std::{io, time::Instant};
@@ -29,7 +29,7 @@ pub enum StateError {
     #[error("A Subscribe packet must contain atleast one filter")]
     EmptySubscription,
     #[error("Mqtt serialization/deserialization error: {0}")]
-    Deserialization(#[from] Error),
+    Deserialization(#[from] mqttbytes::Error),
     #[error("Connection closed by peer abruptly")]
     ConnectionAborted,
 }
--- a/rumqttc/src/state.rs
+++ b/rumqttc/src/state.rs
@@ -1,7 +1,8 @@
 use crate::{Event, Incoming, Outgoing, Request};
 
 use crate::mqttbytes::v4::*;
-use crate::mqttbytes::{self, *};
+use crate::mqttbytes::*;
+use fixedbitset::FixedBitSet;
 use std::collections::VecDeque;
 use std::{io, time::Instant};
 
@@ -28,7 +29,7 @@ pub enum StateError {
     #[error("A Subscribe packet must contain atleast one filter")]
     EmptySubscription,
     #[error("Mqtt serialization/deserialization error: {0}")]
-    Deserialization(#[from] mqttbytes::Error),
+    Deserialization(#[from] Error),
     #[error("Connection closed by peer abruptly")]
     ConnectionAborted,
 }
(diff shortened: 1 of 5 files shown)


hippalus:
Before
Screenshot 2024-05-06 at 15 14 22

After
Screenshot 2024-05-21 at 17 07 01

rumqttc/src/state.rs
```diff
@@ -1,7 +1,8 @@
 use crate::{Event, Incoming, Outgoing, Request};
 
 use crate::mqttbytes::v4::*;
-use crate::mqttbytes::{self, *};
+use crate::mqttbytes::*;
```
de-sh, line 4: This change seems to be unrelated to the core of this PR

rumqttc/src/state.rs
```diff
@@ -28,7 +29,7 @@ pub enum StateError {
     #[error("A Subscribe packet must contain atleast one filter")]
     EmptySubscription,
     #[error("Mqtt serialization/deserialization error: {0}")]
-    Deserialization(#[from] mqttbytes::Error),
+    Deserialization(#[from] Error),
```
de-sh, line 32: same as above

de-sh (changes requested):
Thanks for the PR, these are some really good changes, would appreciate if you could undo the unrelated ones and apply similar updates to `v5/state.rs`. Please update `CHANGELOG.md` as well.

rumqttc/src/state.rs
de-sh: We have placed things this way to make it clear where that particular error is trickling down from.

de-sh (approved):
LGTM 
Great first PR @hippalus! 🎊 

edit: sorry, forgot to mention that we need an update to the `rumqttc/CHANGELOG.md` file

Arend-Jan:
now that is what I call an improvement. Thanks @hippalus 🙏

xiaocq2001:
Reduce the pkid to fixed bit seems work, but it assumes the pkids from host is within 1 to max inflight. It seems in MQTT spec there is no rule to limit packet identifiers in this way.
I think there could be issue to handle if a broker sending randomized or very large pkids, or scattered pkids.

de-sh:
> I think there could be issue to handle if a broker sending randomized or very large pkids, or scattered pkids, e.g., the bit to pkid value issue, the pkid actual ordering issue.

Is this with regards to v4/5?

swanandx:
> Is this with regards to v4/5?
>
both. I think what @xiaocq2001 wants to say is let's say we have max inflight set to 100, but nothing is stopping broker from using 200 as pkid right? 

so now when we get incoming pub with pkid 200, we will try to insert it and crash! please correct me if wrong.

btw, if this is the issue, then i think it was still there before this PR.

xiaocq2001:
There is another while re-sending the PUBREL packets, as described in MQTTv5 spec, "The Client MUST send PUBREC packets in the order in which the corresponding PUBLISH packets were received (QoS 2 messages) [MQTT-4.6.0-3]", this means if a connection is resumed the previous order of outgoing PUBREL must be restored, but `FixedBitSet` of `outgoing_ref` does not support ordering other than continuous and increasing pkid.

E.g., if broker sending packets with pkids 4,1,2,3 the expected `outgoing_ref` should be 4,1,2,3, but currently if it's disconnected and reconnected before the `outgoing_ref` sent, the resumed packet ordering is 1,2,3,4 which is not expected.

Another consideration is, for outgoing PUBLISH and PUBREL packets, they may need more information for status notification (https://github.com/bytebeamio/rumqtt/issues/805), in such cases we must use larger structs instead of just one bit for each packet.

Issue #871, opened 2024-05-24
Option to set the `tcp-nodelay` flag used for the connection from the MQTT client to the broker


FedorSmirnov89:
# Expected Behavior

It would be very useful to be able to specify whether the MQTT Client (in my case, I am using the `AsyncClient`) sets the `nodelay` flag when establishing the TCP connection to the broker.

# Current Behavior

As far as I can tell, (a) there is no option to configure whether the client sets the `nodelay` flag and (b) the client uses tokio's `TcpSocket`, which by default does not set the `nodelay` flag, so that it is not possible for the user to implement a client which does not batch TCP packets when sending them to the broker.

# Note

I would be happy to contribute to addressing this issue. I have submitted a PR (#872) with an extension of the `NetworkOptions` which would enable users to directly configure the `nodelay`, along with an example how this could be used.

PR #872, opened 2024-05-24
Option to configure TCP no_delay

FedorSmirnov89:
- extended `NetworkOptions` by an optional `nodelay` flag
- the flag is used when setting up the socket used for the tcp connection
- provided an example of how the async client can be configured to set `nodelay`

## Type of change

## Description

This pull request adds an option to set the nodelay flag for the TCP connection in the MQTT client. The nodelay flag, when enabled, disables Nagle's algorithm, which can help reduce latency by sending packets immediately without waiting for a full buffer. This is particularly useful in scenarios where low latency is critical.

### Changes
- Added a new configuration option to the MQTT client to enable or disable the nodelay flag for the TCP connection to the broker.
- Updated the provided examples to include instructions on how to use this new option.

### Issue

- addresses Issue #871 

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -14,12 +14,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `ConnectionAborted` variant on `StateError` type to denote abrupt end to a connection
 * `set_session_expiry_interval` and `session_expiry_interval` methods on `MqttOptions`.
 * `Auth` packet as per MQTT5 standards
+* `tcp_nodelay` field on `NetworkOptions` type to enable configuring `tcp_nodelay` for the client
+* `set_tcp_nodelay` method on `Networkoptions`
 
 ### Changed
 
 * rename `N` as `AsyncReadWrite` to describe usage.
 * use `Framed` to encode/decode MQTT packets.
 * use `Login` to store credentials
+* check `tcp_nodelay` field of `NetworkOptions` when setting up the connection socket to set tcp_nodelay if the flag is set
 
 ### Deprecated
 
--- a/rumqttc/examples/asyncpubsub_nodelay.rs
+++ b/rumqttc/examples/asyncpubsub_nodelay.rs
@@ -1,0 +1,57 @@
+//! Same as asyncpubsub.rs but extended by configuring the client to not batch the messages
+//! sent to the broker over TCP
+
+use tokio::{task, time};
+
+use rumqttc::{AsyncClient, MqttOptions, QoS};
+use std::error::Error;
+use std::time::Duration;
+
+#[tokio::main(flavor = "current_thread")]
+async fn main() -> Result<(), Box<dyn Error>> {
+    pretty_env_logger::init();
+    // color_backtrace::install();
+
+    let mut mqttoptions = MqttOptions::new("test-1", "localhost", 1883);
+    mqttoptions.set_keep_alive(Duration::from_secs(5));
+
+    let (client, mut eventloop) = AsyncClient::new(mqttoptions, 10);
+    // Configuring the client to not batch the messages sent to the broker over TCP
+    eventloop.network_options.set_tcp_nodelay(true);
+
+    task::spawn(async move {
+        requests(client).await;
+        time::sleep(Duration::from_secs(3)).await;
+    });
+
+    loop {
+        let event = eventloop.poll().await;
+        match &event {
+            Ok(v) => {
+                println!("Event = {v:?}");
+            }
+            Err(e) => {
+                println!("Error = {e:?}");
+                return Ok(());
+            }
+        }
+    }
+}
+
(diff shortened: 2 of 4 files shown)


rumqttc/src/lib.rs
```diff
@@ -369,6 +369,7 @@ impl From<ClientConfig> for TlsConfiguration {
 pub struct NetworkOptions {
     tcp_send_buffer_size: Option<u32>,
     tcp_recv_buffer_size: Option<u32>,
+    tcp_nodelay: Option<bool>,
```
de-sh, line 372:
```suggestion
    tcp_nodelay: bool,
```
would default the value to false anyways...

rumqttc/src/eventloop.rs
```diff
@@ -317,6 +317,10 @@ pub(crate) async fn socket_connect(
             SocketAddr::V6(_) => TcpSocket::new_v6()?,
         };
 
+        if let Some(nodelay) = network_options.tcp_nodelay {
+            socket.set_nodelay(nodelay)?;
+        }
```
de-sh, line 322:
```suggestion
        socket.set_nodelay(network_options.tcp_nodelay)?;
```

rumqttc/src/lib.rs
```diff
@@ -379,12 +380,17 @@ impl NetworkOptions {
...
+            tcp_nodelay: None,
             conn_timeout: 5,
             #[cfg(any(target_os = "android", target_os = "fuchsia", target_os = "linux"))]
             bind_device: None,
         }
     }
 
+    pub fn set_tcp_nodelay(&mut self, nodelay: bool) {
+        self.tcp_nodelay = Some(nodelay);
+    }
```
de-sh, line 392:
```suggestion
    pub fn set_tcp_nodelay(&mut self, nodelay: bool) {
        self.tcp_nodelay = nodelay;
    }
```

rumqttc/src/lib.rs
```diff
@@ -379,12 +380,17 @@ impl NetworkOptions {
         NetworkOptions {
             tcp_send_buffer_size: None,
             tcp_recv_buffer_size: None,
+            tcp_nodelay: None,
```
de-sh, line 383:
```suggestion
            tcp_nodelay: false,
```

rumqttc/examples/asyncpubsub_nodelay.rs
de-sh, line 1: I don't think an example is necessary to describe each and every option... can we go without this?

rumqttc/CHANGELOG.md
```diff
@@ -14,12 +14,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 * `Auth` packet as per MQTT5 standards
+* `tcp_nodelay` field on `NetworkOptions` type to enable configuring `tcp_nodelay` for the client
+* `set_tcp_nodelay` method on `Networkoptions`
 
 ### Changed
 
 * rename `N` as `AsyncReadWrite` to describe usage.
 * use `Framed` to encode/decode MQTT packets.
 * use `Login` to store credentials
+* check `tcp_nodelay` field of `NetworkOptions` when setting up the connection socket to set tcp_nodelay if the flag is set
```
de-sh, line 25:
```suggestion
* Allow configuring  the `nodelay` property of underlying TCP client with the `tcp_nodelay` field in `NetworkOptions`

### Changed

* rename `N` as `AsyncReadWrite` to describe usage.
* use `Framed` to encode/decode MQTT packets.
* use `Login` to store credentials
```

de-sh (changes requested):
Just a few suggestions...

rumqttc/examples/asyncpubsub_nodelay.rs
FedorSmirnov89: done :)

swanandx (approved):
PR look good, thanks for the contribution!

just one question before merging:
have you tried comparing the actual performance difference with & without this flag? because if latency is bottle-necked by eventloop or some other thing, this might not be that boost right? please correct me if wrong

thanks :)

de-sh (approved):

de-sh:
Congrats on the merge @FedorSmirnov89 🎊

FedorSmirnov89:
Thank you for merging :) . Just to add the information @swanandx was asking about: It's exactly as @de-sh said: We have a distributed application where two clients on different nodes exchange messages. We measured the timing of the messages since we want them to have as little latency and jitter as possible. They are exchanging very small messages on a high frequency  (range of 1-4 ms).

Seeing that configuring the broker (we are using the normal Mosquitto which comes with an `apt-get` in Linux) to set no-delay for the messages it is sending significantly reduced the jitter, we did the same thing on the client side by building our agents of a local branch of `rumqttc` where we had the no-delay option enabled. Again, we saw a significant decrease in message jitter.

Thanks a lot for reviewing and merging :) . We can now go back again to building off the main branch of the repository which is great 😄

PR #874, opened 2024-06-03
Disable default feature for `metrics-exporter-prometheus`

Dirreke:
Disable default feature for `metrics-exporter-prometheus`, because we do not use the feature `push-gateway`, which introduce `native-tls` to the lib.


de-sh (approved):

de-sh:
Happy first merge! 🥳

Issue #875, opened 2024-06-05
Unexpected Reconnect on MqttState errors


vaavva:
# Expected Behavior

UnsubFail/PubAckFail/etc should not act as a disconnect as they are packet specific and fully recoverable.

# Current Behavior

In the event loop, anything that is an Err, not Ok calls self.clean() https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/v5/eventloop.rs#L168, which should only be called when the client is disconnected (https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/v5/eventloop.rs#L120)

However, ConnectionError includes MqttState errors (https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/v5/eventloop.rs#L38), which includes things like EmptySubscription, OutgoingPacketTooLarge, UnsubFail, PubAckFail, etc https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/v5/state.rs#L38 . So any of these errors will act as a disconnect it seems. 

# Failure Information (for bugs)

Please help provide information about the failure if this is a bug. If it is not a bug, please remove the rest of this template.

## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Operating System: WSL2
* Toolchain version: rumqttc = "0.24.0"

## Failure Logs

On duplicate unsubs (second should return a failure rc with NoSubscriptionExisted), which is shown from the event loop, but a reconnect also occurs
```
[MQTT] Outbound Unsubscribe: 3
[MQTT] Outbound Unsubscribe: 4
[MQTT] Incoming UNSUBACK: UnsubAck { pkid: 3, reasons: [Success], properties: None }
[MQTT] Unsub Fail: NoSubscriptionExisted
[MQTT] Connected!
[MQTT] Incoming CONNACK: ConnAck
...
```
mosquitto broker logs:
```
1717600422: Received UNSUBSCRIBE from sample_client
1717600422:     foo_topic/bar
1717600422: sample_client foo_topic/bar
1717600422: Sending UNSUBACK to sample_client
1717600422: Received UNSUBSCRIBE from sample_client
1717600422:     foo_topic/bar
1717600422: sample_client foo_topic/bar
1717600422: Sending UNSUBACK to sample_client
1717600422: Client sample_client closed its connection.
1717600422: New connection from 127.0.0.1:54500 on port 1884.
```

For a pub on an unauthorized topic
```
[...]

PR #877, opened 2024-06-12
Remove SubAck/UnsubAck/PubAck/PubRec/PubRel/PubComp reason code errors from critical state error

xiaocq2001:
## Type of change

Bug fix (non-breaking change which fixes an issue): https://github.com/bytebeamio/rumqtt/issues/875
In v5, filtered error cases so ACK fail cases are not disconnecting.

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -35,7 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Ordering of `State.events` related to `QoS > 0` publishes
 * Filter PUBACK in pending save requests to fix unexpected PUBACK sent to reconnected broker.
 * Resume session only if broker sends `CONNACK` with `session_present == 1`.
-* Remove v5 PubAck/PubRec/PubRel/PubComp/Sub/Unsub failures from `StateError`.
+* Remove v5 PubAck/PubRec/PubRel/PubComp/Sub/Unsub failures from `StateError` and log warnings on these failures.
 
 ### Security
 
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Ordering of `State.events` related to `QoS > 0` publishes
 * Filter PUBACK in pending save requests to fix unexpected PUBACK sent to reconnected broker.
 * Resume session only if broker sends `CONNACK` with `session_present == 1`.
+* No disconnection on v5 PubAck/PubRec/PubRel/PubComp/Sub/Unsub failures.
 
 ### Security
 
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -35,7 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Ordering of `State.events` related to `QoS > 0` publishes
 * Filter PUBACK in pending save requests to fix unexpected PUBACK sent to reconnected broker.
 * Resume session only if broker sends `CONNACK` with `session_present == 1`.
-* No disconnection on v5 PubAck/PubRec/PubRel/PubComp/Sub/Unsub failures.
+* Remove v5 PubAck/PubRec/PubRel/PubComp/Sub/Unsub failures from `StateError`.
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


rumqttc/src/v5/eventloop.rs
```diff
@@ -168,7 +168,19 @@ impl EventLoop {
...
                 // Move pending messages from state to eventloop.
-                self.clean();
+                // The PubAckFail/PubRecFail/PubRelFail/PubCompFail/SubFail/UnsubFail errors are not critical errors.
+                match e {
+                    ConnectionError::MqttState(StateError::PubAckFail { .. })
+                    | ConnectionError::MqttState(StateError::PubRecFail { .. })
+                    | ConnectionError::MqttState(StateError::PubRelFail { .. })
+                    | ConnectionError::MqttState(StateError::PubCompFail { .. })
+                    | ConnectionError::MqttState(StateError::SubFail { .. })
+                    | ConnectionError::MqttState(StateError::UnsubFail { .. }) => {}
```
swanandx, line 178: instead of matching them here and ignoring, we can just remove those variants from `StateError` right?

i think that will be much better. wdyt?

xiaocq2001: Sounds not bad.
Considering the acknowledgement is finished then, `handle_incoming_...ack` could return `Ok(None)`, since there should not be any outgoing packet. `Err(...)` is not used because there is no state error.

swanandx: other logic will remain same ( as there can be collisions and unsolicited / OOO acks ). Only the part where we check for status code ( and return error lets say `PubAckFail` ) will be removed.

xiaocq2001: Updated code.
Keep some branch for warning logs.

rumqttc/src/v5/state.rs
```diff
@@ -429,9 +417,8 @@ impl MqttState {
         self.incoming_pub.set(pubrel.pkid as usize, false);
 
         if pubrel.reason != PubRelReason::Success {
-            return Err(StateError::PubRelFail {
-                reason: pubrel.reason,
-            });
+            warn!("PubRel Pkid = {:?}, reason: {:?}", pubrel.pkid, pubrel.reason);
+            return Ok(None);
```
swanandx, line 421: we do not need to complete futher qos 2 process if reason is failure? 

( same question for other acks as well ) we do not need to pop something from inflight on ack?

I tried to read specs but couldn't find clear answer, can you please check it once and verify if behaviour is correct? Thanks!

xiaocq2001: I think the answer is yes, for different ACKs the checks may be different.
But I think we should open another issue to check those.
Here I just follow current logic, replacing the `Err(...)` with `Ok(None)`.

swanandx (commented):
LGTM! just some minor changes.

PS: I don't think logging `warn!` is correct in case of failure reason code ( because i think it is upto application to determine how to treat reason code ). But let's keep it for now for smooth transition from returning error to logging warning. Can you also add this under changed section of changelog that we will be logging warnings?

Thanks for the PR!

rumqttc/CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Ordering of `State.events` related to `QoS > 0` publishes
 * Filter PUBACK in pending save requests to fix unexpected PUBACK sent to reconnected broker.
 * Resume session only if broker sends `CONNACK` with `session_present == 1`.
+* No disconnection on v5 PubAck/PubRec/PubRel/PubComp/Sub/Unsub failures.
```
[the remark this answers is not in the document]
xiaocq2001: Fixed.

rumqttc/src/v5/eventloop.rs
```diff
@@ -168,6 +168,8 @@ impl EventLoop {
             Err(e) => {
                 // MQTT requires that packets pending acknowledgement should be republished on session resume.
                 // Move pending messages from state to eventloop.
+                // The PubAckFail/PubRecFail/PubRelFail/PubCompFail/SubFail/UnsubFail errors are not critical errors.
+                warn!("Disconnect due to a critical error: {:?}", e);
```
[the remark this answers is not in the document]
xiaocq2001: Removed.

xiaocq2001:
Updated, please check.

swanandx (approved):

swanandx:
Thank you so much for your patience and contribution!

PR #878, opened 2024-06-13
Graceful shutdown

silvestrpredko:
## Type of change

## Checklist:


silvestrpredko:
Short description:
I wanna make a graceful shutdown without a significant redesign of the system. For me `watch` seems like a good solution. Please don't hesitate to suggest improvements to the initial design.

swanandx:
i also made a poc for graceful shutdown using cancellation token, you can check it here https://github.com/bytebeamio/rumqtt/compare/main...rumqttd-shutdown !

we can compare both approaches and combine good stuff from both :)

silvestrpredko:
> i also made a poc for graceful shutdown using cancellation token, you can check it here [main...rumqttd-shutdown](https://github.com/bytebeamio/rumqtt/compare/main...rumqttd-shutdown) !
> 
> we can compare both approaches and combine good stuff from both :)

I like that handle returns in `start`. The reason why I was ignoring `CancellationToken` it's that `tokio-utils` is an optional dependency and I just don't want to change it. So it's more up to you what changes you would like to make, and what it should be used at the end. Please write below your decision.

silvestrpredko:
Hey @swanandx. I redesigned the approach a little bit. It seems like a good solution. Please check it out.

 > P.S I also added `join` API call for situation when you don't want to handle a graceful shutdown

silvestrpredko:
Hey, @swanandx do you have a chance to look at this PR?

silvestrpredko:
@swanandx Just I kindly reminder. Could you please take a look?

baxterjo:
@swanandx Any updates on this? Is there another member that can take over the review?

rumqttd/src/server/broker.rs
```diff
@@ -155,8 +156,12 @@ impl Broker {
         Ok((link_tx, link_rx))
     }
 
+    /// Starts the MQTT broker server and spawns various components like the router, cluster, metrics, and console.
+    ///
+    /// This function sets up the necessary components for the MQTT broker server to run. The function returns a
+    /// [`BrokerHandler`] that can be used to gracefully shut down the server.
     #[tracing::instrument(skip(self))]
-    pub fn start(&mut self) -> Result<(), Error> {
+    pub fn start(self) -> Result<BrokerHandler, Error> {
```
baxterjo, line 164: This would make me weary about breaking current functionality. Current start function does not return, so those currently using broker start and expecting it to go forever could potentially exit their programs prematurely with this change.

silvestrpredko: Unfortunately, when I made this change I didn't think about breaking of current API. In my case, I only need something that will stop the broker when it goes out of execution scope(tests).

I think that it would be better to control how far execution goes by a handler, rather than spawn it and then block execution till it exits. Maybe could be implemented something like that https://github.com/tokio-rs/axum/blob/main/examples/graceful-shutdown/src/main.rs

baxterjo: No worries. Sorry if I'm stepping on your toes but I submitted https://github.com/bytebeamio/rumqtt/pull/929 last week so I could keep moving on my work. I think an ideal future state would be to make a new `start_async()` method for the `Broker` that would look lot what you've built here.

silvestrpredko: No problem, I just hope that one of our PRs will be merged.

PR #880, opened 2024-06-17
Store outgoing_pub in a HashMap for more memory efficient storage of unassigned pkids

ambaxter:
## Type of change

## Checklist:

I wrote an MQTT traffic simulator and noticed that creating many clients within a single process consumes a large amount of memory. The below example is for 1000 clients.

mqtt_before

This is due to rumqttc pre-allocating `MqttState.outgoing_pub` for each client. I replaced the Vec with a HashMap and dramatically reduced memory usage.

mqtt_after

I know that `MqttState.outgoing_pub` is most likely pre-allocated for low powered devices. Would it be possible to make using a HashMap either a runtime configuration or feature flag?

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * use `Login` to store credentials
 * Made `DisconnectProperties` struct public.
 * Replace `Vec<Option<u16>>` with `FixedBitSet` for managing packet ids of released QoS 2 publishes and incoming QoS 2 publishes in `MqttState`.
+* Replace `Vec<Option<Publish>>` with `HashMap` for storing outgoing pub packet ids in `MqttState`. This dramatically reduces memory usage in processes with a large number of clients.
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


de-sh:
By any chance does the code with which you tested use QoS 1+?

Seems like the HashMap is saving space as it never gets utilized(which is the case in QoS 0). That shouldn't be the case when we end up using QoS 1+ where allocations will take up some cycles as the map size increases slowly.

swanandx:
you can also configure how much you want to preallocate using MqttOptions, let's say you set max_inflight to be 10 to save memory, it won't allocate more than that.

xiaocq2001:
IMHO we can use `VecDeque` to save `pkid/publish` and `NoticeTx`, the reasons:
1. Since the packets are always ordered, the `push_back` and `pop_front` can be used, they work faster than `insert` and `push` of Vec.
2. The memory usage is less than HashMap

Some of the proposed changes are included here:
https://github.com/bytebeamio/rumqtt/commit/68fcb73b1d736449626e39d5528d74dbad82287e#diff-35dca8779a1d08a58f298e7f88a1ca9d2c32f79e97024252a6366e4a2f32325a

PR #742, opened 2023-10-31
Add support for `native-tls` when using websocket transport

ondrowan:
This PR adds support for `native-tls` when using `Transport::Wss`.

This is still a WIP since the feature `tokio-native-tls` needs to be set in `async-tungstenite` when using `use-native-tls` feature in this crate. At the same time, if `use-rustls` is set in this crate, `async-tungstenite` needs to set feature `tokio-rustls-native-certs`. I have no idea how to edit `Cargo.toml` to support both of these cases.

At the moment `async-tungstenite` always uses - even if using insecure WS connection - `tokio-rustls-native-certs` feature, which I don't think is correct either.

## Type of change

New feature (non-breaking change which adds functionality)

## Checklist:

- [x] Formatted with `cargo fmt`
- [ ] Make an entry to `CHANGELOG.md` if it's relevant to the users of the library. If it's not relevant mention why.

PR #881, opened 2024-06-18
Add native tls WSS support.

xiaocq2001:
## Type of change

* Added native_tls websocket support

## Checklist:

### DIFF
--- a/rumqttc/Cargo.toml
+++ b/rumqttc/Cargo.toml
@@ -17,8 +17,8 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = ["use-rustls"]
-use-rustls = ["dep:tokio-rustls", "dep:rustls-webpki", "dep:rustls-pemfile", "dep:rustls-native-certs"]
-use-native-tls = ["dep:tokio-native-tls", "dep:native-tls"]
+use-rustls = ["dep:tokio-rustls", "dep:rustls-webpki", "dep:rustls-pemfile", "dep:rustls-native-certs", "async-tungstenite/tokio-rustls-native-certs"]
+use-native-tls = ["dep:tokio-native-tls", "dep:native-tls", "async-tungstenite/tokio-native-tls"]
 websocket = ["dep:async-tungstenite", "dep:ws_stream_tungstenite", "dep:http"]
 proxy = ["dep:async-http-proxy"]
 
@@ -83,3 +83,8 @@ required-features = ["websocket"]
 name = "websocket_proxy"
 path = "examples/websocket_proxy.rs"
 required-features = ["websocket", "proxy"]
+
+[[example]]
+name = "wss"
+path = "examples/wss.rs"
+required-features = ["websocket"]
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Made `DisconnectProperties` struct public.
 * Replace `Vec<Option<u16>>` with `FixedBitSet` for managing packet ids of released QoS 2 publishes and incoming QoS 2 publishes in `MqttState`.
 * Accept `native_tls::TlsConnector` as input for `Transport::tls_with_config`.
+* Accept `native_tls::TlsConnector` as input for `Transport::wss_with_config`.
 
 ### Deprecated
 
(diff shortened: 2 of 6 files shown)


swanandx:
there is PR #742 which tackles same problem. Last time when I tried we faced some issues ( mentioned in discussion there ).

here as this PR says "use-native-tls" overwrites "use-rustls", I think it would be better to not overwrite and let users choose what they want to use through TlsConfig, wdyt?

xiaocq2001:
> there is PR #742 which tackles same problem. Last time when I tried we faced some issues ( mentioned in discussion there ).
> 
> here as this PR says "use-native-tls" overwrites "use-rustls", I think it would be better to not overwrite and let users choose what they want to use through TlsConfig, wdyt?

Do you mean through runtime check instead of through `cfg` like following:

```
            let connector = match &tls_config {
                #[cfg(feature = "use-native-tls")]
                TlsConfiguration::SimpleNative { .. } | TlsConfiguration::Native | TlsConfiguration::NativeConnector(_) => {
                    tls::native_tls_connector(&tls_config).await?
                }
                TlsConfiguration::Simple { .. } | TlsConfiguration::Rustls(_) => {
                    tls::rustls_connector(&tls_config).await?
                }
                #[allow(unreachable_patterns)]
                _ => unreachable!("This cannot be called for other TLS backends than native-tls and rustls"),
            };
```

xiaocq2001:
Actually it seems if in toml the `use-rustls` is in default section, there is no way to remove it, even I uses `--no-default-features`.

```
cargo run --example wss --no-default-features --features="websocket"
    Finished dev [unoptimized] target(s) in 0.16s
     Running `target/debug/examples/wss`
use-native-tls: false
use-rustls    : true
```

That's why I try to configure `use-native-tls` overwrites `use-rustls`.

xiaocq2001:
I tried to enable just based on feature, the following test fails:
```
cargo clippy --manifest-path rumqttc/Cargo.toml --all-features`
```
Is there a way to keep these two features mutually exclusive?

swanandx:
> Is there a way to keep these two features mutually exclusive?
>

we can. but it's better to make them additive. i.e. being able to turn on both without any issues.

xiaocq2001:
> > Is there a way to keep these two features mutually exclusive?
> 
> we can. but it's better to make them additive. i.e. being able to turn on both without any issues.

Currently we are using `async-tungstenite`, where different TLS features are mutually exclusive ([async-tungstenite/src/tokio.rs#L14C1-L27C9](https://github.com/sdroege/async-tungstenite/blob/c7b412e2221dea57cd6472436b46b503a44e2d74/src/tokio.rs#L14C1-L27C9)). Where `rustls` only works when `native-tls` is not enabled. 

IMHO using both TLS implement at the same time could lead to conflicts and unexpected behavior, as they are intended to be alternative options to each other.

rumqttc/Cargo.toml
```diff
@@ -18,7 +18,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 [features]
 default = ["use-rustls"]
 use-rustls = ["dep:tokio-rustls", "dep:rustls-webpki", "dep:rustls-pemfile", "dep:rustls-native-certs"]
-use-native-tls = ["dep:tokio-native-tls", "dep:native-tls"]
+use-native-tls = ["dep:tokio-native-tls", "dep:native-tls", "async-tungstenite/tokio-native-tls"]
```
xiaocq2001, line 21: `async-tungstenite/tokio-native-tls` is required by `websocket` ... may I introduce `websocket-native-tls` for native-tls support to avoid such dependency in `use-native-tls`?

xiaocq2001: That's a bad idea, different feature may conflict, forget it.

xiaocq2001:
Any comment?

rumqttc/Cargo.toml
xiaocq2001: Just saw previous discussion in https://github.com/bytebeamio/rumqtt/pull/742, `?` seems to be good.

xiaocq2001:
Any comments?

xiaocq2001:
@swanandx , can we offer the functionality first align with `async-tungstenite`, and update later when it's updated or there is better lib with additive feature configuration?

xiaocq2001:
Hi, is there any chance to have the feature and leave improvements for additive configuration support in the future?

xiaocq2001:
Any comments?

PR #885, opened 2024-07-04
refactor(rumqttc): make AsyncClient/Eventloop mockable

mbodmer:
I have the need to mock the interface to the rumqttc client for testing/benching purposes. 

There was no evident way I could achieve this other than introducing a new trait `PollableEventLoop` defining the poll interface on the `EventLoop`.

Now it is possible to reuse the `AsyncClient` and mock the `EventLoop`.

I have also added a mocked version of the `asyncpubsub` example.

This change will make it mandatory to `use rumqttc::PollableEventLoop`, that is why I mark this as a breaking change.

## Type of change

## Checklist:

### DIFF
--- a/benchmarks/clients/rumqttasync.rs
+++ b/benchmarks/clients/rumqttasync.rs
@@ -1,4 +1,4 @@
-use rumqttc::{AsyncClient, Event, Incoming, MqttOptions, QoS};
+use rumqttc::{AsyncClient, Event, Incoming, MqttOptions, PollableEventLoop, QoS};
 
 use std::error::Error;
 use std::time::{Duration, Instant};
--- a/benchmarks/clients/rumqttasyncqos0.rs
+++ b/benchmarks/clients/rumqttasyncqos0.rs
@@ -1,4 +1,4 @@
-use rumqttc::{AsyncClient, Event, Incoming, MqttOptions, QoS};
+use rumqttc::{AsyncClient, Event, Incoming, MqttOptions, PollableEventLoop, QoS};
 
 use std::error::Error;
 use std::time::{Duration, Instant};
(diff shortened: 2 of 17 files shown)


biglben:
Great work. I'm also very interested in getting this merged as it would be incredibly helpful for our testing process. Looking forward to seeing this merged soon

swanandx:
Hey @mbodmer , thanks for the PR.

As per the example in PR, mock eventloop is just an wrapper and uses `AsyncClient::from_sender` internally, so I don't think we need the trait `PollableEventLoop`. For testing, all you need to do use use from_sender method and use the rx of that channel to do anything right? even recreating your own eventloop!

Please feel free to lmk if I'm missing something, thank you for understanding :)

mbodmer:
Thanks so much for your feedback @swanandx

I am not sure I completely understand your suggestion. Also the example might not completely capture my situation. My client code is something like:

```rust
pub struct MqttClientService<S> {
    pub(crate) client: AsyncClient,
    pub(crate) eventloop: EventLoop,
    pub(crate) inner: S,
}
```

Here I cannot make the `EventLoop` generic when it is not a trait. How would I integrate with a mocked eventloop or how would your suggestion allow for this? I am not that experienced with rust, so please excuse me, I probably did not understand yet.

swanandx:
let me elaborate: for mock eventlloop 
```rust
struct MockEventloop {
    rx: Receiver<Request>,
}

// impl MockEventloop as you wish

// To create client and eventloop you can do something like:
let (tx, rx) = flume::unbounded();
let client = AsyncClient::from_senders(tx);
let eventloop = MockEventloop { rx };
```

Coming to your code, you want to have Eventloop such that both rumqttc's Eventloop and your mock eventloop can integrate with your MqttClientService right?

For that, you can define and implement that trait inside your code as well right? for Eventloop you can just call it's poll internally, and for Mockeventloop you can do as you need. pseudocode:
```rust
impl Pollable for EventLoop {
    fn poll(..) {
        self.poll().await
    }
}
```

Otherwise, you can have MockMqttClient which holds the rx ( instead of eventloop ) which you can use for testing.

swanandx:
cool! feel free to reopen / comment if required, thanks :)

PR #896, opened 2024-07-30
feat: `LinkRx::recv_async`

de-sh:
## Checklist:

### DIFF
--- a/rumqttd/src/link/local.rs
+++ b/rumqttd/src/link/local.rs
@@ -338,6 +338,23 @@ impl LinkRx {
         }
     }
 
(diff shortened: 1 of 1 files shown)


rumqttd/src/link/local.rs
```diff
@@ -338,6 +338,23 @@ impl LinkRx {
         }
     }
 
+    pub async fn recv_async(&mut self) -> Result<Option<Notification>, LinkError> {
```
swanandx, line 341: there is already method named `next()` which is exactly the same!

PR #916, opened 2024-09-29
feat: return token when request is made, resolve token when request handling completes

de-sh:
Solves #805 

## Type of change

New feature, that notifies the requester when their requests are acknowledged by the broker(in the case of QoS 1/2 Publishes and Subscribe/Unsubscribe) or when they are written to TCP buffer(for all other requests).

**BREAKING:** [`pending`](https://github.com/bytebeamio/rumqtt/pull/916#discussion_r1783221572) definition changes from `VecDeque<Request>` to `VecDeque<(Request, Option<PromiseTx>)>`.

## Checklist:


xiaocq2001:
Glad to see the progress here. Is there any example on how the feature is used?

de-sh:
Thanks for the review, two more things I have been thinking about with respect to the interface are as follows:

rumqttc/src/eventloop.rs
```diff
@@ -75,11 +75,11 @@ pub struct EventLoop {
     /// Current state of the connection
     pub state: MqttState,
     /// Request stream
-    requests_rx: Receiver<Request>,
+    requests_rx: Receiver<(Request, Option<PromiseTx>)>,
```
de-sh, line 78:
```suggestion
    requests_rx: Receiver<PendingRequest>,
```

rumqttc/src/eventloop.rs
```diff
@@ -75,11 +75,11 @@ pub struct EventLoop {
...
     /// Requests handle to send requests
-    pub(crate) requests_tx: Sender<Request>,
+    pub(crate) requests_tx: Sender<(Request, Option<PromiseTx>)>,
     /// Pending packets from last session
-    pub pending: VecDeque<Request>,
+    pub pending: VecDeque<(Request, Option<PromiseTx>)>,
```
de-sh, line 82: This is a breaking change, note on release

rumqttc/src/eventloop.rs
de-sh: Resolved in https://github.com/bytebeamio/rumqtt/pull/917

xiaocq2001:
> Thanks for the review, two more things I have been thinking about with respect to the interface are as follows:
> 
> 1. Returning pkid instead of `()`.
> 2. Errors with more context about why a request was refused by the broker, both subscribe and unsub acks have reason codes as response. Should this just return the acknowledgement packets received from the broker or repackage the same into a more presentable type?

For 1, that's great. That also helps on 2, before we decide to put more things in error cases, the pkid helps to extract the error packet from event loop events.
For 2, I vote the acknowledgement packets, since that's not only the reason code, but also with reason string and maybe other user defined properties.

rumqttc/examples/ack_promise_sync.rs
```diff
@@ -0,0 +1,117 @@
...
+        .blocking_wait()
+    {
+        Ok(pkid) => println!("Acknowledged Pub({pkid})"),
+        Err(e) => println!("Publish failed: {e:?}"),
+    }
+
+    // Spawn threads for each publish, use channel to notify result
+    let (tx, rx) = bounded(1);
+
+    let future = client
```
tekjar, line 69: Rename these to tokens instead of futures

izolyomi:
Thank you for the efforts, I'm also interested in using this once it's merged.

Leandros:
Any way that one can help to get this merged? 

Thank you!

swanandx:
just an heads up: this will be merged and released with rumqttc upcoming version by / in January!

thank you so much for your patience and contribution 💯

CosminPerRam:
Bump, any news regarding merging this?

izolyomi:
Do you maybe have any updates like a current schedule or planned deadline for a merge and release of this feature? Thank you for the efforts.

tekjar:
Hi @izolyomi . We are unfortunately delayed on this. We'll try to focus on these issues in Feb and cut a new release

cartertinney:
I checked this out and it works quite well, aside from a handful of issues:

1) The reason for rejection by the broker that are reported via the Token cannot seem to be matched upon, or interacted with in any meaningful way. It would be nice if I could write conditional logic the respond to particular rejections in different ways. Perhaps I missed something?

2) If a subscribe or unsubscribe is done when the client thinks it is connected, but is not and has just not timed out on the ping/keep alive yet, the Token will hang indefinitely - I am assuming is because, per MQTT spec, unacknowledged SUB and UNSUB are discarded upon connection loss/reconnect, but the discard does not seem to get reported to the Token, resulting in the token.await hanging forever.

3) When using QoS2 and manual ack for a received message (perhaps this also applies with auto ack), I'm seeing the Token.await return after the PUBREC is sent, without waiting for the rest of the PUBREL/PUBCOMP handshake

4) When using QoS1 or QoS2 and manual ack for a received message if there is a disconnect between receiving the message and sending the manual ack with .ack(), the PUBACK will queue up and be redelivered upon reconnect, which is a MQTT spec violation as it can lead to acking the wrong message, or duplicate acks in edge cases, or even broker disconnects. I assume this also applies to automatic acking, but it's very hard to reproduce this condition without manual ack.

I see that the release of this feature has been delayed - is there some issue with the implementation that I'm missing? Aside from the above issues I found, everything worked fantastic, and it would be extremely helpful to get this integrated into my codebase ASAP. Are there any updates?

de-sh:
>  checked this out and it works quite well, aside from a handful of issues:

Thanks for pointing these out @cartertinney, will figure out how to solve this over the weekend!

swanandx:
> #946 appears to have solved issues 1 and 3 that I raised above. Thanks much, I continue to look forward to fixes for 2 and 4.
> 
> To expand on issue 2, further testing exposed that this same hanging issue affecting `.subscribe()` and `.unsubscribe()` also appears when using manual acks with the `.ack()` method on QoS2 messages, which makes sense and follows from the original observations, but I just thought I should clarify.

for 2, if client is timed out on keep alive, the connection is closed and so token should resolve to `Disconnected` error! you are mentioning that this isn't working right?

for 3, we tested and it was working fine.

note that when using manual acks with qos 2 you just need to ack the incoming publish ( rest of the qos 2 flow is handled internally, i.e. pubrel/pubcomp hadshake is done internally ) and for outgoing publishes, token is resolved only after getting pubcomp!

[ shall we have manual acks for PubRec &/ PubRel as well? cc: @tekjar @de-sh ]

can you please provide a minimal poc so that we can reproduce the issue to fix it?

thanks!

cartertinney:
> > #946 appears to have solved issues 1 and 3 that I raised above. Thanks much, I continue to look forward to fixes for 2 and 4.
> > To expand on issue 2, further testing exposed that this same hanging issue affecting `.subscribe()` and `.unsubscribe()` also appears when using manual acks with the `.ack()` method on QoS2 messages, which makes sense and follows from the original observations, but I just thought I should clarify.
> 
> for 2, if client is timed out on keep alive, the connection is closed and so token should resolve to `Disconnected` error! you are mentioning that this isn't working right?
> 
> for 3, we tested and it was working fine.
> 
> note that when using manual acks with qos 2 you just need to ack the incoming publish ( rest of the qos 2 flow is handled internally, i.e. pubrel/pubcomp hadshake is done internally ) and for outgoing publishes, token is resolved only after getting pubcomp!
> 
> [ shall we have manual acks for PubRec &/ PubRel as well? cc: @tekjar @de-sh ]
> 
> can you please provide a minimal poc so that we can reproduce the issue to fix it?
> 
> thanks!

Yes, like I said, issues 1 and 3 appear to be solved. I'm happy with the way QoS2 handles the PUBREC/REL/COMP handshake, it is much improved from prior iterations - I personally don't think manual PUBCOMP has much value from a user perspective, I like the current solution.

As for 2, I can run the test again, perhaps I missed something, but you are correct - I'm not seeing the Token resolve to a Disconnected error, I'm seeing the same hanging behavior as before on Tokens created by SUB and UNSUB operations.

swanandx:
>  I'm not seeing the Token resolve to a Disconnected error, I'm seeing the same hanging behavior as before on Tokens created by SUB and UNSUB operations
>

my observations:
`TokenError::Disconnected` is returned when the sender ( say `tx` ) is dropped. At a given time tx can be at:
- in channel between client and eventloop
this is when we call the method ( publish/subscribe/etc ) and eventloop is yet to process that request.

- in state of eventloop
one that Request is processed and we are waiting for the Ack

Issue is that the **`tx` is never dropped** if we don't have network connection!

<details>

<summary>Minimal PoC</summary>

```rust
use rumqttc::{AsyncClient, MqttOptions, QoS};
use std::time::Duration;

#[tokio::main(flavor = "current_thread")]
async fn main() {
    let mqttoptions = MqttOptions::new("poc", "never-connect-here", 1883);

    // spawn eventloop poll in separate task
    tokio::task::spawn(async move {
        loop {
            let event = eventloop.poll().await;
            match &event {
                Ok(v) => {
                    println!("Event = {v:?}");
                }
                Err(e) => {
                    println!("Error = {e:?}");
                    tokio::time::sleep(Duration::from_secs(60)).await;
                }
            }
        }
    });

    let token = client
        .subscribe("hello/world", QoS::AtMostOnce)
        .await
        .unwrap();

    // as we never connect to network, we never process any requests in eventloop
    // that are pending in channel. Hence sender is never dropped and token is
    // never resolved!
    match token.await {
        Ok(pkid) => println!("Acknowledged Sub({pkid:?})"),
        Err(e) => println!("Subscription failed: {e:?}"),
    };
}
```
</details>

For keep alive timeouts / disconnects due to other errors:

packets that are sent, i.e. for handles in state, we should clear them em, pushed the commit https://github.com/bytebeamio/rumqtt/pull/916/commits/c7ec9b4a033e3d0e6dc3ad7d27e0c05731ffdbc6 for fixing this!


But for the ones that are in channel ( yet to reach the eventloop ). we don't intend to drop the **requests** which are in channel, which leads to that hanging behaviour as in above poc and mentioned above.

please comment on the above behaviour for suggestions! 

thanks!

swanandx:
> 4. When using QoS1 or QoS2 and manual ack for a received message if there is a disconnect between receiving the message and sending the manual ack with .ack(), the PUBACK will queue up and be redelivered upon reconnect, which is a MQTT spec violation as it can lead to acking the wrong message, or duplicate acks in edge cases, or even broker disconnects. I assume this also applies to automatic acking, but it's very hard to reproduce this condition without manual ack.
> 

we do not retransmit the pubacks though: https://github.com/bytebeamio/rumqtt/blob/cb89fcfb75b794b1f6438e33469df4cb9ed64f9c/rumqttc/src/eventloop.rs#L137 

can you verify again if i'm missing something or maybe something related to qos2?

rumqttc/src/lib.rs
```diff
@@ -145,6 +144,8 @@ use rustls_native_certs::load_native_certs;
 pub use state::{MqttState, StateError};
 #[cfg(any(feature = "use-rustls", feature = "use-native-tls"))]
 pub use tls::Error as TlsError;
+use tokens::Resolver;
```
joshuachp, line 147:
```suggestion
pub use tokens::Resolver;
```

Could we make the resolver public for testing?

cartertinney:
@de-sh @swanandx 

Trying out the latest version, I now see the expected behavior for SUB and UNSUB in the case of connection drops, I now get a `TokenError` indicating the "sender side of channel was dropped". This is the right behavior, although the error text is a little confusing (the connection was lost, the underlying channel or Event Loop wasn't dropped from memory and still exists, and if connection were to be regained things would again begin working normally). That quirk of semantics aside, this is a huge improvement for correctness, I am thrilled with this change.

Now, regarding the acknowledgement process:

I understand that the ACK requests that queue up while disconnected are in the channel (yet to reach the EventLoop/ MQTT session), but the issue is that allowing them to queue up like this violates the MQTT specification. Acking a message from a previous connection upon reconnect is dangerous because we can't guarantee we are acknowledging the packet we think we are. Consider the case:

A publish is sent to the client from the broker. After receiving the publish, but before sending the PUBACK, the client loses connection. The PUBACK is queued up in the channel. Eventually, the client reconnects. The broker redelivers the original publish, now with the DUP flag set to true. The queued up PUBACK the client had in the channel is also delivered upon reconnect. This PUBACK was intended to be sent for the original copy of the message on a prior connection, but now acks the redelivery from the broker's perspective. The broker sees no outstanding PKID that needs to be acked, but the client thinks it still needs to ack the duplicate (treated as a new message per spec), resulting in a second ACK being sent for the same PKID. At this point:

1) The broker may choose to end the session with the client for sending an unexpected ACK (in my experience this is most common on QoS2 with unsolicited PUBREC errors)
2) If the broker had reassigned the PKID to a new message (because the PKID was acked means it's available for reassignment), and delivers it to the client before the client sends the second ACK for the duplicate message, the client could incorrectly ack a completely different message.

This is admittedly a fairly unlikely scenario when using QoS1 and automatic acking, but when using QoS2, or when using manual ack (and ESPECIALLY when using QoS2 with manual ack), this scenario actually becomes relatively easy to trigger.

cartertinney:
@swanandx My apologies for the delay in response. Got caught up in some other items.

Even on QoS1 when I invoke `.ack()` while disconnected, I see the `Outgoing(PubAck)` appear on the `EventLoop` after reconnect. Furthermore, the `AckOfAck` I get back after the reconnect resolves to `None` which I believe should indicate successful send of the PubAck for QoS 1.

So, unless I'm missing something, I believe my previous point stands - I'm seeing PubAcks queue up while offline, although the problem is significantly worse on QoS2.

The repro for this is rather simple:

- Create a client that receives messages using manual ack on QoS1.
- After receiving a message, but before invoking `.ack()` drop the network
- After the `EventLoop` has detected the disconnect, allow the invocation of `.ack()` on the message previously received.
- Reconnect the network
- `EventLoop` will reconnect
- You will see the queued `Outgoing(PubAck)` on the `EventLoop` in the logs
- The `Token<AckOfAck>` that was provided by the invocation to `.ack()` will resolve to `None`.

swanandx:
hey @cartertinney , thanks for your patience.

> PubAck for a publish from a previous connection was queued offline and delivered upon reconnect, which is invalid.
>

we do the cleanup when network is dropped ( or there is some other error ).

once that cleanup is done, anything pushed to channel between client and eventloop don't have anything to do with old session.

In case you mentioned above, you are calling `.ack()` _after_ the cleanup is done, so it gets buffered in channel and sent when we are able to. Just like callling `.publish()`, if you call `.publish()` it will put that request in channel and send in on new session. 

I believe this should be users responsibility to call `.ack()` only when it is supposed to be called.

cartertinney:
@swanandx 

I see where you're coming from, and if that's your API philosophy, I do recognize it as consistent with other choices you've made, but I think that in practice expecting the user to behave correctly is a bit impractical given that the connection state information is only known by the event loop polling thread, wheras the part of a hypothetical application that's doing manual acking, or publishing, or anything else is... well, in a different thread. You're basically forcing every application that does manual acking to build constructs to track and share data regarding connection state, which is doable, but seems a bit of an imposition on application design when it would seemingly be easier to implement internally.

Similar things could be said of manual ack ordering though, and again, I do recognize that you making ideologically consistent API design choices, so at least there is consistency. 

Let me try implementing such a scheme to prevent inappropriate calls to `.ack()` and see how much work is required to do it.

swanandx:
> Let me try implementing such a scheme to prevent inappropriate calls to .ack() and see how much work is required to do it.
>
cool, thanks!

> You're basically forcing every application that does manual acking to build constructs to track and share data regarding connection state
>

Only if that is required by application. Generally you won't need to build any construct, it can be achieved by setting the capacity of channel to 0 ( we do it in [uplink](https://github.com/bytebeamio/uplink/blob/0d98ca5c05d343b02f89982fb36d5f5f363c48b1/uplink/src/base/mqtt/mod.rs#L81) as well )

Having it set to 0 ensures that packets don't get accepted when network is down ( because of how poll works ). Then in your application you can opt to _block_ on publish, subscribe, ack etc ( or do it without _blocking_ by try_publish, try_subscribe, try_ack etc. ). 

That is the behavior you are expecting right?  

Thanks for understanding my pov and for the feedback!

izolyomi:
> Hi @izolyomi . We are unfortunately delayed on this. We'll try to focus on these issues in Feb and cut a new release

Does anyone have an updated rough estimated time of arrival for a next release including this change?

KaSroka:
Any updates on this? This is essential to implement durable message queue on top of mqtt.

giridher-art:
hey @KaSroka we are currrently working internal fork of rumqttc which includes all these features and we will open it for testing ASAP.

thehouseisonfire:
> hey @KaSroka we are currrently working internal fork of rumqttc which includes all these features and we will open it for testing ASAP.

Mate you've been saying this for 3 months. Can you just release the fork so we can get these features going (or at least *help you* with the changes)? Else I might just beat you to the punch with my own fork.

izolyomi:
I've seen the recently forked maintained version by @thehouseisonfire . This PR does not seem to be included, am I correct? It's very unfortunate that there's still zero activity around this feature.

With the current implementation publishing messages only enqueue the message in the underlying message channel and it always succeeds as long as it's not full yet, even if the network connection the Mqtt broker is completely failing. Maybe there's a better way of making publishing more reliable by properly detecting network errors but I'm unaware of how it could be done.

thehouseisonfire:
Hi @izolyomi, sorry for the delay.

Working on it. In fact, you may see some traces of it on my latest release (0.28.0) though I don't consider that path up to par yet. It will be my next focus after I cleanup some low hanging fruit I had left to pick.

Hoping to get that done and shipped in 2-3 weeks, maybe earlier, but let's be conservative.

CosminPerRam:
Hey, could this branch be rebased with `main` to be able to use the `use-rustls-no-provider` feature? Thanks.

de-sh:
> Hey, could this branch be rebased with `main` to be able to use the `use-rustls-no-provider` feature? Thanks.

Hope this helps: https://github.com/bytebeamio/rumqtt/commit/6449dffcbe186d56d778950b7ceb50a57109b5c9

PR #929, opened 2024-12-05
Reworked PoC to not return from broker start.

baxterjo:
Instead users must fetch handle before starting broker to ensure it is cancelable.


baxterjo:
@swanandx could you take a look at this?

rumqttd/src/lib.rs
```diff
@@ -23,7 +23,8 @@ pub use link::local;
 pub use link::meters;
 pub use router::{Alert, Forward, IncomingMeter, Meter, Notification, OutgoingMeter, Router};
 use segments::Storage;
-pub use server::{Broker, LinkType, Server};
+pub use server::{Broker, BrokerHandle LinkType, Server};
```
YarekTyshchenko, line 26: typo here, missing comma

rumqttd/src/server/mod.rs
```diff
@@ -4,7 +4,9 @@ mod broker;
 #[cfg(any(feature = "use-rustls", feature = "use-native-tls"))]
 mod tls;
 
-pub use broker::{Broker, LinkType, Server};
+
+pub use broker::{Broker, BrokerHandle LinkType, Server};
```
YarekTyshchenko, line 8: And here too

PR #931, opened 2024-12-29
chore(rumqttd): sync shared dependencies with rumqttc

Sieluna:
## Type of change

## Checklist:


Sieluna:
I think unifying the management of shared dependencies would not only allow developers to consider updates for rumqttd when updating rumqttc but also reduce unnecessary compilation when both are used simultaneously.

Issue #948, opened 2025-02-26
[BUG] Incorrect timeout handling during connection


Ddystopia:
Hello, I am running rumqttd and having connection issues. Any mqtt client, I tried a lot of different ones, is experiencing problems - they keep reconnecting, rumqttd logs that there is timeout.

I looked at the code and that's what I found:

```rust
impl<P: Protocol> Network<P> {
    pub fn new(
        socket: Box<dyn N>,
        max_incoming_size: usize,
        max_connection_buffer_len: usize,
        protocol: P,
    ) -> Network<P> {
        Network {
            socket,
            read: BytesMut::with_capacity(10 * 1024),
            write: BytesMut::with_capacity(10 * 1024),
            max_incoming_size,
            max_connection_buffer_len,
            keepalive: Duration::ZERO,
            protocol,
        }
    }
   // ...
}
```

```rust
async fn remote<P: Protocol>(
    config: Arc<ConnectionSettings>,
    tenant_id: Option<String>,
    router_tx: Sender<(ConnectionId, Event)>,
    stream: Box<dyn N>,
    protocol: P,
    will_handlers: Arc<Mutex<HashMap<String, Sender<AwaitingWill>>>>,
) {
    let mut network = Network::new(
        stream,
        config.max_payload_size,
        config.max_inflight_count,
        protocol,
    );
 
    let dynamic_filters = config.dynamic_filters;

    let connect_packet = match mqtt_connect(config, &mut network).await {
        Ok(p) => p,
        Err(e) => {
            error!(error=?e, "Error while handling MQTT connect packet");
            return;
        }
    };
    // ...
```

And this is `mqtt_connect`:

```rust
pub async fn mqtt_connect<P>(
    config: Arc<ConnectionSettings>,
    network: &mut Network<P>,
) -> Result<Packet, Error>
where
    P: Protocol,
{
    // Wait for MQTT connect packet and error out if it's not received in time to prevent
    // DOS attacks by filling total connections that the server can handle with idle open
    // connections which results in server rejecting new connections
    let connection_timeout_ms = config.connection_timeout_ms.into();
    let packet = time::timeout(Duration::from_millis(connection_timeout_ms), async {
        let packet = network.read().await?;
        Ok::<_, network::Error>(packet)
    })
    .await??;
   // ...
```

And `read`:

```rust
    /// Waits on network for 1 packet
    pub async fn read(&mut self) -> Result<Packet, Error> {
        loop {
[...]

Issue #945, opened 2025-02-19
WebSocket issues when installing from source


alexveecle:
A colleague yesterday had issues with connecting to rumqttd (installed with `cargo install` from source) with <https://rroemhild.github.io/hodd/>. Connections fail, rumqttd logs:

```
  2025-02-19T12:35:46.850237Z  INFO rumqttd::server::broker: accept, name: "ws-1", addr: 127.0.0.1:55772, count: 0, tenant: None

  2025-02-19T12:35:46.852517Z ERROR rumqttd::server::broker: Error while handling MQTT connect packet, error: Network(KeepAlive(Elapsed(())))
    in rumqttd::server::broker::websocket_link

  2025-02-19T12:35:47.159131Z  INFO rumqttd::server::broker: accept, name: "ws-1", addr: 127.0.0.1:55786, count: 1, tenant: None

  2025-02-19T12:35:47.160053Z ERROR rumqttd::server::broker: Error while handling MQTT connect packet, error: Network(Io(Custom { kind: ConnectionAborted, error: "connection closed by peer" }))
    in rumqttd::server::broker::websocket_link
```

`rumqttd-0.19.0-linux-gnu` does not have issues.

Not sure if this is related to recent issues in the tracker.

Ddystopia:
> Thanks for reporting, seems to be this is introduced by PR [#784](https://github.com/bytebeamio/rumqtt/pull/784) .
> 
> as default keepalive is set to `ZERO` so sometimes if the client didn't send full MQTT connect packet, we we end up waiting for network for more data, as timeout is 0 it elapses right away leading to this issue!
> 
> better approach would be to setting it to some sane default value.
> 
> or we can have optional `default_network_read_timeout` like something ( better name plz ) in config as well along with the sane default.
> 
> diff --git a/rumqttd/src/server/broker.rs b/rumqttd/src/server/broker.rs
> index 9886541..f4e9b5f 100644
> --- a/rumqttd/src/server/broker.rs
> +++ b/rumqttd/src/server/broker.rs
> @@ -504,6 +504,8 @@ async fn remote<P: Protocol>(
>          protocol,
>      );
> 
> +    network.set_keepalive(config.default_network_read_timeout); // it can be optional
>      let dynamic_filters = config.dynamic_filters;
> 
>      let connect_packet = match mqtt_connect(config, &mut network).await {
> but say it keeps sending a byte everytime we will end up waiting for that much time which isn't good.
> 
[...]

PR #949, opened 2025-02-26
fix: set default network timeout as `Duration::MAX` instead of zero.

Ddystopia:
## Type of change

## Checklist:

Closes https://github.com/bytebeamio/rumqtt/issues/945.
Closes https://github.com/bytebeamio/rumqtt/issues/948.

### DIFF
--- a/rumqttd/CHANGELOG.md
+++ b/rumqttd/CHANGELOG.md
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - MQTT keep alive interval
+- Change MQTT keep alive interval from `Duration::ZERO` to `Duration::MAX`.
 - record client id for remote link's span
 - session present flag in connack
 - Make write method return the number of bytes written correctly everywhere
--- a/rumqttd/CHANGELOG.md
+++ b/rumqttd/CHANGELOG.md
@@ -25,7 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - MQTT keep alive interval
-- Change MQTT keep alive interval from `Duration::ZERO` to `Duration::MAX`.
+- Change default network keep alive interval from `Duration::ZERO` to `Duration::MAX`.
 - record client id for remote link's span
 - session present flag in connack
 - Make write method return the number of bytes written correctly everywhere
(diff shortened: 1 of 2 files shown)


swanandx (commented):
can we use `set_keepalive` method instead of passing value to `new()`? ( like change suggested in the issue comment )

also, we should add comment on top of it mentioning why we are setting it to connection_timeout_ms.

Ddystopia:
I find it the most logical to pass to the new. Are there some reasons you don't want this?

swanandx:
> I find it the most logical to pass to the new. Are there some reasons you don't want this?

- minimal changes in API
- no overhead of keeping it consistent in multiple places ( like in this PR )
- `set_keepalive` is method intended for this purpose, we are going to overwrite it anyways

Ddystopia:
> > I find it the most logical to pass to the new. Are there some reasons you don't want this?
> 
> * minimal /no changes in API ( non-breaking change )
> * no overhead of keeping it consistent in multiple places ( like in this PR )
> * `set_keepalive` is method intended for this purpose, we are going to overwrite it anyways

As far as I can see, `Network` is not a part of public api. About consistency, are you sure that in other places we *do not* need setting keepalive? And `set_keepalive` accepts whole seconds, so we would need to round up those milliseconds from conf?

Maybe having `Duration::MAX` as a default would suit you more? Anyway there is a timeout in `mqtt_connect`.

swanandx:
> And set_keepalive accepts whole seconds, so we would need to round up those milliseconds from conf?
>

fair point, ideally there should be different config entry but would be too much, and if we want to use connection_timeout_ms we should round it up to bigger value, like `7100` should round to `8`. ( don't know how complicated it will be )


 > Maybe having Duration::MAX as a default would suit you more? Anyway there is a timeout in mqtt_connect.
 >
 reason for using `Duration::ZERO` was so that it would be easier to spot bugs ( like the one we are fixing rn ), it is MAX it will just block indefinitely and would be harder to spot.
 
 
but using `connection_timeout_ms` in bridge doesn't make sense ( as we should either be using keepalive or waiting forever  like before #784 ).

so yeah, let's go with `Duration::MAX` as default and also add comment on top of it mentioning that it would be overwritten by keepalive value in connect packet, otherwise we would wait indefinitely until we get a packet / network error when we call read.

rumqttd/CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - MQTT keep alive interval
+- Change MQTT keep alive interval from `Duration::ZERO` to `Duration::MAX`.
```
swanandx, line 28:
```suggestion
- Change default network keep alive interval from `Duration::ZERO` to `Duration::MAX`.
```

swanandx (approved):
thanks for the PR @Ddystopia ! 🚀

PR #968, opened 2025-05-20
feat(rumqttd): Make Server public

Ichmed:
Publicly export `Server`, `Router` and `LinkType`.

I don't really see a problem with making these types `pub`, the constructor for `Server` is pub already but not accessible from outside the crate afaik.

As-is, calling `Broker::start` and thus spawning a new runtime for each config panics inside already async contexts. With this PR new Servers can just be spawned with `tokio::spawn` or `my_runtime::spawn` instead.

## Type of change

New feature

## Checklist:


tekjar (approved):

tekjar:
Thank you :)

PR #970, opened 2025-05-22
fix: adjust logging

JacksonUtsch:
Change error_span instances to info_span

Don't seem like instances should be error level and they spam the logs.

## Adjust logging

## Checklist:


swanandx:
putting here for discussion ref: #782

Issue #964, opened 2025-05-09
MQTTv5 client: Keep alive required to be >= 5 seconds


Cadynum:
When using v5, the keep-alive is restricted to at least 5 seconds, but I don't find any reference to this restriction in https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901045.

https://github.com/bytebeamio/rumqtt/blob/b1c3a201048f64a13373e2bf9bb28c9cd1aae680/rumqttc/src/v5/mod.rs#L229-L234

A keep alive of 0 is also permitted in v5:

> A Keep Alive value of 0 has the effect of turning off the Keep Alive mechanism. If Keep Alive is 0 the Client is not obliged to send MQTT Control Packets on any particular schedule.

swanandx:
thanks for reporting. looks like this was fixed for v4 client in #655 but not for v5.

also as discussed in mentioned PR, i believe we should change set_keep_alive to take u16 ( which we will treat as seconds ), wdyt?

Cadynum:
I agree, a u16  fits better, as it indicates what kind of resolution is supported.  I'd name it "seconds".

PR #976, opened 2025-07-09
Fix: MQTTv5 keep-alive duration <= 5 sec allowed

Jathn:
This pull request removes the minimum requirement of 5 seconds of keep-alive for MQTTv5. 
Additionally, the case where the keep-alive is 0 is implemented for MQTTv5 as described in the standard. 

The duration is now set with u16 instead of Duration as discussed in #964.
edit: the duration set also changed in optional added commit [691912f](https://github.com/bytebeamio/rumqtt/pull/976/commits/691912f80e53aa9684bf3001daaa88aeb87fbbe5) for v4
It is similar to pull request #655 which changed the minimum requirement and 0-case handling.

Issue: #964 
Attribute: This PR adds support for this in MQTTv5, following the approach used in @danieldougherty 's implementation for MQTTv4 in PR #655.

## Type of change

## Checklist:

### DIFF
--- a/benchmarks/clients/rumqttasync.rs
+++ b/benchmarks/clients/rumqttasync.rs
@@ -18,7 +18,7 @@ async fn main() {
 
 pub async fn start(id: &str, payload_size: usize, count: usize) -> Result<(), Box<dyn Error>> {
     let mut mqttoptions = MqttOptions::new(id, "localhost", 1883);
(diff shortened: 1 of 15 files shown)


Jathn:
@giridher-art I think that could and should be separated to another pull request no?

Issue #1000, opened 2025-10-23
Saturated publish fairness


Indy2222:
Rumqttc disables the default features of the `flume` dependency, which in turn disables its `eventual-fairness` feature flag. Therefore, I believe that when the maximum number of in-flight messages is reached, all publishing tasks start to race, and message ordering can become biased due to hard-to-predict factors (e.g., the OS thread scheduler). I suggest enabling the `eventual-fairness` feature flag to address this.

The dependency:

https://github.com/bytebeamio/rumqtt/blob/c03ba8bbb785dc6cd7809ce14fc2845d14b6bb74/rumqttc/Cargo.toml#L32

Flume feature flags:

https://github.com/zesterer/flume?tab=readme-ov-file#cargo-features

PR #1002, opened 2025-11-05
Make saturated publishes fair

Indy2222:

Fixes https://github.com/bytebeamio/rumqtt/issues/1000

## Checklist:


thehouseisonfire:
Pull should be even better with a version bump to flume 0.12 (should shave off a bit more of the p50 downside)

thehouseisonfire:
Seems like this might not do anything, given flume's `Selector` is not used in the crate.

Issue #1014, opened 2025-12-15
rumqttc says "Timeout while waiting to resolve collision" and can't publish


njaard:
I have a client under load. I establish a connection (AsyncClient) with mostly default options. The client has no subscriptions, it's only used for publishing. I use `set_inflight(1000)` and `set_clean_session(true)`.

Finally, my program publishes many messages over quite a long time with `rumqttc::QoS::AtLeastOnce`.

Upon the MQTT broker restarting (VerneMQ), I get `ConnectionError::MqttState(CollisionTimeout)` from `EventLoop::poll()`. I get it repeatedly every 2 minutes, and my program cannot publish anything anymore indefinitely.

I saw this only once and I'm still trying to find a way to reproduce it.

kiyov09:
I think I found the problem. TL;DR: because you are using `clean_session = true`, the collision will never be resolved.

## Explanation

`ConnectionError::MqttState(CollisionTimeout)` is returned when a collision is detected and stays like that for at least 2 ping request cycles. Each of those cycles takes 60 seconds by default (that explains the 2 minutes interval).

When that error is returned, the event loop is "cleaned," effectively causing a reconnection on the next `poll`. When the connection is established again, if no session is present in the broker (your case because of `clean_session = true`), all pending packets are cleared (starting fresh). The problem is that the collision is still there—nothing clears it. So, the state is "forever" waiting for an ack that will never come.

I think the solution is as easy as clearing the collision from the state along with the pending packets, but I could be missing some context on why it's not being done:

https://github.com/bytebeamio/rumqtt/blob/main/rumqttc/src/eventloop.rs#L161-L163

tekjar:
@giridher-art Can you take a look at this

giridher-art:
@njaard  yeah , not clearing the collision packet, along the pending packets from previous session  will indeed result indefinitely block on publishing the messages when the clean_session = true . the solution specified by @kiyov09 would resolve the issue . and we include it in our next release ASAP.
cc : @tekjar

kiyov09:
I can open a PR now that the bug has been validated, if you haven't done it already. Just let me know.

giridher-art:
@kiyov09 yeah for sure

PR #1019, opened 2026-01-08
fix: clear collision state on reconnection with clean session

kiyov09:
When reconnecting with a clean session (session_present=false), the collision state was not being cleared along with the pending packets. This could cause issues if a collision had occurred in the previous session, as the stale collision state would persist into the new session.

This is a fix for #1014 

## Type of change

## Checklist:


kiyov09:
Hi @giridher-art, pls take a look at my changes and let me know any additional changes I need to make. Thnx!

giridher-art:
hey @kiyov09 , we are currently working on a new architectural design for rumqttc . it contains all these changes and we may release it ASAP.

PR #1021, opened 2026-01-12
fix:(rumqttc): Remove tls default provider from dependencies

tlevora:
Updating 'rustls-webpki' to 0.103.0 removes default ring tls-provider from 'rustls-webpki' dependency and allows to specifically choose between 'ring' and 'aws-lc-rs'. Consider this commit fixing usage of 'use-rustls-no-provider' feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2230,7 +2230,7 @@ dependencies = [
  "pretty_env_logger",
  "rustls-native-certs",
  "rustls-pemfile",
- "rustls-webpki 0.102.8",
+ "rustls-webpki 0.103.8",
  "serde",
  "thiserror 2.0.17",
  "tokio",
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -11,7 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 ### Deprecated
 ### Removed
-### Fixed 
+### Fixed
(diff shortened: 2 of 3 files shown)


giridher-art:
@tlevora this looks good , but does it break any clients that depend on use-rustls feature ? and i try to review it ASAP

tlevora:
> @tlevora this looks good , but does it break any clients that depend on use-rustls feature ? and i try to review it ASAP

@giridher-art, you are right this is a breaking change. It basically removes all that `ring` verification algorithms from `ALL_VERIFICATION_ALGS` in `rustls-webpki` if they are not added with an appropriate feature flag in any other dependency.

According to this issue what is the proper solution? Shall I just modify `CHANGELOG.md` record and a commit message?

---

Then there is a question whether this is a good approach or not. If this commit will be applied as it is and a user of `rumqttc` will want to use either `ring` or `aws-lc`, he/she will need to add the `rustls-webpki` to his/her `Cargo.toml` as a dependency with the proper feature flag (either `ring` or `aws_lc_rs`).

Another solution could be adding another two feature flags to `rumqttc`:
```toml
[features]
# this is a current solution
use-rustls-no-provider = ["dep:tokio-rustls", "dep:rustls-webpki", "dep:rustls-pemfile", "dep:rustls-native-certs"]
# proposed features
use-rustls-ring = ["rustls-webpki?/ring"]
use-rustls-aws-lc = ["rustls-webpki?/aws_lc_rs"]
```

Then a user of `rumqttc` will be able to choose between `ring` and `aws-lc` in `rustls-webpki` dependency directly.

PR #1022, opened 2026-01-13
feat: add backoffs when reconnecting

ryan-resideo:
hiya, thanks for making a neat library!

we ran into an issue where on a connection failure the library loops attempting re-connection, which (depending on your timeouts) can occur at a pace that is hazardous to other things sharing the executor.

while it's pretty easy to wrap the connection failed path with a backoff, we thought it might be nice to solve this at the library level? so this PR introduces an (optional) exponential backoff to re-connection attempts, with a flag to enable/disable this on the `NetworkOptions` object.

at the moment this just uses the defaults from [backon](https://docs.rs/backon/latest/backon/struct.ExponentialBuilder.html#default) with `max_times` unset, but more configuration could be exposed if it was useful.

## Type of change

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -321,6 +321,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "backon"
+version = "1.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cffb0e931875b666fc4fcb20fee52e9bbd1ef836fd9e9e04ec21555f9f85f7ef"
+dependencies = [
+ "fastrand",
+]
+
+[[package]]
 name = "backtrace"
 version = "0.3.76"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


giridher-art:
@ryan-resideo one solution would be take a reconnection delay from the user to wait for the timeout before the trying to reconnect. instead of trying to add seperate crate to this , this would be one of the solution.

ryan-resideo:
yep. the advantage of a backoff is that you get quick reconnects when it's possible to reconnect _and_ a useful delay in more serious outages, where with a timeout you have to trade off between these.

giridher-art:
hey @ryan-resideo , instead of applying backoff directly in the lib , you can implement your own in eventloop right .
```
 match eventloop.poll().await {
            Ok(event) => {
                println!("Connected event: {:?}", event);
            }
            Err(e) => {
            // you can add your backoff or const timeout before you poll again                 );
            }
        }
```

ryan-resideo:
>  instead of applying backoff directly in the lib , you can implement your own in eventloop right

yep, and we have! it's just that, every client should really use a backoff so it seemed worth a PR. can close if y'all aren't interested.

PR #1033, opened 2026-02-19
feat(rumqttc): update async-tungstenite to 0.32, remove ws_stream_tungstenite

michelhe:
async-tungstenite 0.32 includes a built-in bytes module with ByteReader and ByteWriter types that provide tokio::io::AsyncRead/AsyncWrite over WebSocket streams. This makes the ws_stream_tungstenite crate redundant, and that crate pins async-tungstenite ^0.29, blocking the upgrade.

Changes:
- Update async-tungstenite from 0.29 to 0.32
- Remove ws_stream_tungstenite dependency
- Add a custom WsStream adapter in websockets.rs that wraps async-tungstenite's ByteReader/ByteWriter to provide tokio IO traits
- Add futures-io and pin-project-lite as optional websocket dependencies
- Update v4 and v5 eventloop imports accordingly

## Type of change

## Checklist:

- [x] Make an entry to `CHANGELOG.md` if it's relevant to the users of the library. If it's not relevant mention why. - Not relevant, this is an internal implementation detail.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -200,9 +200,9 @@ dependencies = [
 
 [[package]]
 name = "async-tungstenite"
-version = "0.29.1"
+version = "0.32.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef0f7efedeac57d9b26170f72965ecfd31473ca52ca7a64e925b0b6f5f079886"
+checksum = "8acc405d38be14342132609f06f02acaf825ddccfe76c4824a69281e0458ebd4"
 dependencies = [
  "atomic-waker",
  "futures-core",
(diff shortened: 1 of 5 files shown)


thehouseisonfire:
`pin_project!` is actually better than my current approach if the downstream transport is `!Unpin`, though from what I can tell most today are Unpin (like `tokio::net::TcpStream` and `quinn`'s `RecvStream/SendStream`). 

Would you (or anyone reading) happen to have a reference or use case for supporting `!Unpin`?

thehouseisonfire:
Also noting that the dependency tree does include `proc-macro2` somewhat extensively, so [`pin-project`](https://github.com/taiki-e/pin-project) could also be used [per `pin-project-lite`'s maintainer.](https://docs.rs/pin-project-lite/latest/pin_project_lite/#different-no-proc-macro-related-dependencies)

PR #1038, opened 2026-04-01
fix(v5): Protocol Error on invalid property values

de-sh:
Fixes three cases where MQTT 5.0 property values of `0` were silently accepted instead of rejected with a Protocol Error.

## Type of change

  | Packet | Property | Spec | Forbidden value |
  |---|---|---|---|
  | `SUBSCRIBE` | Subscription Identifier | [§3.8.2.1.2](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901166) | `0` — valid range is 1–268,435,455 |
  | `PUBLISH` | Topic Alias | [§3.3.2.3.4](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901113) | `0` — aliases should not have value 0 |
  | `CONNECT` / `CONNACK` | Receive Maximum | [§3.1.2.11.3](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901049), [§3.2.2.3.3](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901083) | `0` — would mean no in-flight messages permitted |

  Each case now returns `Error::MalformedPacket` on parse and is covered by a test that constructs the raw bytes directly.

 
>  **Note:** the spec requires sending a `DISCONNECT` with a specific reason code for each violation (respectively `0x82`, `0x94`, `0x82`). This is not yet done — the error surfaces as a connection drop rather than a clean disconnect. See the `TODO` code comments.

## Checklist:

### DIFF
--- a/rumqttc/src/v5/mqttbytes/v5/connack.rs
+++ b/rumqttc/src/v5/mqttbytes/v5/connack.rs
@@ -233,7 +233,12 @@ impl ConnAckProperties {
                     cursor += 4;
                 }
                 PropertyType::ReceiveMaximum => {
-                    receive_max = Some(read_u16(bytes)?);
+                    let val = read_u16(bytes)?;
+                    // MQTT 5.0 §3.2.2.3.3: Receive Maximum value of 0 is not permitted
+                    if val == 0 {
+                        return Err(Error::MalformedPacket);
+                    }
+                    receive_max = Some(val);
                     cursor += 2;
                 }
                 PropertyType::MaximumQos => {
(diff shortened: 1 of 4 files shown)


de-sh:
I have observed many more issues in the protocol layer, do you want me to raise inidividual PRs to fix them one by one or would you rather have them as commits in a single PR?

thehouseisonfire:
Hi @de-sh, how have you been doing?

I definitely prefer, if that would be possible, if you opened individual PRs for each issue you wish fixed. Mapping out compliance with each spec requirements is the next step I have on (after #916), but I will definitely prioritize ones people need first.

de-sh:
> Hi @de-sh, how have you been doing?
> 
> I definitely prefer, if that would be possible, if you opened individual PRs for each issue you wish fixed. Mapping out compliance with each spec requirements is the next step I have on (after #916), but I will definitely prioritize ones people need first.

Hey @thehouseisonfire, I have been figuring out how to surf the market turmoils 😅. Have just started a career break, decided to give back to the community since I have been pretty occupied of late, trying(a bit too hard) to keep my career going, a bit burnt by the experience, good personality development though 😆 

Ok, will open PRs against your fork as well, keeping this in mind.

thehouseisonfire:
@de-sh I feel you man, hope you can enjoy your break and land a great one once you get back on the race.

> Ok, will open PRs against your fork as well, keeping this in mind.

I actually had some of those rejections implemented already, though I wasn't returning `Error::MalformedPacket` as I should have been. If you can, I would definitely appreciate if you checked if everything is right as-is on my end, even though it's quite a bit different from upstream by now mostly due to the split crates. No pressure though, I can figure it out from your PRs here too.

PR #1041, opened 2026-04-05
perf: replace flume with `tokio::sync::mpsc`

de-sh:
Replace flume channel with `tokio::sync::mpsc` in the eventloop and client. The sync Client gains a runtime Handle for block_on sends.

Benchmark Results: tokio-sync vs main

refer: [bench/throughput](https://github.com/de-sh/rumqtt/tree/bench/throughput)

  | Payload | main (median) | tokio-sync (median) | Δ time | Δ throughput |
  |---------|---------------|---------------------|--------|--------------|
  | 32 B    | 1.402 ms      | 1.316 ms            | −5.4%  | +5.8%        |
  | 100 B   | 1.397 ms      | 1.383 ms            | −3.5%  | +3.6%        |
  | 1024 B  | 1.664 ms      | 1.716 ms            | +3.3%  | −3.2%        |

  All results are statistically significant (p < 0.05).

  ---
  Interpretation

  Small messages (32–100 B): tokio-sync wins by 3.5–5.4%. These are channel-bound — the message is tiny relative to the per-message bookkeeping cost, so the reduction in lock contention from dropping flume's mutex pays off directly. The channel send/receive path sees fewer cache line bounces.

  Large messages (1024 B): tokio-sync is 3.3% slower. At 1 KB, the payload dominates. The likely cause is that tokio's bounded mpsc uses an intrusive linked-list of Arc-allocated nodes rather than flume's pre-allocated ring buffer, so 1 KB payloads incur an extra allocation + Arc refcount per message where flume's ring buffer would reuse a slot. This is a known design tradeoff in tokio mpsc.

  The crossover point is somewhere between 100 B and 1024 B. For the typical MQTT use case (small control messages, moderate publish payloads), the switch is neutral-to-positive overall.

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2220,7 +2220,6 @@ dependencies = [
  "bytes",
  "color-backtrace",
  "fixedbitset 0.5.7",
- "flume",
  "futures-util",
  "http 1.3.1",
  "log",
--- a/rumqttc/Cargo.toml
+++ b/rumqttc/Cargo.toml
@@ -29,7 +29,6 @@ tokio = { version = "1.36", features = ["rt", "macros", "io-util", "net", "time"
 tokio-util = { version = "0.7", features = ["codec"] }
 bytes = "1.5"
 log = "0.4"
-flume = { version = "0.11", default-features = false, features = ["async"] }
 thiserror = "2.0.8"
 
 # Optional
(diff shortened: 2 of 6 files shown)


de-sh:
This PR was for documentation purpose. A note of why both `tokio::sync::mpsc` and flume are good enough.

Issue #1029, opened 2026-02-17
I Will Be Maintaining a Fork For Rumqttc For Now


thehouseisonfire:
Hello, and thank you for maintaining Rumqttc.

Some feature PRs and fixes that are relevant to my upcoming research project have been open for a while. To make my work easier in the meantime, I created a fork merging them. Given that:

- I will be dogfooding these changes in my own project.
- Most of their scopes are small.
- They are based on open PRs I have reviewed and adjusted.

I am confident that advertising them here can help someone until the architectural upgrade you guys seem to have planned can be merged.

My goal here is not to fragment the ecosystem or compete with the project, just provide a drop-in option for others users who need these features today since I'll be doing this work anyways.

Fork: https://crates.io/crates/rumqttc-next

Repo: https://github.com/thehouseisonfire/rumqtt

### PRs merged (as of 0.26.1):

  - #852 - This is the biggest change merged so far.
  - #901 - Websocket still not supported.
  - #909
  - #915 - reworked to preserve them end-to-end.
  - #976 
  - #977 - reworked to not add more complexity to the EventLoop API.
  - #980 - small refactor to avoid an allocation.
  - #1011
  - #1017 - significantly reworked, added an enum for correctness.
  - #1019
  - #1021 - simpler fix route

  Other work:

  - Bump dependencies:
      - `tokio` 1.36 -> 1.40
      - `flume` 0.11 -> 0.12
      - `rustls-webpki` 0.102.8 -> 0.103.9 (related to #1021)
      - `async-tungstenite` 0.29.0 -> 0.29.1
      - `color-backtrace` 0.6.1 -> 0.7.2
      - Benchmarks/support crates: `pprof` 0.14 -> 0.15, `thiserror` 1 -> 2
      - Added optional scram dependency and auth-scram feature (related to #852).
  - Change deprecated bincode dependency to wincode (as recommended by bincode).
  - Fix or ignore clippy lints.
  - Fix reliability test keep-alive usage and that old docs code-fence issue.

### PRs and changes merged (as of 0.27.0):

- #932 – Defaults to `use-rustls`.
  - Compatibility note: `async-tungstenite` defaults to `native-tls` for WSS when both TLS features are enabled. Attempting to use a rustls `TlsConfiguration` variant for WSS when both are enabled will return a runtime `tls::Error::TlsBackendConflict`.
[...]

PR #1045, opened 2026-04-09
fix(rumqttc): support binary passwords in Login

### LINKED ISSUE #1044: (rumqttc) v5: password field in Login struct should support binary data
The `Login` struct in `v5` currently stores the password as a `String`, which prevents using binary passwords (e.g. tokens, hashed credentials, or raw bytes).

The MQTT 5.0 specification ([section 3.1.3.5](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901070)) defines the Password field as [Binary Data](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901012), not as UTF-8 encoded text.

f-did:
Closes: https://github.com/bytebeamio/rumqtt/issues/1044

## Type of change

Change the password field in the Login struct (v5) from String to Vec<u8>, aligning with the MQTT 5.0 spec, which defines the Password as Binary Data.

## Checklist:

- [ x] Formatted with `cargo fmt`
- [ x] Make an entry to `CHANGELOG.md` if it's relevant to the users of the library. If it's not relevant mention why.

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 ### Changed
+- Use `Vec<u8>` for password in `Login` to comply with MQTT 5.0 spec
+
 ### Deprecated
 ### Removed
 ### Fixed 
--- a/rumqttc/src/v5/mod.rs
+++ b/rumqttc/src/v5/mod.rs
@@ -267,7 +267,7 @@ impl MqttOptions {
     }
 
     /// Username and password
-    pub fn set_credentials<U: Into<String>, P: Into<String>>(
+    pub fn set_credentials<U: Into<String>, P: Into<Vec<u8>>>(
         &mut self,
         username: U,
         password: P,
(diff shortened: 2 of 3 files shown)


thehouseisonfire:
The spec requirement is correct, though v4 also expects the same, not only v5. This is already fixed in the newest release of my fork.

For context, this is particular useful for passing binary tokens during CONNECT, and Mosquitto has recently added this spec requirement for their Auth Plugin, so it should be supported in their next release as well (or today if you compile it).

f-did:
Hi @thehouseisonfire, thank you for your reply.

> This is already fixed in the newest release of my fork.

Nice. What's the plan here? Is it planned to merge your fork into the upstream? Or will this remain as a separate fork going forward?

thehouseisonfire:
Hi @f-did,

Currently, I'm maintaining a fork to address some standing feature requests and keep things updated. I plan to keep doing this until @bytebeamio has the bandwidth to pick the project back up, at which point I hope to help move things forward upstream.

If any of the open PRs interest you, there's a good chance my fork already supports them. Otherwise, this upstream version is still completely functional.

You can find more information on the issue I opened to announce it, at #1029.

PR #670, opened 2023-08-03
rumqttc | Unbounded client support

carlocorradini:
## Type of change

Fix #658 

- Breaking change (fix or feature that would cause existing functionality to not work as expected)
   Add support for unbounded client.
   Changed `cap` parameter's type from `usize` to `Option<usize>` in `Client::new` and `AsyncClient::new` to decide whether the channel is bounded or unbounded.
   _Before:_
   ```rs
   Client::new(mqttoptions, 10);   // Bounded only
   ```
   _Now:_
   ```rs
   Client::new(mqttoptions, Some(10));   // Bounded
   Client::new(mqttoptions, None);         // Unbounded
   ```

## Checklist:

- [X] Formatted with `cargo fmt`
- [X] Make an entry to `CHANGELOG.md` if its relevant of user of the library. If its not relevant mention why.

## 🆘 Help 🆘

I need help with *tests*, thanks!

Issue #263, opened 2021-03-15
Subscribe potentially locks event loop


rcastill:
Maybe this is an user problem more than it is a library design issue, but I'll leave the following use case here in the event that it wasn't discussed yet.

If an user decides to subcribe to topics upon confirmed connection, this could potentially lock the event loop if the requests async channel is full. Reproducible example:

```rust
use std::time::Duration;

use rumqttc::{AsyncClient, ConnectReturnCode, Incoming, MqttOptions, QoS};

async fn publisher(mqcli: AsyncClient) {
    loop {
        println!("Publishing payload");
        if let Err(e) = mqcli
            .publish("some/random/topic", QoS::ExactlyOnce, false, "payload")
            .await
        {
            println!("Could not publish message: {}", e);
        }
        tokio::time::sleep(Duration::from_secs(1)).await;
    }
}

#[tokio::main]
async fn main() {
    let mut opts = MqttOptions::new("SomeRandomId", "localhost", 1883);
    let max_packet_size = 1024 * 1024;
    opts.set_max_packet_size(max_packet_size, max_packet_size)
        .set_keep_alive(15)
        .set_inflight(10)
        .set_request_channel_capacity(3)
        .set_clean_session(true);
    let (mqcli, mut mqev) = AsyncClient::new(opts, 10);
    tokio::spawn(publisher(mqcli.clone()));
    loop {
        match mqev.poll().await {
            Ok(rumqttc::Event::Incoming(Incoming::ConnAck(ack))) => {
                if ack.code == ConnectReturnCode::Success {
                    println!("Successfully connected to mqtt broker");
                    if let Err(_) = mqcli.subscribe("some/other/topic", QoS::ExactlyOnce).await {
                        println!("Could not subscribe");
                    }
                } else {
                    println!("Could not connect to mqtt broker: {:?}", ack.code);
                    tokio::time::sleep(Duration::from_secs(5)).await;
                }
            }
            // On disconnection, wait before reconnection
            Ok(rumqttc::Event::Incoming(Incoming::Disconnect)) => {
                println!("Disconnected from mqtt broker");
                tokio::time::sleep(Duration::from_secs(5)).await;
            }
            Ok(_) => {}
            // On error, wait before reconnection
            Err(e) => {
                println!("Error while polling mqtt eventloop: {}", e);
                tokio::time::sleep(Duration::from_secs(5)).await;
            }
        }
    }
}
```

[...]

PR #1047, opened 2026-04-24
feat(rumqttc): add support for unbounded client

imjyotiraditya:
## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Description

Add `ClientBuilder` and `AsyncClientBuilder` for both v4 and v5 clients.
The channel is bounded by default using the capacity set in `MqttOptions`,
with opt-in unbounded channels via `.unbounded()`.

### Before

```rust
let (client, connection) = Client::new(mqttoptions, 10);
let (client, eventloop) = AsyncClient::new(mqttoptions, 10);
```

### After

```rust
// bounded with explicit capacity
let (client, connection) = Client::builder(mqttoptions).capacity(10).build();
let (client, eventloop) = AsyncClient::builder(mqttoptions).capacity(10).build_async();

// bounded with capacity from MqttOptions (new default)
let (client, connection) = Client::builder(mqttoptions).build();
let (client, eventloop) = AsyncClient::builder(mqttoptions).build_async();

// unbounded (opt-in, removes backpressure)
let (client, connection) = Client::builder(mqttoptions).unbounded().build();
let (client, eventloop) = AsyncClient::builder(mqttoptions).unbounded().build_async();
```

The existing `Client::new` and `AsyncClient::new` are deprecated but not removed,
so existing code continues to work without changes.

## Checklist

- [x] Added entry to `CHANGELOG.md`

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -8,8 +8,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `ClientBuilder` and `AsyncClientBuilder` for both v4 and v5 clients,
+  enabling bounded channels by default using the capacity set in `MqttOptions`,
+  with opt-in unbounded channels via `.unbounded()`.
+
 ### Changed
 ### Deprecated
+- `Client::new(options, cap)` and `AsyncClient::new(options, cap)` -
+  use `Client::builder(options).build()` or
+  `Client::builder(options).capacity(cap).build()` instead.
+
 ### Removed
 ### Fixed 
 ### Security
--- a/rumqttc/examples/unbounded.rs
+++ b/rumqttc/examples/unbounded.rs
@@ -1,0 +1,25 @@
+use rumqttc::{Client, MqttOptions, QoS};
+use std::thread;
+use std::time::Duration;
+
+fn main() {
+    let mqtt_opts = MqttOptions::new("test-1", "localhost", 1883);
+
+    // unbounded channel, no backpressure
+    let (mut client, mut connection) = Client::builder(mqtt_opts).unbounded().build();
+
+    client.subscribe("hello/rumqtt", QoS::AtMostOnce).unwrap();
+
+    thread::spawn(move || {
+        for i in 0..10 {
+            client
+                .publish("hello/rumqtt", QoS::AtLeastOnce, false, vec![1; i])
+                .unwrap();
+            thread::sleep(Duration::from_millis(100));
+        }
+    });
+
+    for notification in connection.iter() {
+        println!("Notification = {:?}", notification);
+    }
+}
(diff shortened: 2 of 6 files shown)


thehouseisonfire:
Hi @imjyotiraditya, and thanks for the PR.

I do think the builder pattern might be better long term, though with just two options for now it's not that beneficial.

As for the unbounded clients, any reason that setting a really high bound wouldn't cut it?

imjyotiraditya:
Hi @thehouseisonfire, thanks for the review!

On the builder pattern - agreed it's lightweight with just two options today, but it's
the exact API shape the maintainers requested when closing #670 ("builder pattern where
unbounded is default and users can opt into .capacity()"). This PR is a direct
implementation of that request.

On high bound vs unbounded - a few references worth noting:

1. eventloop.rs itself already recommends unbounded in a doc comment on the clean()
   method: "For this reason we recommend setting AsyncClient's channel capacity to 0"
   (capacity 0 in flume = unbounded). So the library's own source code agrees.

2. The official FAQ has a dedicated section "Program is stuck forever on publish /
   subscribe" documenting the bounded channel deadlock - it's an acknowledged footgun,
   not an edge case.

3. Issue #263 traces this back to 2021 - calling subscribe on ConnAck deadlocks if the
   channel is full because the event loop can't drain the channel while blocked waiting
   to send to it. A high bound reduces the window but doesn't eliminate this.

4. We hit this exact bug in production: on network reconnect we flush offline-stored
   messages in the ConnAck callback. With a bounded channel the flush fills the channel
   immediately and all subsequent publishes silently fail - the service appears connected
   but nothing goes through. Unbounded fixed it completely.

Happy to simplify the API further if you have a preference on shape!

imjyotiraditya:
Hi @thehouseisonfire, just checking in - happy to address any further feedback or simplify the API if needed!

thehouseisonfire:
Hey @imjyotiraditya,

Sorry for the delay — I was working on a larger feature in my fork at the same time.

I ended up making unbounded clients opt-in with an `.unbounded()` call rather than the new default, since bounded clients are historically more common and likely still the more commonly used behavior (of course, I may be wrong about that).

That said, I did like the API shape from your PR. It has now been merged and is working in my fork of this crate, so feel free to take a look there and see whether it fits yours need.

Longer term, I do think your API design is better than what the upstream crate currently exposes.

imjyotiraditya:
Hey @thehouseisonfire,

Sorry for the late update, I thought I had already replied!

Thanks for the feedback and for sharing your fork, that was really helpful. I've updated the PR to make unbounded opt-in via `.unbounded()`, matching the approach you described. Referencing your changes made it straightforward to align, and everything is working as expected on my end.

Do you think we can get this merged?

thehouseisonfire:
Hi again @imjyotiraditya,

Although I quite liked the idea of these changes, so much so I adopted them (in fact, I added the builder pattern to others parts of the code too, it fit quite well) I am not currently a maintainer of this upstream repo. Any merge decision is rightfully on @tekjar hands and other folks working at @bytebeamio.

PR #1052, opened 2026-05-25
rumqttc: emit Event::Reconnect on successful re-CONNACK (closes #250)

### LINKED ISSUE #250: Automatic reconnect and subscribed topics?
This might be duplicate to #211 

`rumqttc` seems to recover/reconnect properly after connection error, but after the reconnect, I'm not seeing any data coming in from subscribed topics:
```
... data coming in from topics...
Error = MqttState(Io(Custom { kind: ConnectionAborted, error: "connection closed by peer" }))
Incoming = ConnAck(ConnAck { session_present: false, code: Success })
... no more incoming data...
```

This is basically a modified async example. To trigger the issue, just jank the network cable or disable wireless a bit (in case broker is on another machine)...
```
#[tokio::main]
async fn main() {

    let mut mqttoptions = MqttOptions::new("test-subscribe", "mqtt-server", 1883);
    mqttoptions.set_keep_alive(5);

    // Subscribe to everything...
    client.subscribe("#", QoS::AtMostOnce).await.unwrap();

    loop {
        match eventloop.poll().await {
            Ok(Event::Incoming(Packet::Publish(p))) => {
                    println!("Incoming = {:?}, {:?}", p.topic, p.payload);
            },
            Ok(Event::Incoming(Packet::PingResp)) |
            Ok(Event::Outgoing(Outgoing::PingReq)) => {},
            Ok(Event::Incoming(i)) => {
                println!("Incoming = {:?}", i);
            },
            Ok(Event::Outgoing(o)) => {
                println!("Outgoing = {:?}", o);
            },
            Err(e) => {
                println!("Error = {:?}", e);
                // XXX: Here I have to re-subscribe to the topics :(
                client.subscribe("#", QoS::AtMostOnce).await.unwrap();
            }
        }
    }
}
```

pbujanda1978:
## Problem

After a transport drop and successful reconnect, the broker has a fresh session. With `clean_session = true` (or session expiry), every subscription the client had is gone. rumqttc itself doesn't remember them either, so the application is silent on the new session until it re-subscribes — but it has no signal that a reconnect happened.

This is [#250](https://github.com/bytebeamio/rumqtt/issues/250): a long-standing rumqttc footgun where a transient network blip leaves a process running but unable to receive any publishes, with no error in the event loop (the connection is "fine" again, just empty of subscriptions).

## Fix

Track `has_connected_before` in the `EventLoop`. After the second (and later) successful CONNACK, push a synthetic `Event::Reconnect` into the state events queue so the caller's `poll()` loop sees it.

The initial CONNACK is unchanged (Reconnect is **not** emitted on first connect, only subsequent ones), so existing callers that subscribe once on startup keep working — they just have to add the `Reconnect` arm to keep working across reconnects.

## Caller pattern

```rust
match eventloop.poll().await {
    Ok(Event::Reconnect) => {
        for (topic, qos) in &my_subs {
            client.subscribe(topic, *qos).await?;
        }
    }
    Ok(Event::Incoming(packet)) => { /* ... */ }
    Ok(Event::Outgoing(_)) => {}
    Err(e) => { /* ... */ }
}
```

## Validation

Tested against multiple brokers (rumqttd, rmqtt, mosquitto): a broker restart that previously left the client silently dead now fires `Event::Reconnect` within seconds, the caller re-subscribes, and message flow resumes without process restart.

## Notes

- Scope: only v5 here — the same fix applies to v4; happy to add it in this PR or a follow-up, whichever you prefer.
- No public API breaks (new enum variant only); downstream pattern matches that don't include `Reconnect` will get a warning rather than an error if they have a wildcard arm.

Closes #250

### DIFF
--- a/rumqttc/src/v5/eventloop.rs
+++ b/rumqttc/src/v5/eventloop.rs
@@ -82,6 +82,10 @@ pub struct EventLoop {
     network: Option<Network>,
     /// Keep alive time
     keepalive_timeout: Option<Pin<Box<Sleep>>>,
+    /// True once the first connect succeeded. Used to distinguish the
+    /// initial CONNACK from a reconnect CONNACK so we can emit
+    /// `Event::Reconnect` only on the latter.
+    has_connected_before: bool,
 }
 
 /// Events which can be yielded by the event loop
@@ -89,6 +93,14 @@ pub struct EventLoop {
 pub enum Event {
     Incoming(Incoming),
     Outgoing(Outgoing),
+    /// Synthetic event emitted by the event loop right after a
+    /// successful reconnect (NOT after the first connect). The caller
+    /// should re-subscribe to all its topics here: rumqttc does not
+    /// remember subscriptions across reconnects, and with
+    /// `clean_session = true` the broker drops them too, so on the new
+    /// session the client is silent until it re-subscribes.
+    /// See https://github.com/bytebeamio/rumqtt/issues/250
+    Reconnect,
 }
 
 impl EventLoop {
@@ -110,6 +122,7 @@ impl EventLoop {
             pending,
             network: None,
             keepalive_timeout: None,
+            has_connected_before: false,
         }
     }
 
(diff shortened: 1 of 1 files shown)


thehouseisonfire:
I’ve added an example showing how to achieve similar behavior in my fork, although I chose not to introduce an additional `Event` enum variant.

PR #1053, opened 2026-06-04
fix(rumqttc): respond with PUBCOMP instead of disconnecting on unsoli…

shucanwu:
…cited PUBREL

When a broker retransmits PUBREL because PUBCOMP was delayed in transit (e.g. under broker congestion), rumqttc currently returns StateError::Unsolicited(pkid), which triggers TCP disconnect/reconnect.

This creates an infinite loop under short message_retry_interval:
1. Broker sends PUBREL, starts retry timer (e.g. 1s)
2. Client sends PUBCOMP, clears QoS 2 state
3. PUBCOMP is delayed > retry interval due to congestion
4. Broker retransmits PUBREL
5. Client has no state for this pkid → Unsolicited error → disconnect
6. Reconnect with session_present=true, broker replays all pending PUBRELs
7. Back to step 4 → infinite reconnect storm

Fix for v5 (src/v5/state.rs):
Per MQTT v5 spec §3.6.2, on receiving PUBREL with unknown Packet Identifier, respond with PUBCOMP and continue the session.

Fix for v3.1.1 (src/state.rs):
The spec is silent here, but sending PUBCOMP and staying connected is strictly safer than disconnecting, since the broker is completing a transaction the client already acknowledged.

Adds regression test unsolicited_pubrel_should_send_pubcomp_without_disconnecting to both v3.1.1 and v5 state modules.

## Checklist:

### DIFF
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -11,7 +11,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 ### Deprecated
 ### Removed
-### Fixed 
+### Fixed
+* Respond with PUBCOMP instead of disconnecting on unsolicited PUBREL (#1053).
+  When a broker retransmits PUBREL (e.g. PUBCOMP was delayed beyond
+  `message_retry_interval`), rumqttc previously returned
+  `StateError::Unsolicited`, causing a disconnect/reconnect loop. Now it
+  sends PUBCOMP and continues the session, matching MQTT v5 spec §3.6.2.
+  The same conservative fix is applied to the v3.1.1 state machine.
 ### Security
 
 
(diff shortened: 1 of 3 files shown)


thehouseisonfire:
Hi @shucanwu, thank you for the PR.

The spec does seem to support this behavior, although returning reason code `0x92` (`PacketIdentifierNotFound`) would be more correct than sending a plain `PubCompReason::Success`.

I’ve implemented something similar in my fork, in case anyone else needs this behavior.

shucanwu:
> Hi @shucanwu, thank you for the PR.
> 
> The spec does seem to support this behavior, although returning reason code `0x92` (`PacketIdentifierNotFound`) would be more correct than sending a plain `PubCompReason::Success`.
> 
> I’ve implemented something similar in my fork, in case anyone else needs this behavior.

Hi @thehouseisonfire , thanks for the suggestion!

I've updated the v5 path to respond with PubCompReason::PacketIdentifierNotFound (0x92) as required by spec §3.6.2, and added an assertion in the test to verify the reason code.

PR #1059, opened 2026-07-13
feat(rumqttc): add pluggable Connector for custom transports

mlowicki-atai:
Introduce a `Connector` trait that supplies the base byte stream to the broker, replacing the default `tokio::net` TCP socket. This lets the client run over arbitrary transports — most usefully a simulated network such as [turmoil](https://github.com/tokio-rs/turmoil) for deterministic testing.

The connector is installed on `NetworkOptions` (the seam shared by the v4 and v5 event loops). When set, it fully owns socket creation and the `proxy` setting is bypassed. The configured `NetworkOptions` are handed to `Connector::connect`, so an implementation can honor the low-level TCP settings (nodelay, send/recv buffer sizes, `bind_addr`, `bind_device`) where they apply to its transport; getters for these were added so they are readable by connector implementations. TLS/WebSocket transports still layer on top of the stream the connector returns.

The API is always available (no cargo feature); `async-trait` is a regular dependency.

Includes a turmoil-based integration test running the client and a minimal in-simulation broker over the connector.

## Type of change

## Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1439,6 +1439,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "libm"
+version = "0.2.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
+
+[[package]]
 name = "linux-raw-sys"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1670,6 +1676,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "071dfc062690e90b734c0b2273ce72ad0ffa95f0c74596bc250dcfd960262841"
 dependencies = [
  "autocfg",
+ "libm",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


thehouseisonfire:
Hi @mlowicki-atai, and thank you for the issue.

From my understanding, my fork already supports using `turmoil` (and other custom transports):

```rs
use rumqttc::MqttOptions;
use turmoil::net::TcpStream;

let mut options = MqttOptions::new(client_id, ("broker", 1883));
options.set_socket_connector(|addr, _| async move { TcpStream::connect(addr).await });
```

Two advances of this approach is that we don't need to pull in `async-trait`, and it doesn't need to bypass the proxy.

I wish I had known about `turmoil` before, it would have been useful for the project that got me using `rumqtt`. I have also added some explicit tests for it on my end, so it should be explicitly supported from now on.

mlowicki-atai:
> I wish I had known about `turmoil` before, it would have been useful for the project that got me using `rumqtt`. I have also added some explicit tests for it on my end, so it should be explicitly supported from now on.

Thanks for the reply. Why your changes are only inside the fork and not pushed upstream? Wondering if I should continue  using https://github.com/bytebeamio/rumqtt or maybe your fork or some other might be a better choice.

thehouseisonfire:
Hi @mlowicki-atai,

As of today, I do not have stewardship of this upstream repo, given it belongs to @bytebeamio. I happened to need some of the pending features from their pull requests, however, and they were not in a position to continue development for the time being, so I took the opportunity to pick up development myself.

As for whether you should continue using upstream or switch over, that largely depends on your requirements. Upstream currently has some active advisories from outdated dependencies, though those are not particularly difficult to address if you decide to look into them. Outside of that, it is the more stable and battle-tested version. If you happen to run into issues, whether from spec compliance or missing features, chances are my fork already has them covered (for example, the last five or so issues opened here on upstream were already addressed in my fork months ago).

You can try either or both and see what works best for you. If you run into issues with my fork, feel free to report them and I'll take a look. Likewise, if you decide to stay on upstream, you can obviously still report problems here. I'll be happy to let you know if they're already solved on my end, and if not, they'll likely end up being addressed there as well.

PR #1061, opened 2026-07-21
feat: add Multipath TCP support via NetworkOptions::set_mptcp

moroznah:
MPTCP lets a single MQTT session survive a path failure by moving traffic to
another subflow, instead of stalling until the client reconnects.

### Safety of the flag

- MPTCP negotiates transparently **down to regular TCP** against a peer that
  doesn't support it.
- If the running **kernel lacks MPTCP** support, socket creation fails and the
  client **silently falls back to a plain TCP socket**.

So enabling the option can never make a connection fail — the socket protocol
is the only thing that changes; `nodelay` / buffers / `bind_device` /
`bind_addr` and the connect path are untouched.

### Scope

- One change site: `socket_connect` (shared by the v4 and v5 event loops), so
  both protocol versions are covered.
- `IPPROTO_MPTCP` is Linux-only, so the field, the `set_mptcp` setter, and the
  `socket2` dependency are all gated to `target_os = "linux"` (mirroring the
  existing `set_bind_device` gating).
- New dependency: `socket2` (Linux target only), used solely to create the
  `IPPROTO_MPTCP` socket, whose fd is handed to tokio's `TcpSocket`.

### Tests

A hermetic Linux-only test (`mptcp_connect_or_fallback_reaches_a_plain_tcp_listener`)
asserts that an `mptcp(true)` connect succeeds against a plain-TCP listener —
covering both the negotiate-down and the kernel-fallback paths. `cargo test -p
rumqttc` passes; `cargo fmt --check` and `cargo clippy` are clean (no new
warnings).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2232,6 +2232,7 @@ dependencies = [
  "rustls-pemfile",
  "rustls-webpki 0.102.8",
  "serde",
+ "socket2 0.5.10",
  "thiserror 2.0.17",
  "tokio",
  "tokio-native-tls",
--- a/rumqttc/CHANGELOG.md
+++ b/rumqttc/CHANGELOG.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- `NetworkOptions::set_mptcp` to opt the outgoing connection onto Multipath TCP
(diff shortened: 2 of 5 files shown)


thehouseisonfire:
Hi @moroznah, and thank you for the PR.

This does seem like a useful feature, and considerably easy to support as well. I have added support for it over on my fork of the project, so if you are in need for support for it, you should be able to use it from there by the next release.

PR #1068, opened 2026-08-06
Fixed deserialization error on zero-length disconnect packet

Andrepuel:
## Type of change

## Checklist:

I am not sure if an entry should be added to CHANGELOG.md

## Summary

A success Disconnect packed can not be read back because the decoder does not allow zero length disconnect packet.

Mqtt Specs states on "3.14.2.1 Disconnect Reason Code" that "The Reason Code and Property Length can be omitted if the Reason Code is 0x00 (Normal disconnecton) and there are no Properties. In this case the DISCONNECT has a Remaining Length of 0."

### DIFF
--- a/rumqttc/src/v5/mqttbytes/v5/mod.rs
+++ b/rumqttc/src/v5/mqttbytes/v5/mod.rs
@@ -70,6 +70,9 @@ impl Packet {
             return match packet_type {
                 PacketType::PingReq => Ok(Packet::PingReq(PingReq)),
                 PacketType::PingResp => Ok(Packet::PingResp(PingResp)),
+                PacketType::Disconnect => Ok(Packet::Disconnect(Disconnect::new(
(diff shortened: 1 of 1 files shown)


thehouseisonfire:
Hi @Andrepuel, and thanks for the PR.

The bug is indeed real, and it has been previously fixed on my fork. One suggestion for your PR would be to use `Disconnect::read` instead of inlining `Disconnect::new`, so something like:

```rs
return match packet_type {
    PacketType::PingReq => Ok(Packet::PingReq(PingReq)),
    PacketType::PingResp => Ok(Packet::PingResp(PingResp)),
    PacketType::Disconnect => Disconnect::read(fixed_header, packet.freeze()).map(Packet::Disconnect),
    _ => Err(Error::PayloadRequired),
};
```

This is better because `Disconnect::read` already knows how to decode a zero length packet and validates it, which the inline shortcut is bypassing.

PR #1037, opened 2026-03-23
chore(deps): bump rustls-webpki and tokio-rustls

sl8vz:
Due to recent vulnerabilities findings, minimal bump to allow lib users to bump necessary crates.

- https://rustsec.org/advisories/RUSTSEC-2026-0047
- https://rustsec.org/advisories/RUSTSEC-2026-0049

PR #1072, opened 2026-09-09
deps(rumqttc): bump rustls-webpki to 0.103

jedwards1230:
## Summary

Bumps `rumqttc`'s `rustls-webpki` pin from `0.102.8` to `0.103.13`.

`rustls-webpki` 0.102.8 has four open advisories, all fixed in 0.103.13:

| Severity | Advisory | Issue |
|---|---|---|
| high | [GHSA-82j2-j2ch-gfr8](https://github.com/advisories/GHSA-82j2-j2ch-gfr8) | Denial of service via panic on a malformed CRL BIT STRING |
| medium | [GHSA-pwjx-qhcg-rvj4](https://github.com/advisories/GHSA-pwjx-qhcg-rvj4) | CRLs not considered authoritative by distribution point due to faulty matching logic |
| low | [GHSA-xgp8-3hg3-c2mh](https://github.com/advisories/GHSA-xgp8-3hg3-c2mh) | Name constraints accepted for certificates asserting a wildcard name |
| low | [GHSA-965h-392x-2mh5](https://github.com/advisories/GHSA-965h-392x-2mh5) | Name constraints for URI names incorrectly accepted |

## Why this has to change here

Downstream crates cannot work around it from their own manifests. `^0.102.8`
does not admit `0.103.x`, so a `[patch.crates-io]` entry for `rustls-webpki` is
applied to whatever *other* 0.103 edge exists in the graph (the one `rustls`
0.23 already pulls) and leaves `rumqttc`'s untouched — the lockfile just grows a
third copy. The pin has to move in this crate.

That is currently the situation for anyone using `rumqttc` with a rustls
feature: `rustls` 0.23 resolves `rustls-webpki` 0.103.x correctly, and
`rumqttc` pulls 0.102.8 in beside it.

## Why it is a one-line change

`rumqttc` uses exactly one item from the crate — `webpki::Error`, in the
`Error::WebPki` variant of `src/tls.rs`. Everything else in that module comes
from `rustls` via `tokio_rustls::rustls::{...}`, and the root store is built
with `rustls::RootCertStore::add_parsable_certificates`. `webpki::Error` is
unchanged between 0.102 and 0.103, so no source changes are needed.

## Verification

Both rustls feature legs build clean against 0.103 with no source edits:

```
cargo build --no-default-features --features use-rustls-no-provider
cargo build --no-default-features --features use-rustls
```

Happy to adjust the version requirement (e.g. `"0.103"` rather than the exact
`"0.103.13"`) or to split it differently if you'd prefer.


jedwards1230:
Closing — this duplicates #1037, which is already approved and additionally bumps `rumqttd`. I should have searched before opening; apologies for the noise.

One note in case it's useful when #1037 lands: its `0.103.10` floor doesn't cover RUSTSEC-2026-0104, which needs `0.103.13`.

PR #1073, opened 2026-09-14
chore(rumqttc): update rustls-webpki to 0.103

tinegachris:
`rustls-webpki` is declared at `0.102.8`, which is the last release of that line. Four advisories were filed against it and are fixed only in 0.103.x:

- RUSTSEC-2026-0049 — CRLs not authoritative by distribution point
- RUSTSEC-2026-0098 — URI name constraints incorrectly accepted
- RUSTSEC-2026-0099 — name constraints accepted for a wildcard name
- RUSTSEC-2026-0104 — reachable panic parsing a CRL

A downstream project running `cargo deny` therefore has no version to move to and has to ignore all four — even though rumqttc itself parses no certificate with the crate. It is linked for the `TlsError::WebPki` variant alone; verification is rustls's own, through the `rustls-webpki` 0.103 that `tokio-rustls` already pulls in. That also means both versions sit in the same dependency tree today, which this collapses to one.

No source change is needed. Checked with:

```
cargo check -p rumqttc --features use-rustls,websocket,proxy
```

`rumqttd` is deliberately left alone: it declares `rustls-webpki` 0.102.2 **and** `tokio-rustls` 0.25, so moving it is a breaking bump and a separate change.


thehouseisonfire:
Hi @tinegachris, and thanks for the PR, but this is a duplicate of #1037.

tinegachris:
Thanks. Closing in favour of #1037.

Two notes for whoever picks that one up, since it has been open since March and is
currently blocked:

- It targets `rustls-webpki` 0.103.10, and a later comment on the same PR points
  out the newer advisories are only fixed in **0.103.12**, so the bump needs to go
  a little further than it currently asks for.
- `rumqttc` 0.25.1 on crates.io still declares `rustls-webpki = "0.102.8"`, and so
  does `main`. Until a release carries 0.103, downstream users are left ignoring
  the four 0.102.x advisories rather than taking a fix — which is what prompted
  this PR. Bumping that one line compiles with no code change (checked against
  `e886a78`).

Happy to rebase and reopen if a one-line bump to `rumqttc` alone turns out to be
easier to land than #1037's wider change.

